PR #1, opened 2015-02-20
U/pgee/dm 980b

SimonKrughoff:


include/lsst/ip/diffim/DipoleAlgorithms.h
```diff
@@ -11,232 +11,375 @@
...
+    LSST_CONTROL_FIELD(maxFnCalls, int, "Maximum function calls for non-linear fitter; 0 = unlimited");
+    PsfDipoleFluxControl() : DipoleFluxControl(), maxPixels(500),
+                             stepSizeCoord(0.1), stepSizeFlux(1.0), errorDef(1.0), maxFnCalls(100000) {}
+};
 
 /**
  *  @brief Intermediate base class for algorithms that compute a centroid.
  */
-class DipoleCentroidAlgorithm : public meas::algorithms::Algorithm {
+class DipoleCentroidAlgorithm : public lsst::meas::base::SimpleAlgorithm {
```
SimonKrughoff: Why is the whole namespace used here when it is not elsewhere in this file?

include/lsst/ip/diffim/DipoleAlgorithms.h
```diff
@@ -11,232 +11,375 @@
...
-protected:
-    explicit DipoleFluxControl(std::string const & name_, double priority=2.0) :
-        AlgorithmControl(name_, priority) {}
+    ResultKey _positiveKeys;
+    ResultKey _negativeKeys;
 };
 
 inline DipoleCentroidAlgorithm::DipoleCentroidAlgorithm(
-    DipoleCentroidControl const & ctrl, KeyTuple const & positiveKeys, KeyTuple const & negativeKeys
+    Control const & ctrl, std::string const & name, afw::table::Schema & schema, std::string const & doc
```
SimonKrughoff: Why not use DipoleCentroidControl?  You do it below with DipoleFluxControl.  It seems more readable to me.

include/lsst/ip/diffim/DipoleAlgorithms.h
```diff
@@ -11,232 +11,375 @@
...
-    LSST_CONTROL_FIELD(stepSizeFlux, float, "Default initial step size for flux in non-linear fitter");
-    LSST_CONTROL_FIELD(errorDef, double, "How many sigma the error bars of the non-linear fitter represent");
-    LSST_CONTROL_FIELD(maxFnCalls, int, "Maximum function calls for non-linear fitter; 0 = unlimited");
-    PsfDipoleFluxControl() : DipoleFluxControl("flux.dipole.psf"), maxPixels(500), 
-                             stepSizeCoord(0.1), stepSizeFlux(1.0), errorDef(1.0), maxFnCalls(100000) {}
+
+    typedef PsfDipoleFluxControl Control;
+
+    PsfDipoleFlux(PsfDipoleFluxControl const & ctrl, std::string const & name, afw::table::Schema & schema) :
+        DipoleFluxAlgorithm(ctrl, name, schema, "jointly fitted psf flux counts"),
```
SimonKrughoff: Might as well use Control instead of PsfDipoleFluxControl, I think.

python/lsst/ip/diffim/dipole.i
```diff
@@ -0,0 +1,102 @@
+// -*- lsst-c++ -*-
+/*
+ * LSST Data Management System
+ * Copyright 2008-2013 LSST Corporation.
```
SimonKrughoff: copyleft  is updated in other files.  Might as well do it here as well.

python/lsst/ip/diffim/diffimLib.i
```diff
@@ -269,23 +273,8 @@ lsst::afw::image::Image<PIXTYPE>
...
-%shared_ptr(lsst::ip::diffim::DipoleCentroidAlgorithm)
-%shared_ptr(lsst::ip::diffim::DipoleCentroidControl)
-%shared_ptr(lsst::ip::diffim::DipoleFluxAlgorithm)
-%shared_ptr(lsst::ip::diffim::DipoleFluxControl)
-%shared_ptr(lsst::ip::diffim::NaiveDipoleCentroidControl)
-%shared_ptr(lsst::ip::diffim::NaiveDipoleFluxControl)
-%shared_ptr(lsst::ip::diffim::PsfDipoleFluxControl)
-
-%include "lsst/ip/diffim/DipoleAlgorithms.h"
+%include  "lsst/ip/diffim/dipole.i"
```
SimonKrughoff: There seems to be significant duplicated code between dipole.i and this file.  Is it worth putting these few lines in a new file if it also requires a bunch of boilerplate as well?

python/lsst/ip/diffim/dipoleMeasurement.py
```diff
@@ -310,14 +312,14 @@ def classify(self, sources):
 
             source.set(key, val)
 
-    def run(self, exposure, sources, **kwds):
+    def run(self, sources, exposure, **kwds):
```
SimonKrughoff: The parameter order was swapped.  Please update the docstring to reflect the change.

src/DipoleAlgorithms.cc
```diff
@@ -274,107 +212,42 @@ class NaiveDipoleFootprinter : public afw::detection::FootprintFunctor<MaskedIma
...
-    afw::table::SourceRecord & source, 
-    afw::image::Exposure<PixelT> const& exposure,
-    afw::geom::Point2D const & center
+void NaiveDipoleFlux::measure(
+    lsst::afw::table::SourceRecord & source, 
+    lsst::afw::image::Exposure<float> const & exposure
 ) const {
-    source.set(getPositiveKeys().flag, true); // say we've failed so that's the result if we throw
-    source.set(getNegativeKeys().flag, true); // say we've failed so that's the result if we throw
+    afw::geom::Point2D center = _centroidExtractor(source, _flagHandler);
```
SimonKrughoff: Here's another example where above you expanded to the fully qualified namespace, but leave it relative here.  I may be missing something, but it makes sense to me to try to be as consistent as possible.

src/DipoleAlgorithms.cc
```diff
@@ -274,107 +212,42 @@ class NaiveDipoleFootprinter : public afw::detection::FootprintFunctor<MaskedIma
...
-    afw::table::KeyTuple<afw::table::Centroid> _posCentroid;
-
-    afw::table::Key< afw::table::Flag > _flagMaxPixelsKey;
-};
-
 /**
  * Class to minimize PsfDipoleFlux; this is the object that Minuit minimizes
  */
-template<typename PixelT>
 class MinimizeDipoleChi2 : public ROOT::Minuit2::FCNBase {
```
SimonKrughoff: Just out of curiosity, why did you remove the template?  It seems fairly likely that float and double would both be useful, but maybe not.

src/DipoleAlgorithms.cc
```diff
@@ -491,34 +363,30 @@ std::pair<double,int> PsfDipoleFlux::chi2(
     return std::pair<double,int>(chi2, nPix);
 }    
  
-template<typename PixelT>
-void PsfDipoleFlux::_apply(
-    afw::table::SourceRecord & source, 
-    afw::image::Exposure<PixelT> const& exposure,
-    afw::geom::Point2D const & center // Not used; source required to have footprint&peaks
+void PsfDipoleFlux::measure(
```
SimonKrughoff: Same question here re. removing the template.

TallJimbo: Here, at least, it's because we've changed the interface in the base class to just take a single-precision float.  That's all we were using in practice, and there was quite a bit of code complexity involved in simulating a templated virtual method, which we've since been able to simplify.

PR #4, opened 2015-05-27
ip_diffim tickets/DM-2535: switch from compound fields to FunctorKeys

TallJimbo:


tests/diaCatalogSourceSelector.py
```diff
@@ -72,7 +72,7 @@ def makeMatches(self, refCat, srcCat, nSrc):
             refSrc.set("r_flux", 10**(-0.4*18))
             refSrc.set("resolved", False)
             refSrc.set("photometric", True)
-            refSrc.set("coord", coord)
+            refSrc.setCoord(coord)
```
r-owen: Do you plan to support setCoord for the long haul? If not, I suggest a key.

TallJimbo: No, I don't.  I'll change it.

TallJimbo: On second thought, I'm going to leave it for now.  I'm actually deferring the removal of those setters and getters for a while, and part of the problem is that I know I need to find a better syntax to replace them, and I don't have that yet.  The only alternative that would work right now in this case is really verbose, which just reemphasizes that problem.

TallJimbo:
Merged to master on tickets/DM-1766.

PR #6, opened 2015-09-23
tickets/DM-3213

r-owen:


python/lsst/ip/diffim/diffimTools.py
```diff
@@ -307,7 +307,10 @@ def sourceToFootprintList(candidateInList, templateExposure, scienceExposure, ke
 
     for kernelCandidate in candidateInList:
         if not type(kernelCandidate) == afwTable.SourceRecord:
-            raise RuntimeError, ("Candiate not of type afwTable.SourceRecord")
+            try:
+                kernelCandidate = kernelCandidate.getSource()
+            except:
+                raise RuntimeError, ("Candiate not of type afwTable.SourceRecord")
```
r-owen: The error message is incomplete, since the problem is that the kernelCandidate is neither a SourceRecord nor an object (what kind?) that has a getSource() method.

Cadiate -> Candidate

Do you expect all candidates to be of the same type? If so, I think it would be safer and cleaner to preprocess candidateInList based on the type of the first entry (e.g. convert it to a list of source records, regardless of what it started out as).

python/lsst/ip/diffim/psfMatch.py
```diff
@@ -1037,7 +1037,7 @@ def _solve(self, kernelCellSet, basisList, returnOnExcept=False):
 
         except pexExcept.Exception as e:
             pexLog.Trace(self.log.getName()+"._solve", 1, "ERROR: Unable to calculate psf matching kernel")
-            pexLog.Trace(self.log.getName()+"._solve", 2, e.args[0].what())
+            pexLog.Trace(self.log.getName()+"._solve", 2, e.args[0])
```
r-owen: Why not just e instead of e.args[0]? Can you be sure the other args are not interesting?

TallJimbo: Just `e` should now work.  It would not have on the HSC side, since it still doesn't have the revamped exception translation.

python/lsst/ip/diffim/diffimTools.py
PaulPrice: Use the modern version of `raise`:

    raise RuntimeError("Candidate something something")

yalsayyad: Depending on where the kernel candidates come from they are either a lsst.afw.table.SourceRecord or a lsst.meas.algorithms.algorithmsLib.PsfCandidateF.  

That is a much better idea to check the type and convert the whole list; I'll move this to the calling task, ImagePsfMatchTask.makeCandidateList(), which already has some type checking in it.

python/lsst/ip/diffim/psfMatch.py
yalsayyad: e does not appear to work with 11.0.rc2 pex_logging that I'm testing with:
(Pdb) pexLog.Trace(self.log.getName()+"._solve", 2, e)
*** NotImplementedError: Wrong number or type of arguments for overloaded function 'new_Trace'.
  Possible C/C++ prototypes are:
    lsst::pex::logging::Trace::Trace(std::string const &,int const,std::string const,...)
    lsst::pex::logging::Trace::Trace(std::string const &,int const,char const *)
    lsst::pex::logging::Trace::Trace(std::string const &,int const,boost::format const &

ktlim: Try `str(e)` instead.

r-owen: Exactly. I also suggest making it a single self-contained error message, and making it a normal logged error, e.g.:
pexLog.error(self.log, "Unable to calculate psf matching kernel: %s" % (e,))

PR #13, opened 2016-04-29
U/yusra/dm 3515

djreiss:


python/lsst/ip/diffim/dipoleMeasurement.py
```diff
@@ -23,34 +23,78 @@
...
-        doc = "Maximum flux ratio in either lobe to be considered a dipole",
+        doc="Maximum flux ratio in either lobe to be considered a dipole",
         dtype = float, default = 0.65
         )
 
+
+@register("ip_diffim_ClassificationDipole")
+class ClassificationDipolePlugin(SingleFramePlugin):
+    """
+    A binary measure of the whether a diasource is a dipole.
```
djreiss: Maybe better: "A plugin to classify whether a diaSource is a dipole."

python/lsst/ip/diffim/dipoleMeasurement.py
```diff
@@ -23,34 +23,78 @@
...
+    ConfigClass = ClassificationDipoleConfig
+
+    @classmethod
+    def getExecutionOrder(cls):
+        return cls.CLASSIFY_ORDER
+
+    def __init__(self, config, name, schema, metadata):
+        SingleFramePlugin.__init__(self, config, name, schema, metadata)
+        self.dipoleAnalysis = DipoleAnalysis()
+        self.keyProbability = schema.addField(name + "_value", type="D",
```
djreiss: Rename from "probability" to "classification" since it is not actually a probability.

python/lsst/ip/diffim/dipoleMeasurement.py
```diff
@@ -172,76 +206,14 @@ class DipoleMeasurementTask(SingleFrameMeasurementTask):
...
-
- . flux.dipole.psf.neg.centroid.err : covariance matrix for flux.dipole.psf.neg.centroid
-
- . flux.dipole.psf.neg.centroid.flags : set if the flux.dipole.psf.neg.centroid measurement did not fully succeed
-
- . flux.dipole.psf.pos.centroid : psf fitted center of positive lobe
-
- . flux.dipole.psf.pos.centroid.err : covariance matrix for flux.dipole.psf.pos.centroid
-
- . flux.dipole.psf.pos.centroid.flags : set if the flux.dipole.psf.pos.centroid measurement did not fully succeed
```
djreiss: Is there a reason you got rid of this documentation rather than updating it? I found it to be very useful. (Or maybe you moved it somewhere that I missed?)

python/lsst/ip/diffim/dipoleMeasurement.py
yalsayyad: This was just copied pasted. I will update.

python/lsst/ip/diffim/dipoleMeasurement.py
yalsayyad: Interesting... I gave some serious thought to the relationship between classification and probability  -- more than is necessary than deciding a name for an instance variable. It is a number between 0.0 and 1.0 inclusive which makes it a probability in the technical sense.

I'm not a fan of the name classification because that's the name of the plugin. This classification currently has two components: an overall probability and a flag. I'd prefer {{self.keyValue}}

Also, Becker's original implementation (just a hook for a better implementation) sets a float with the intention of leaving the door open for the general implementation returning a float between 0 and 1. As a user, I'd like the field to be a float so that I can chose my own cut off for dipoles depending on the application.

python/lsst/ip/diffim/dipoleMeasurement.py
yalsayyad: These are redundant with the docstrings on the original plugins.  Not a single developer who has updated the default plugins in DipoleMeasurementConfig in the last 2 years has bothered to copy and paste the plugin docstrings to this task documentation. I believe this will continue to be the rule and not the exception, and that this falls under code-duplication principles.

PR #14, opened 2016-05-04
DM-5323

r-owen:
lsst.meas.algorithms.SubtractBackgroundTask replaces getBackground
and SubtractBackgroundTask replaces BackgroundConfig

### DIFF
--- a/python/lsst/ip/diffim/imagePsfMatch.py
+++ b/python/lsst/ip/diffim/imagePsfMatch.py
@@ -27,7 +27,7 @@ import lsst.afw.math as afwMath
 import lsst.afw.geom as afwGeom
 import lsst.afw.table as afwTable
 import lsst.pipe.base as pipeBase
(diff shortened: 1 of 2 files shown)


python/lsst/ip/diffim/imagePsfMatch.py
```diff
@@ -281,6 +281,10 @@ def __init__(self, *args, **kwargs):
         PsfMatchTask.__init__(self, *args, **kwargs)
         self.kConfig = self.config.kernel.active
         self._warper = afwMath.Warper.fromConfig(self.kConfig.warpingConfig)
+        # the background subtraction task uses a config from an unusual location,
+        # so cannot easily be constructed with makeSubtask
```
PaulPrice: I think that suggests that some refactoring of the config is warranted.

r-owen: I think ip_diffim needs significant refactoring, but it is out of scope for this ticket.

PR #19, opened 2016-05-16
fixes enabling DipoleFitTask in imageDifference command-line task

SimonKrughoff:
1. add dipoleFitTask.py to exports
2. fix error/warning reporting
3. add DipoleFitTaskConfig

### DIFF
--- a/python/lsst/ip/diffim/dipoleFitTask.py
+++ b/python/lsst/ip/diffim/dipoleFitTask.py
@@ -34,7 +34,7 @@ import lsst.pex.logging as pexLog
 import lsst.pex.config as pexConfig
 from lsst.pipe.base import Struct
 
-__all__ = ("DipoleFitPluginConfig", "DipoleFitTask", "DipoleFitPlugin")
+__all__ = ("DipoleFitTask", "DipoleFitPlugin", "DipoleFitTaskConfig", "DipoleFitPluginConfig")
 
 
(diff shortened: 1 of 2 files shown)


python/lsst/ip/diffim/dipoleFitTask.py
```diff
@@ -88,6 +88,29 @@ class DipoleFitPluginConfig(measBase.SingleFramePluginConfig):
...
+                              "base_PixelFlags",
+                              "base_SkyCoord",
+                              "base_PsfFlux",
+                              "ip_diffim_NaiveDipoleCentroid",
+                              "ip_diffim_NaiveDipoleFlux",
+                              "ip_diffim_PsfDipoleFlux"
+                              ]
+
+        self.slots.calibFlux = None
+        self.slots.modelFlux = None
```
SimonKrughoff: Couldn't we make the modelFlux slot the ip_diffim_PsfDipoleFlux?

djreiss: That is not the current behavior. PsfDipoleFlux does not set an overall flux (just separate fluxes for pos. and neg. lobes). I would set it to the flux from ip_diffim_DipoleFit, but I am not sure how to, since that task is not formally "registered" as a plugin (this will be fixed in a future ticket after the sfm overhaul).

SimonKrughoff: O.K.  Good enough.

PR #20, opened 2016-05-23
Add early error message if using calexp as template and no id supplied

r-owen:


python/lsst/ip/diffim/getTemplate.py
```diff
@@ -168,6 +168,8 @@ def run(self, exposure, sensorRef, templateIdList):
         if len(templateIdList) > 1:
             self.log.warn("Multiple template visits supplied. Getting template from first visit: %s" %
                           (templateIdList[0]['visit']))
+        if len(templateIdList) == 0:
+            raise RuntimeError("No template supplied! Please supply a template visit id.")
```
r-owen: I suggest you swap the two tests, to fail early.

Also, please improve the documentation for the templateIdList argument: indicate that the list should contain a single item, and if it contains more then only the first is used.

PR #21, opened 2016-06-01
DM-5877: Use Afterburners to clean up aperture correction logic

PaulPrice:


PaulPrice:
Second commit includes several whitespace changes that should be moved to a separate commit.

PR #23, opened 2016-06-09
Tickets/dm 5932

r-owen:


python/lsst/ip/diffim/imageDecorrelation.py
```diff
@@ -0,0 +1,191 @@
+from __future__ import absolute_import, division, print_function
+#
+# LSST Data Management System
+# Copyright 2008-2016 AURA/LSST.
```
r-owen: Trivial but...for a new file you may as well just specify 2016

python/lsst/ip/diffim/imageDecorrelation.py
```diff
@@ -0,0 +1,191 @@
...
+from scipy.ndimage.filters import convolve
+
+import lsst.afw.image as afwImage
+import lsst.meas.algorithms as measAlg
+import lsst.afw.math as afwMath
+
+__all__ = ("performExposureDecorrelation")
+
+
+def computeClippedImageStats(im, low=3, high=3):
```
r-owen: We have sigma clipped statistics already available in afw. Please use that if you can, else please document why you chose not to. A subtlety: I'm not sure if the afw code works efficiently on numpy arrays; it may copy the data. If so, consider making an Image of the same type and then working with the array returned by image.getArray(), since that shares the memory with the Image.

python/lsst/ip/diffim/imageDecorrelation.py
```diff
@@ -0,0 +1,191 @@
...
+    """
+    xim = np.arange(np.int(-np.floor(im.shape[0]/2.)), np.int(np.floor(im.shape[0]/2)))
+    yim = np.arange(np.int(-np.floor(im.shape[1]/2.)), np.int(np.floor(im.shape[1]/2)))
+    x0im, y0im = np.meshgrid(xim, yim)
+    return x0im, y0im
+
+
+def computeDecorrelationCorrectionKernel(kappa, sig1=0.2, sig2=0.2):
+    """! Compute the Lupton/ZOGY post-conv. kernel for decorrelating an
+    image difference, based on the PSF-matching kernel.
```
r-owen: Please add a reference to the documentation for this technique if one is available.

python/lsst/ip/diffim/imageDecorrelation.py
```diff
@@ -0,0 +1,191 @@
...
+    @param kappa  A matching kernel array derived from Alard & Lupton PSF matching
+    @param sig1   Average sqrt(variance) of template image used for PSF matching
+    @param sig2   Average sqrt(variance) of science image used for PSF matching
+    @return a 2-d numpy.array containing the correction kernel
+
+    @note As currently implemented, kappa is a static (single) kernel.
+    """
+    def post_conv_kernel_ft2(kernel, sig1=1., sig2=1.):
```
r-owen: Please document this function and the next one as well

r-owen: Also, I am puzzled why you have these two functions instead of just the one below. This one seems too small to stand on its own and I don't think any other code is calling it.

python/lsst/ip/diffim/imageDecorrelation.py
```diff
@@ -0,0 +1,191 @@
...
+    # This is the same as taking the complex conjugate in Fourier space before FFT-ing back to real space.
+    if False:  # TBD: figure this out. For now, we are turning it off.
+        pck = pck[::-1, :]
+
+    return pck
+
+
+def computeCorrectedDiffimPsf(kappa, im2_psf, sig1=0.2, sig2=0.2):
+    """! Compute the (decorrelated) difference image's new PSF:
+    post_conv_psf = phi_1(k) * sym.sqrt((sig1**2 + sig2**2) / (sig1**2 + sig2**2 * kappa_ft(k)**2))
```
r-owen: I suggest moving this equation to the next line and providing a reference to a paper where it can be found. Also, what is "sym" in "sym.sqrt"?

python/lsst/ip/diffim/imageDecorrelation.py
```diff
@@ -0,0 +1,191 @@
...
+    @param kappa  A matching kernel array derived from Alard & Lupton PSF matching
+    @param im2_psf The uncorrected psf array of the science image (and also of the diffim)
+    @param sig1   Average sqrt(variance) of template image used for PSF matching
+    @param sig2   Average sqrt(variance) of science image used for PSF matching
+    @return a 2-d numpy.array containing the new PSF
+    """
+    def post_conv_psf_ft2(psf, kernel, sig1=1., sig2=1.):
+        # Pad psf or kernel symmetrically to make them the same size!
+        if psf.shape[0] < kernel.shape[0]:
+            while psf.shape[0] < kernel.shape[0]:
```
r-owen: Why use a loop instead of simply doing all the padding at once? If that is truly the best way then an explanatory comment would be appreciated.

python/lsst/ip/diffim/imageDecorrelation.py
```diff
@@ -0,0 +1,191 @@
...
+        out = psf_ft * np.sqrt((sig1**2 + sig2**2) / (sig1**2 + sig2**2 * kft**2))
+        return out
+
+    def post_conv_psf(psf, kernel, sig1=1., sig2=1.):
+        kft = post_conv_psf_ft2(psf, kernel, sig1, sig2)
+        out = ifft2(kft)
+        return out
+
+    im2_psf_small = im2_psf
+    # First compute the science image's (im2's) psf, subset on -16:15 coords
```
r-owen: what does -16:15 refer to and where do those values come from? More explanation here would be much appreciated

python/lsst/ip/diffim/imageDecorrelation.py
```diff
@@ -0,0 +1,191 @@
...
+        im2_psf_small = im2_psf[(x0im.max()+x.min()+1):(x0im.max()-x.min()+1),
+                                (y0im.max()+y.min()+1):(y0im.max()-y.min()+1)]
+    pcf = post_conv_psf(psf=im2_psf_small, kernel=kappa, sig1=sig2, sig2=sig1)
+    pcf = pcf.real / pcf.real.sum()
+    return pcf
+
+
+def performExposureDecorrelation(templateExposure, exposure, subtractedExposure, psfMatchingKernel, log):
+    """! Compute decorrelation correction on an image difference exposure, given the
+    input template/science exposures and the (potentially spatially-varying) PSF matching kernel.
```
r-owen: Picky, but please document as a brief one-liner, a blank line, then more info (if wanted). Documenting this way is a PEP8 standard that we do use. In this case you could stop the brief line at "image difference exposure" and then give more detail.

r-owen: Also, if you know that some parameters are modified and others are not, then @param[in], @param[out], and @param[in,out] are helpful information for the user.

On the whole the documentation in this file is very good and much appreciated. I realize I'm bringing up some picky points, but overall I'm quite impressed with the docs.

python/lsst/ip/diffim/imageDecorrelation.py
```diff
@@ -0,0 +1,191 @@
...
+    sig1 = templateExposure.getMaskedImage().getVariance().getArray()
+    sig1squared, _ = computeClippedImageStats(sig1)
+    sig1 = np.sqrt(sig1squared)
+
+    sig2 = exposure.getMaskedImage().getVariance().getArray()
+    sig2squared, _ = computeClippedImageStats(sig2)
+    sig2 = np.sqrt(sig2squared)
+
+    corrKernel = computeDecorrelationCorrectionKernel(kimg.getArray(), sig1=sig1, sig2=sig2)
+    # Eventually, use afwMath.convolve(), but for now we just use scipy.
```
r-owen: This worries me. The afw convolution code is supposed to be easy to use, and has the advantage of dealing with the variance and mask planes correctly. Why not use it?

python/lsst/ip/diffim/imageDecorrelation.py
```diff
@@ -0,0 +1,191 @@
...
+
+    @param templateExposure the template afwImage.Exposure used for PSF matching
+    @param exposure the science afwImage.Exposure used for PSF matching
+    @param subtractedExposure the resulting subtracted exposure produced by 
+    ip_diffim.ImagePsfMatchTask.subtractExposures()
+    @param psfMatchingKernel an (optionally spatially-varying) PSF matching kernel produced 
+    by ip_diffim.ImagePsfMatchTask.subtractExposures()
+    @param log an input pexLog for outputting messages
+
+    @return the corrected subtractedExposure (updated in-place) with an updated PSF, and
```
r-owen: I gently suggest that since you are updating the exposure in place, it is less confusing to not return it. By returning it you have the possibly false impression that the original is not altered. (And again, if you are going to alter an input, please mark it as @param[in,out] (or [out], but that is impossible in Python).

tests/testImageDecorrelation.py
```diff
@@ -0,0 +1,253 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the LSST License Statement and
+# the GNU General Public License along with this program.  If not,
+# see <https://www.lsstcorp.org/LegalNotices/>.
+
+import unittest
+import numpy as np
```
r-owen: Picky, but...please insert a blank line between standard imports and 3rd party imports

tests/testImageDecorrelation.py
```diff
@@ -0,0 +1,253 @@
...
+
+import lsst.utils.tests
+import lsst.afw.image as afwImage
+import lsst.afw.geom as afwGeom
+import lsst.afw.math as afwMath
+import lsst.meas.algorithms as measAlg
+# import lsst.ip.diffim as ipDiffim
```
r-owen: Again, very picky but we prefer you not include commented-out code unless there is a real need for it. If you really want to keep something around, you can use an "if False:" to disable it (which allows linters to parse it, and in the case of C++, the compiler checks the validity).

python/lsst/ip/diffim/imageDecorrelation.py
```diff
@@ -0,0 +1,191 @@
...
+        y = x.copy()
+        x0, y0 = np.meshgrid(x, y)
+        im2_psf_small = im2_psf[(x0im.max()+x.min()+1):(x0im.max()-x.min()+1),
+                                (y0im.max()+y.min()+1):(y0im.max()-y.min()+1)]
+    pcf = post_conv_psf(psf=im2_psf_small, kernel=kappa, sig1=sig2, sig2=sig1)
+    pcf = pcf.real / pcf.real.sum()
+    return pcf
+
+
+def performExposureDecorrelation(templateExposure, exposure, subtractedExposure, psfMatchingKernel, log):
```
r-owen: Please consider renaming to `decorrelateExposure`. It would be nice if you could find a name short enough that you were not tempted to rename it when using it in other code (e.g. the unit test), as that renaming can be confusing.

tests/testImageDecorrelation.py
```diff
@@ -0,0 +1,253 @@
...
+# import lsst.daf.base as dafBase
+import lsst.pex.logging as pexLog
+
+from lsst.ip.diffim.imageDecorrelation import performExposureDecorrelation as doDecorr
+
+def singleGaussian2d(x, y, xc, yc, sigma_x=1., sigma_y=1., theta=0.):
+    """! Generate a 2-d Gaussian, possibly elongated and rotated, on a grid of pixel
+    coordinates given by x,y.
+    @param x,y each a 1-d numpy.array containing x- and y- coordinates for independent variables.
```
r-owen: I'm afraid I don't understand why x and y are arrays instead of scalars. I suggest another go at clarifying this description, or providing an example somewhere in this doc string.

tests/testImageDecorrelation.py
```diff
@@ -0,0 +1,253 @@
...
+    xxc, yyc = x-xc, y-yc
+    out = np.exp(-(a*(xxc**2.) + 2.*b*xxc*yyc + c*(yyc**2.)))
+    out /= out.sum()
+    return out
+
+
+def makeFakeImages(xim=None, yim=None, sig1=0.2, sig2=0.2, psf1=2.2, psf2=3.3, offset=None,
+                   psf_yvary_factor=0., varSourceChange=1/50., theta1=0., theta2=0., im2background=0.,
+                   n_sources=500, seed=66, verbose=False):
+    """! Make the two "images". im1 is the template, im2 is the science image.
```
r-owen: I think it might be clearer to say: `Make two images: a template and a science image"

tests/testImageDecorrelation.py
```diff
@@ -0,0 +1,253 @@
...
+
+    # Add a (constant, for now) background offset to im2
+    if im2background != 0.:  # im2background = 10.
+        if verbose:
+            print('Background:', im2background)
+        im2 += im2background
+
+    im1_psf = singleGaussian2d(x0im, y0im, 0, 0, psf1[0], psf1[1], theta=theta1)
+    im2_psf = singleGaussian2d(x0im, y0im, offset[0], offset[1], psf2[0], psf2[1], theta=theta2)
+    return im1, im2, im1_psf, im2_psf
```
r-owen: Would it make sense to return exposures instead of images? Then you would have a place to put the PSFs and you'd have mask planes and variance planes, should your test ever want to deal with them.

tests/testImageDecorrelation.py
```diff
@@ -0,0 +1,253 @@
...
+        del self.im2ex
+
+    def testDiffimCorrection(self):
+        """! Check that the variance of the corrected diffim matches the theoretical value
+        (to within a 1% tolerance).
+        """
+
+        # Create the matching kernel. We use gaussians, so it's just:
+        sig_match = np.sqrt((self.psf2**2. - self.psf1**2.))
+        matchingKernel = measAlg.SingleGaussianPsf(25, 25, sig_match)
```
r-owen: I admit I didn't know the matching kernel was a pure Gaussian, and I thought you were generating double Gaussians above? Where does 25 come from?

tests/testImageDecorrelation.py
```diff
@@ -0,0 +1,253 @@
...
+
+        #print(self.im1.var())
+        #print(self.im2.var())
+        #print((self.im1 - self.im2).var())
+        print((matched_im1 - self.im2).var())
+        # Uncorrected diffim - variance is wrong (too low)
+        self.assertNotClose((matched_im1 - self.im2).var(), expected_var, rtol=0.1)
+        # In fact, it should have a variance close to the variance of the science image.
+        self.assertClose((matched_im1 - self.im2).var(), self.sig2**2., rtol=0.1)
+
```
r-owen: How are the tests on the next few lines different from those above? I am probably missing something obvious, but to me they similar or identical, but one with numpy arrays and the other with exposures containing numpy arrays.

python/lsst/ip/diffim/imageDecorrelation.py
djreiss: OK.

python/lsst/ip/diffim/imageDecorrelation.py
djreiss: Not quite available yet. Will be in a DMTN (ticket to be created).
I will create the ticket, initiate the technote (to get the correct number) and then add the reference here.

python/lsst/ip/diffim/imageDecorrelation.py
djreiss: OK, fixed. The reference will be the aforementioned DMTN.

python/lsst/ip/diffim/imageDecorrelation.py
djreiss: OK, updated.

python/lsst/ip/diffim/imageDecorrelation.py
djreiss: OK, changed. Also updated the calling code in pipe_tasks (imageDifference.py).

python/lsst/ip/diffim/imageDecorrelation.py
djreiss: Agreed. Also changed caller in pipe_tasks.

tests/testImageDecorrelation.py
djreiss: It was accidentally left in there. Removed.

tests/testImageDecorrelation.py
djreiss: Done.

tests/testImageDecorrelation.py
djreiss: You are partly right -- the test on line 224 is essentially a repeat of the one on line 215, but testing the exposure rather than the numpy arrays (for sanity). The test on line 232 is the central test - that the correction is performing as expected. I am keeping both because I want to ensure that the correct data is going into the main test of the correction.

tests/testImageDecorrelation.py
djreiss: Sorry, this was a little confusing. The matching kernel in this case is Gaussian because the (input) PSFs of the two synthesized images are Gaussian. The PSF1 and PSF2 sigma's are already stored in setUp as psf1 and psf2. But now I will just compute them from the exposures' PSF's shapes as you suggested.

tests/testImageDecorrelation.py
djreiss: I do convert these arrays into Exposures in the setUp function but I suppose it makes sense to just move that code into this function and just return the exposures themselves. I will do so.

tests/testImageDecorrelation.py
djreiss: Done.

tests/testImageDecorrelation.py
djreiss: They provide the possibility of non-zero offset bounding boxes, for example if x equals np.arange(-16,15) and xc=0. then we will make a gaussian centered at zero with the x coordinate going from -16 to +15.

python/lsst/ip/diffim/imageDecorrelation.py
djreiss: At times, I compute the clipped stats on the variance array. Is there a way to tell afwMath to run statistics on the variance plane rather than the image plane?

tests/testImageDecorrelation.py
r-owen: Are you familiar with assertMaskedImagesNearlyEqual? It might do the job, without the need to test twice. It is defined in afw/image/testUtils and can be used as self.assertMaskedImages... and either of the two items being compared can be a MaskedImage or a set of numpy arrays such as is returned from MaskedImage.getArrays(). There are similar functions for comparing just images (assetImagesNearlyEqual) and masks (assertMasksEqual), and a lot of flexibility about ignoring masked pixels and such.

python/lsst/ip/diffim/imageDecorrelation.py
r-owen: Yes. The variance plane can be treated as an ordinary image, so you can compute statistics on it just like you do on the image plane. It gets trickier if you are ignoring masked pixels.

r-owen: I'm not sure I was clear. I mean: variance = maskedImage.getVariance(); compute statistics on variance as if it was an image (since it is an instance of Image)

RobertLuptonTheGood: That's easy, the afw stats stuff can handle an Image or a MaskedImage.   The problem comes (as Russell implied) if you want to use the Mask plane too.  That can't be done (but it'd be a good idea....)

r-owen: Actually, I think you can also compute variance plus mask. I looked at makeStatistics and found this overloaded version, which looks just right:

    makeStatistics(lsst::afw::image::Image<Pixel> const &img,
                  lsst::afw::image::Mask<image::MaskPixel> const &msk, 
                  int const flags,  
                  StatisticsControl const& sctrl = StatisticsControl() 
                  )

python/lsst/ip/diffim/imageDecorrelation.py
djreiss: OK, removed the two internal functions.

python/lsst/ip/diffim/imageDecorrelation.py
djreiss: OK, trying to get this to work.

tests/testImageDecorrelation.py
djreiss: That is useful info, thanks - I was not aware. However, I am just comparing variances here, not entire images. Perhaps I will use it in the future, however.

r-owen: assertImagesNearlyEqual will handle the variance plane (since it is just an image). I strongly recommend using it, since it is our standard way to compare images, variances, masks and masked images.

python/lsst/ip/diffim/imageDecorrelation.py
djreiss: OK, this is all fixed now. I compute the statistics on the variance image using the correct mask flags via the StatisticsControl.

python/lsst/ip/diffim/imageDecorrelation.py
djreiss: This was leftover code from my development and is not going to be needed in production, so I removed it.

PR #25, opened 2016-06-23
Add DecorrelateALKernelTask

djreiss:
move utility functions into task

### DIFF
--- a/python/lsst/ip/diffim/imageDecorrelation.py
+++ b/python/lsst/ip/diffim/imageDecorrelation.py
@@ -28,142 +28,268 @@ from scipy.ndimage.filters import convolve as scipy_convolve
 import lsst.afw.image as afwImage
 import lsst.meas.algorithms as measAlg
 import lsst.afw.math as afwMath
+import lsst.pex.config as pexConfig
+import lsst.pipe.base as pipeBase
 
-__all__ = ("decorrelateExposure", "computeDecorrelationKernel", "doConvolve",
-           "computeClippedImageStats")
+__all__ = ("DecorrelateALKernelTask", "DecorrelateALKernelConfig")
 
 
-def computeClippedImageStats(im):
-    """! Utility function for sigma-clipped array statistics on an image or exposure.
-    @param im An afw.Exposure, masked image, or image.
-    @return sigma-clipped mean, std, and variance of input array
+class DecorrelateALKernelConfig(pexConfig.Config):
+    """!
+    \anchor DecorrelateALKernelConfig_
+
+    \brief Configuration parameters for the DecorrelateALKernelTask
+
+    Currently there are no parameters for DecorrelateALKernelTask.
     """
-    statsControl = afwMath.StatisticsControl()
-    statsControl.setNumSigmaClip(3.)
-    statsControl.setNumIter(3)
-    statsControl.setAndMask(afwImage.MaskU.getPlaneBitMask(["INTRP", "EDGE",
-                                                            "DETECTED", "BAD",
-                                                            "NO_DATA", "DETECTED_NEGATIVE"]))
-    statObj = afwMath.makeStatistics(im, afwMath.MEANCLIP | afwMath.STDEVCLIP | afwMath.VARIANCECLIP,
-                                     statsControl)
-    mean = statObj.getValue(afwMath.MEANCLIP)
-    std = statObj.getValue(afwMath.STDEVCLIP)
-    var = statObj.getValue(afwMath.VARIANCECLIP)
-    return mean, std, var
+    pass
 
+## \addtogroup LSST_task_documentation
+## \{
+## \page DecorrelateALKernelTask
+## \ref DecorrelateALKernelTask_ "DecorrelateALKernelTask"
+##      Decorrelate the effect of convolution by Alard-Lupton matching kernel in image difference
+## \}
 
-def getImageGrid(im):
-    """! Convenience function for creating a grid of pixel coordinates covering a 2-d array,
-    using `numpy.meshgrid()`
-    @param im input 2-d numpy array
-    @return two numpy arraysm each with same shape as `im`, containing x- and y- coordinate
-    grid.
+class DecorrelateALKernelTask(pipeBase.Task):
+    """!
+    \anchor DecorrelateALKernelTask_
+
+    \brief Decorrelate the effect of convolution by Alard-Lupton matching kernel in image difference
+
+    \section pipe_tasks_multiBand_Contents Contents
+
+      - \ref ip_diffim_imageDecorrelation_DecorrelateALKernelTask_Purpose
+      - \ref ip_diffim_imageDecorrelation_DecorrelateALKernelTask_Config
+      - \ref ip_diffim_imageDecorrelation_DecorrelateALKernelTask_Run
+      - \ref ip_diffim_imageDecorrelation_DecorrelateALKernelTask_Debug
+      - \ref ip_diffim_imageDecorrelation_DecorrelateALKernelTask_Example
+
+    \section ip_diffim_imageDecorrelation_DecorrelateALKernelTask_Purpose	Description
+
+    Pipe-task that removes the neighboring-pixel covariance in an
+    image difference that are added when the template image is
+    convolved with the Alard-Lupton PSF matching kernel.
+
+    The image differencing pipeline task \link
+    ip.diffim.psfMatch.PsfMatchTask PSFMatchTask\endlink and \link
+    ip.diffim.psfMatch.PsfMatchConfigAL PSFMatchConfigAL\endlink uses
+    the Alard and Lupton (1998) method for matching the PSFs of the
+    template and science exposures prior to subtraction. The
+    Alard-Lupton method identifies a matching kernel, which is then
+    (typically) convolved with the template image to perform PSF
+    matching. This convolution has the effect of adding covariance
+    between neighboring pixels in the template image, which is then
+    added to the image difference by subtraction.
+
+    The pixel covariance may be corrected by whitening the noise of
+    the image difference. This task performs such a decorrelation by
+    computing a decorrelation kernel (based upon the A&L matching
+    kernel and variances in the template and science images) and
+    convolving the image difference with it. This process is described
+    in detail in [DMTN-021](http://dmtn-021.lsst.io).
+
+    \section ip_diffim_imageDecorrelation_DecorrelateALKernelTask_Initialize       Task initialization
+
+    \copydoc \_\_init\_\_
+
+    \section ip_diffim_imageDecorrelation_DecorrelateALKernelTask_Run       Invoking the Task
+
+    \copydoc run
+
+    \section ip_diffim_imageDecorrelation_DecorrelateALKernelTask_Config       Configuration parameters
+
+    This task currently has no relevant configuration parameters.
+    See \ref DecorrelateALKernelConfig
+
+    \section ip_diffim_imageDecorrelation_DecorrelateALKernelTask_Debug		Debug variables
+
+    This task has no debug variables
+
+    \section ip_diffim_imageDecorrelation_DecorrelateALKernelTask_Example	Example of using DecorrelateALKernelTask
+
+    This task has no standalone example, however it is applied as a
+    subtask of \link pipe.tasks.imageDifference.ImageDifferenceTask ImageDifferenceTask\endlink .
+
     """
-    xim = np.arange(np.int(-np.floor(im.shape[0]/2.)), np.int(np.floor(im.shape[0]/2)))
-    yim = np.arange(np.int(-np.floor(im.shape[1]/2.)), np.int(np.floor(im.shape[1]/2)))
-    x0im, y0im = np.meshgrid(xim, yim)
-    return x0im, y0im
+    ConfigClass = DecorrelateALKernelConfig
+    _DefaultName = "ip_diffim_decorrelateALKernel"
 
+    def __init__(self, *args, **kwargs):
+        """! Create the image decorrelation Task
+        @param *args arguments to be passed to lsst.pipe.base.task.Task.__init__
+        @param **kwargs keyword arguments to be passed to lsst.pipe.base.task.Task.__init__
+        """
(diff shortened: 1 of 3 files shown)


python/lsst/ip/diffim/imageDecorrelation.py
```diff
@@ -28,19 +28,21 @@
...
+import lsst.pex.config as pexConfig
+import lsst.pipe.base as pipeBase
 
-__all__ = ("decorrelateExposure", "computeDecorrelationKernel", "doConvolve",
+__all__ = ("ImageDecorrelationTask", "ImageDecorrelationConfig",
            "computeClippedImageStats")
 
 
-def computeClippedImageStats(im):
+def computeClippedImageStats(im, numSigmaClip=3., numIter=3):
```
r-owen: I suggest removing this function entirely. You only call it in two places in this file, and those two places can use the same statsControl object, so it's very few extra lines of code. Furthermore, in this file you ignore two out of the three statistics that you compute (you apparently only use the variance). I don't feel this is a good function to make public, either.

python/lsst/ip/diffim/imageDecorrelation.py
```diff
@@ -52,188 +54,211 @@ def computeClippedImageStats(im):
...
-    x0im, y0im = np.meshgrid(xim, yim)
-    return x0im, y0im
+class ImageDecorrelationConfig(pexConfig.Config):
+    """! Configuration for image decorrelation"""
+    useScipy = pexConfig.Field(
+        dtype = bool,
+        doc = """If True use scipy for image convolution (debugging option);
+                 if False use afwMath""",
+        default = False
+    )
```
r-owen: Minor suggestions here:
- no spaces around "=" when specifying values (run AutoPep8 to see this and clean it up)
- I suggest that doc strings not include hard returns. Consider this, instead (which uses automatic concatenation of adjacent strings):

```
doc="If True use scipy for image convolution (debugging option); "
    "if False use afwMath",

```

python/lsst/ip/diffim/imageDecorrelation.py
```diff
@@ -52,188 +54,211 @@ def computeClippedImageStats(im):
...
+    _DefaultName = "imageDecorrelation"
+
+    def __init__(self, *args, **kwargs):
+        """! Create the image decorrelation Task
+        @param *args arguments to be passed to lsst.pipe.base.task.Task.__init__
+        @param **kwargs keyword arguments to be passed to lsst.pipe.base.task.Task.__init__
+        """
+        pipeBase.Task.__init__(self, *args, **kwargs)
+
+    def decorrelateExposure(self, templateExposure, exposure, subtractedExposure, psfMatchingKernel):
```
r-owen: If this is the primary method of the task then please call it "run"

python/lsst/ip/diffim/imageDecorrelation.py
```diff
@@ -52,188 +54,211 @@ def computeClippedImageStats(im):
...
 
 
-def computeDecorrelationKernel(kappa, sig1=0.2, sig2=0.2):
-    """! Compute the Lupton/ZOGY post-conv. kernel for decorrelating an
-    image difference, based on the PSF-matching kernel.
-    @param kappa  A matching kernel 2-d numpy.array derived from Alard & Lupton PSF matching
-    @param sig1   Average sqrt(variance) of template image used for PSF matching
-    @param sig2   Average sqrt(variance) of science image used for PSF matching
-    @return a 2-d numpy.array containing the correction kernel
+class ImageDecorrelationTask(pipeBase.Task):
```
r-owen: Picky, but consider this name, instead of the passive form: DecorrelateImageTask

python/lsst/ip/diffim/imageDecorrelation.py
```diff
@@ -52,188 +54,211 @@ def computeClippedImageStats(im):
...
+        psfcI = afwImage.ImageD(psfc.shape[0], psfc.shape[1])
+        psfcI.getArray()[:, :] = psfc
+        psfcK = afwMath.FixedKernel(psfcI)
+        psfNew = measAlg.KernelPsf(psfcK)
+        correctedExposure.setPsf(psfNew)
+        self.log.info("Complete.")
+
+        return correctedExposure, corrKern
+
+    def getImageGrid(self, im):
```
r-owen: This does not use self, so can be a static method. Should it be public? Why? If not, please start with _ in the name

r-owen: Similarly below: please rename any private methods that are not part of the API with a leading underscore to hint that they are private and not to be relied on.

python/lsst/ip/diffim/imageDecorrelation.py
```diff
@@ -52,188 +54,211 @@ def computeClippedImageStats(im):
...
+        fkernel = self.fixEvenKernel(kernel)
+        if use_scipy:
+            pci = scipy_convolve(exposure.getMaskedImage().getImage().getArray(),
+                                 fkernel, mode='constant', cval=np.nan)
+            outExp = exposure.clone()
+            outExp.getMaskedImage().getImage().getArray()[:, :] = pci
+            kern = fkernel
 
-
-def fixEvenKernel(kernel):
```
r-owen: If you know the offset of the kernel in advance (as I suspect you do) then I suggest just applying it, e.g. by changing the position of the center of the kernel or trimming if you are keen to have a smaller kernel. Measuring the position of the center seems more complex and prone to add behavior if the kernel is unexpected. On the other hand, if you cannot predict which way to shift the center, then measuring is safer.

python/lsst/ip/diffim/imageDecorrelation.py
```diff
@@ -52,188 +54,211 @@ def computeClippedImageStats(im):
...
-
-
-def doConvolve(exposure, kernel, use_scipy=False):
-    """! Convolve an Exposure with a convolution kernel.
-    @param exposure Input afw.image.Exposure to be convolved.
-    @param kernel Input 2-d numpy.array to convolve the image with
-    @param use_scipy Use scipy to do convolution instead of afwMath
-    @return a new Exposure with the convolved pixels.
-
-    @note we use afwMath.convolve() but keep scipy.convolve for debugging.
```
r-owen: I suggest not leaving scipy.convolve for debugging unless you have a really important reason. It clutters up the code and mistreats mask and variance.

python/lsst/ip/diffim/imageDecorrelation.py
```diff
@@ -52,188 +54,211 @@ def computeClippedImageStats(im):
...
-    psfcI.getArray()[:, :] = psfc
-    psfcK = afwMath.FixedKernel(psfcI)
-    psfNew = measAlg.KernelPsf(psfcK)
-    correctedExposure.setPsf(psfNew)
-
-    return correctedExposure, corrKern
+            kernelImg = afwImage.ImageD(fkernel.shape[0], fkernel.shape[1])
+            kernelImg.getArray()[:, :] = fkernel
+            kern = afwMath.FixedKernel(kernelImg)
+            maxloc = np.unravel_index(np.argmax(fkernel), fkernel.shape)
```
r-owen: Is it necessary to measure the max location? If you always know it should be shifted by a particular amount then I suggest you simply perform the shift.

python/lsst/ip/diffim/imageDecorrelation.py
```diff
@@ -52,188 +54,211 @@ def computeClippedImageStats(im):
...
 
-    pcf = post_conv_psf(psf=im2_psf, kernel=kappa, sig1=sig2, sig2=sig1)
-    pcf = pcf.real / pcf.real.sum()
-    return pcf
+    def doConvolve(self, exposure, kernel, use_scipy=False):
+        """! Convolve an Exposure with a convolution kernel.
+        @param exposure Input afw.image.Exposure to be convolved.
+        @param kernel Input 2-d numpy.array to convolve the image with
+        @param use_scipy Use scipy to do convolution instead of afwMath
+        @return a new Exposure with the convolved pixels.
```
r-owen: This appears to be incorrect, since you return two items: the exposure and `kern`.

Why return the convolved exposure, since you apparently modify it in place? Best to do one or the other.

For any arguments that are modified please mark them @param[in,out]

python/lsst/ip/diffim/imageDecorrelation.py
```diff
@@ -52,188 +54,211 @@ def computeClippedImageStats(im):
...
+        psf = subtractedExposure.getPsf().computeImage().getArray()
+        psfc = self.computeCorrectedDiffimPsf(fcorrKernel, psf, sig1=sig1, sig2=sig2)
+        psfcI = afwImage.ImageD(psfc.shape[0], psfc.shape[1])
+        psfcI.getArray()[:, :] = psfc
+        psfcK = afwMath.FixedKernel(psfcI)
+        psfNew = measAlg.KernelPsf(psfcK)
+        correctedExposure.setPsf(psfNew)
+        self.log.info("Complete.")
+
+        return correctedExposure, corrKern
```
r-owen: Please return a pipeBase.Struct. This is our standard for all task run methods, and any other method that returns multiple items, because it supports adding additional items later without breaking existing code.

python/lsst/ip/diffim/imageDecorrelation.py
```diff
@@ -52,188 +54,211 @@ def computeClippedImageStats(im):
...
+            if out.shape[1] - maxloc[1] > maxloc[1]:
+                out = out[:, :-1]
+            else:
+                out = out[:, 1:]
         return out
 
-    pcf = post_conv_psf(psf=im2_psf, kernel=kappa, sig1=sig2, sig2=sig1)
-    pcf = pcf.real / pcf.real.sum()
-    return pcf
+    def doConvolve(self, exposure, kernel, use_scipy=False):
```
r-owen: Please eliminate the use_scipy argument unless you have a really strong argument for keeping it in. It adds clutter and scipy doesn't handle the mask and variance planes.

Also, why does this method even exist? Without use_scipy it is basically a one-liner.

r-owen: Let me rephrase that. I suggest you eliminate this method. The strange handling of the kernel is probably best left to a different method . Then this method is basically a one-liner (maybe 3 lines) and can be removed.

python/lsst/ip/diffim/imageDecorrelation.py
djreiss: It is also used by the unit test, which is why I made it public -- but I understand your point. I will do what you suggested, and reimplement it in the unit test.

python/lsst/ip/diffim/imageDecorrelation.py
djreiss: Thanks, fixed.

python/lsst/ip/diffim/imageDecorrelation.py
djreiss: OK, done.

python/lsst/ip/diffim/imageDecorrelation.py
djreiss: OK, changed throughout.

python/lsst/ip/diffim/imageDecorrelation.py
djreiss: OK, done. I left computeCorrectedDiffimPsf() as "public" but made it static.

python/lsst/ip/diffim/imageDecorrelation.py
djreiss: I do not have that info a priori. This is the easiest way to make sure that all edge-cases are handled correctly.

python/lsst/ip/diffim/imageDecorrelation.py
djreiss: Can afwMath.convolve() do convolution in-place? If it can, I'm impressed and I will do it. I could not find such assurances in the docs.

python/lsst/ip/diffim/imageDecorrelation.py
djreiss: Understood. I will do that from now on.

python/lsst/ip/diffim/imageDecorrelation.py
djreiss: I am sympathetic to your suggestion. However, for now until I am 100% confident that the convolution is doing what I want it to (subject to more testing, which is going to happen in the next sprint), I am going to keep the scipy code in there. I can add a ticket to remove it after the more thorough tests are done.

python/lsst/ip/diffim/imageDecorrelation.py
r-owen: I'm not sure, now that you mention it. Convolve doesn't usually work in place and maybe cannot. My mistake.

RobertLuptonTheGood: It can be done in place (I did for SDSS);  it takes a little space, but not much (especially for separable kernels).

But we didn't.

djreiss: It would be worth adding that capability for LSST, I can file a ticket.

python/lsst/ip/diffim/imageDecorrelation.py
r-owen: It is your call, but the difficulty with kernels worries me. Convolution is well tested and this seems a strange place to test it.

python/lsst/ip/diffim/imageDecorrelation.py
r-owen: I don't understand what the safe thing to do is. If the kernel was reliably computed with the center rounded up (since LSST rounds it down, and it's off by one) then you really ought to just set the center to the correct value (and trim if you feel you must, but that is strictly optional). If the high point can be anywhere then this makes sense. What worries me is if the center is always up and to the right, but the computed kernel is weird for some reason, then your measurement of the peak may do the wrong thing.

python/lsst/ip/diffim/imageDecorrelation.py
djreiss: OK, that is fair. I will remove the scipy option.

PaulPrice:
Please be sure to squash the new commits before merging.

PR #26, opened 2016-07-21
fix variance calculation to use masks

djreiss:
fix normalization of correction kernel
remove commented print statements
add coordinate parameters to `run` for testing
update test for stricter tolerance after bug fix
turn on old ClassificationDipole again
add options to DecorrelateALKerneltask.run()
change all sig**2 to variance
Fix mixup of var1 and var2 for science and template

### DIFF
--- a/python/lsst/ip/diffim/imageDecorrelation.py
+++ b/python/lsst/ip/diffim/imageDecorrelation.py
@@ -22,8 +22,7 @@ from __future__ import absolute_import, division, print_function
 #
 
 import numpy as np
-from scipy.fftpack import fft2, ifft2, ifftshift
-from scipy.ndimage.filters import convolve as scipy_convolve
+import scipy.fftpack
 
 import lsst.afw.image as afwImage
 import lsst.meas.algorithms as measAlg
@@ -39,10 +38,13 @@ class DecorrelateALKernelConfig(pexConfig.Config):
     \anchor DecorrelateALKernelConfig_
 
     \brief Configuration parameters for the DecorrelateALKernelTask
-
-    Currently there are no parameters for DecorrelateALKernelTask.
     """
-    pass
+
+    ignoreMaskPlanes = pexConfig.ListField(
+        dtype = str,
+        doc = """Mask planes to ignore for sigma-clipped statistics""",
+        default = ("INTRP", "EDGE", "DETECTED", "SAT", "CR", "BAD", "NO_DATA", "DETECTED_NEGATIVE")
+    )
 
 ## \addtogroup LSST_task_documentation
 ## \{
@@ -125,24 +127,42 @@ class DecorrelateALKernelTask(pipeBase.Task):
         self.statsControl = afwMath.StatisticsControl()
         self.statsControl.setNumSigmaClip(3.)
         self.statsControl.setNumIter(3)
-        self.statsControl.setAndMask(afwImage.MaskU.getPlaneBitMask(["INTRP", "EDGE",
-                                                                     "DETECTED", "BAD",
-                                                                     "NO_DATA", "DETECTED_NEGATIVE"]))
+        self.statsControl.setAndMask(afwImage.MaskU.getPlaneBitMask(self.config.ignoreMaskPlanes))
+
+    def computeVarianceMean(self, exposure):
+        statObj = afwMath.makeStatistics(exposure.getMaskedImage().getVariance(),
+                                         exposure.getMaskedImage().getMask(),
+                                         afwMath.MEANCLIP, self.statsControl)
+        var = statObj.getValue(afwMath.MEANCLIP)
+        return var
 
     @pipeBase.timeMethod
-    def run(self, templateExposure, exposure, subtractedExposure, psfMatchingKernel):
+    def run(self, exposure, templateExposure, subtractedExposure, psfMatchingKernel,
+            xcen=None, ycen=None, svar=None, tvar=None):
         """! Perform decorrelation of an image difference exposure.
 
         Decorrelates the diffim due to the convolution of the templateExposure with the
         A&L PSF matching kernel. Currently can accept a spatially varying matching kernel but in
-        this case it simply uses a static kernel from the center of the exposure.
+        this case it simply uses a static kernel from the center of the exposure. The decorrelation
+        is described in [DMTN-021, Equation 1](http://dmtn-021.lsst.io/#equation-1), where
+        `exposure` is I_1; templateExposure is I_2; `subtractedExposure` is D(k);
+        `psfMatchingKernel` is kappa; and svar and tvar are their respective
+        variances (see below).
 
-        @param[in] templateExposure the template afwImage.Exposure used for PSF matching
         @param[in] exposure the science afwImage.Exposure used for PSF matching
+        @param[in] templateExposure the template afwImage.Exposure used for PSF matching
         @param[in] subtractedExposure the subtracted exposure produced by
         `ip_diffim.ImagePsfMatchTask.subtractExposures()`
         @param[in] psfMatchingKernel an (optionally spatially-varying) PSF matching kernel produced
         by `ip_diffim.ImagePsfMatchTask.subtractExposures()`
+        @param[in] xcen X-pixel coordinate to use for computing constant matching kernel to use
+        If `None` (default), then use the center of the image.
+        @param[in] ycen Y-pixel coordinate to use for computing constant matching kernel to use
+        If `None` (default), then use the center of the image.
+        @param[in] svar image variance for science image
+        If `None` (default) then compute the variance over the entire input science image.
+        @param[in] tvar image variance for template image
+        If `None` (default) then compute the variance over the entire input template image.
 
         @return a `pipeBase.Struct` containing:
             * `correctedExposure`: the decorrelated diffim
@@ -159,35 +179,40 @@ class DecorrelateALKernelTask(pipeBase.Task):
         """
         self.log.info("Starting.")
         kimg = None
-        try:
-            if psfMatchingKernel.isSpatiallyVarying():
-                spatialKernel = psfMatchingKernel
-                kimg = afwImage.ImageD(spatialKernel.getDimensions())
-                bbox = subtractedExposure.getBBox()
-                xcen = (bbox.getBeginX() + bbox.getEndX()) / 2.
-                ycen = (bbox.getBeginY() + bbox.getEndY()) / 2.
-                spatialKernel.computeImage(kimg, True, xcen, ycen)
-        except:  # Not a spatially-varying kernel (usually not true)
-            kimg = psfMatchingKernel.computeImage()
 
-        # Compute the images' sigmas (sqrt of variance)
-        statObj = afwMath.makeStatistics(templateExposure.getMaskedImage().getVariance(), afwMath.MEANCLIP,
-                                         self.statsControl)
-        sig1 = np.sqrt(statObj.getValue(afwMath.MEANCLIP))
+        spatialKernel = psfMatchingKernel
+        kimg = afwImage.ImageD(spatialKernel.getDimensions())
+        bbox = subtractedExposure.getBBox()
+        if xcen is None:
+            xcen = (bbox.getBeginX() + bbox.getEndX()) / 2.
+        if ycen is None:
+            ycen = (bbox.getBeginY() + bbox.getEndY()) / 2.
+        self.log.info("Using matching kernel computed at (%d, %d)" % (xcen, ycen))
+        spatialKernel.computeImage(kimg, True, xcen, ycen)
 
-        statObj = afwMath.makeStatistics(exposure.getMaskedImage().getVariance(), afwMath.MEANCLIP,
-                                         self.statsControl)
-        sig2 = np.sqrt(statObj.getValue(afwMath.MEANCLIP))
+        if False:  # debug code to save spatially varying kernel for analysis
+            import cPickle
+            import gzip
+            cPickle.dump(spatialKernel, gzip.GzipFile('spatialKernel.p.gz', 'wb'))
 
-        corrKernel = DecorrelateALKernelTask._computeDecorrelationKernel(kimg.getArray(), sig1=sig1, sig2=sig2)
-        fcorrKernel = DecorrelateALKernelTask._fixEvenKernel(corrKernel)
+        if svar is None:
+            svar = self.computeVarianceMean(exposure)
+        self.log.info("Variance (science): %f" % svar)
+        if tvar is None:
+            tvar = self.computeVarianceMean(templateExposure)
+        self.log.info("Variance (template): %f" % tvar)
+
+        var = self.computeVarianceMean(subtractedExposure)
+        self.log.info("Variance (uncorrected diffim): %f" % var)
+
+        corrKernel = DecorrelateALKernelTask._computeDecorrelationKernel(kimg.getArray(), svar, tvar)
         self.log.info("Convolving.")
-        correctedExposure, corrKern = DecorrelateALKernelTask._doConvolve(subtractedExposure, fcorrKernel)
+        correctedExposure, corrKern = DecorrelateALKernelTask._doConvolve(subtractedExposure, corrKernel)
         self.log.info("Updating correctedExposure and its PSF.")
 
         # Compute the subtracted exposure's updated psf
         psf = subtractedExposure.getPsf().computeImage().getArray()
-        psfc = DecorrelateALKernelTask.computeCorrectedDiffimPsf(fcorrKernel, psf, sig1=sig1, sig2=sig2)
+        psfc = DecorrelateALKernelTask.computeCorrectedDiffimPsf(corrKernel, psf, svar=svar, tvar=tvar)
         psfcI = afwImage.ImageD(psfc.shape[0], psfc.shape[1])
         psfcI.getArray()[:, :] = psfc
         psfcK = afwMath.FixedKernel(psfcI)
@@ -195,43 +220,48 @@ class DecorrelateALKernelTask(pipeBase.Task):
         correctedExposure.setPsf(psfNew)
         self.log.info("Complete.")
 
+        var = self.computeVarianceMean(correctedExposure)
+        self.log.info("Variance (corrected diffim): %f" % var)
+
         return pipeBase.Struct(correctedExposure=correctedExposure, correctionKernel=corrKern)
 
     @staticmethod
-    def _computeDecorrelationKernel(kappa, sig1=0.2, sig2=0.2):
+    def _computeDecorrelationKernel(kappa, svar=0.04, tvar=0.04):
         """! Compute the Lupton/ZOGY post-conv. kernel for decorrelating an
         image difference, based on the PSF-matching kernel.
         @param kappa  A matching kernel 2-d numpy.array derived from Alard & Lupton PSF matching
-        @param sig1   Average sqrt(variance) of template image used for PSF matching
-        @param sig2   Average sqrt(variance) of science image used for PSF matching
+        @param svar   Average variance of science image used for PSF matching
+        @param tvar   Average variance of template image used for PSF matching
         @return a 2-d numpy.array containing the correction kernel
 
-        @note As currently implemented, kappa is a static (single) kernel.
+        @note As currently implemented, kappa is a static (single, non-spatially-varying) kernel.
         """
-        kft = fft2(kappa)
-        kft = np.sqrt((sig1**2 + sig2**2) / (sig1**2 + sig2**2 * kft**2))
-        pck = ifft2(kft)
-        pck = ifftshift(pck.real)
+        kappa = DecorrelateALKernelTask._fixOddKernel(kappa)
+        kft = scipy.fftpack.fft2(kappa)
+        kft = np.sqrt((svar + tvar) / (svar + tvar * kft**2))
+        pck = scipy.fftpack.ifft2(kft)
+        pck = scipy.fftpack.ifftshift(pck.real)
+        fkernel = DecorrelateALKernelTask._fixEvenKernel(pck)
 
         # I think we may need to "reverse" the PSF, as in the ZOGY (and Kaiser) papers...
         # This is the same as taking the complex conjugate in Fourier space before FFT-ing back to real space.
         if False:  # TBD: figure this out. For now, we are turning it off.
-            pck = pck[::-1, :]
+            fkernel = fkernel[::-1, :]
 
-        return pck
+        return fkernel
 
     @staticmethod
-    def computeCorrectedDiffimPsf(kappa, psf, sig1=0.2, sig2=0.2):
+    def computeCorrectedDiffimPsf(kappa, psf, svar=0.04, tvar=0.04):
         """! Compute the (decorrelated) difference image's new PSF.
-        new_psf = psf(k) * sqrt((sig1**2 + sig2**2) / (sig1**2 + sig2**2 * kappa_ft(k)**2))
+        new_psf = psf(k) * sqrt((svar + tvar) / (svar + tvar * kappa_ft(k)**2))
 
         @param kappa  A matching kernel array derived from Alard & Lupton PSF matching
         @param psf    The uncorrected psf array of the science image (and also of the diffim)
-        @param sig1   Average sqrt(variance) of template image used for PSF matching
-        @param sig2   Average sqrt(variance) of science image used for PSF matching
+        @param svar   Average variance of science image used for PSF matching
+        @param tvar   Average variance of template image used for PSF matching
         @return a 2-d numpy.array containing the new PSF
         """
-        def post_conv_psf_ft2(psf, kernel, sig1=1., sig2=1.):
+        def post_conv_psf_ft2(psf, kernel, svar, tvar):
             # Pad psf or kernel symmetrically to make them the same size!
             # Note this assumes they are both square (width == height)
             if psf.shape[0] < kernel.shape[0]:
@@ -240,19 +270,40 @@ class DecorrelateALKernelTask(pipeBase.Task):
             elif psf.shape[0] > kernel.shape[0]:
                 diff = (psf.shape[0] - kernel.shape[0]) // 2
                 kernel = np.pad(kernel, (diff, diff), mode='constant')
-            psf_ft = fft2(psf)
-            kft = fft2(kernel)
-            out = psf_ft * np.sqrt((sig1**2 + sig2**2) / (sig1**2 + sig2**2 * kft**2))
+            psf_ft = scipy.fftpack.fft2(psf)
+            kft = scipy.fftpack.fft2(kernel)
+            out = psf_ft * np.sqrt((svar + tvar) / (svar + tvar * kft**2))
             return out
 
-        def post_conv_psf(psf, kernel, sig1=1., sig2=1.):
-            kft = post_conv_psf_ft2(psf, kernel, sig1, sig2)
-            out = ifft2(kft)
+        def post_conv_psf(psf, kernel, svar, tvar):
+            kft = post_conv_psf_ft2(psf, kernel, svar, tvar)
+            out = scipy.fftpack.ifft2(kft)
             return out
 
-        pcf = post_conv_psf(psf=psf, kernel=kappa, sig1=sig2, sig2=sig1)
+        pcf = post_conv_psf(psf=psf, kernel=kappa, svar=svar, tvar=tvar)
         pcf = pcf.real / pcf.real.sum()
         return pcf
+
+    @staticmethod
+    def _fixOddKernel(kernel):
+        """! Take a kernel with odd dimensions and make them even for FFT
+
+        @param kernel a numpy.array
+        @return a fixed kernel numpy.array. Returns a copy if the dimensions needed to change;
+        otherwise just return the input kernel.
+        """
+        # Note this works best for the FFT if we left-pad
+        out = kernel
+        changed = False
+        if (out.shape[0] % 2) == 1:
+            out = np.pad(out, ((1, 0), (0, 0)), mode='constant')
+            changed = True
+        if (out.shape[1] % 2) == 1:
+            out = np.pad(out, ((0, 0), (1, 0)), mode='constant')
+            changed = True
+        if changed:
+            out *= (np.mean(kernel) / np.mean(out))  # need to re-scale to same mean for FFT
+        return out
 
     @staticmethod
     def _fixEvenKernel(kernel):
@@ -288,12 +339,10 @@ class DecorrelateALKernelTask(pipeBase.Task):
         @note We use afwMath.convolve() but keep scipy.convolve for debugging.
         @note We re-center the kernel if necessary and return the possibly re-centered kernel
         """
-        fkernel = DecorrelateALKernelTask._fixEvenKernel(kernel)
-
-        kernelImg = afwImage.ImageD(fkernel.shape[0], fkernel.shape[1])
-        kernelImg.getArray()[:, :] = fkernel
+        kernelImg = afwImage.ImageD(kernel.shape[0], kernel.shape[1])
+        kernelImg.getArray()[:, :] = kernel
         kern = afwMath.FixedKernel(kernelImg)
-        maxloc = np.unravel_index(np.argmax(fkernel), fkernel.shape)
+        maxloc = np.unravel_index(np.argmax(kernel), kernel.shape)
         kern.setCtrX(maxloc[0])
         kern.setCtrY(maxloc[1])
         outExp = exposure.clone()  # Do this to keep WCS, PSF, masks, etc.
--- a/tests/testImageDecorrelation.py
+++ b/tests/testImageDecorrelation.py
@@ -32,7 +32,12 @@ import lsst.meas.algorithms as measAlg
 
 from lsst.ip.diffim.imageDecorrelation import DecorrelateALKernelTask
 
-def singleGaussian2d(x, y, xc, yc, sigma_x=1., sigma_y=1., theta=0.):
+
+def setup_module(module):
+    lsst.utils.tests.init()
+
+
+def singleGaussian2d(x, y, xc, yc, sigma_x=1., sigma_y=1., theta=0., ampl=1.):
     """! Generate a 2-d Gaussian, possibly elongated and rotated, on a grid of pixel
     coordinates given by x,y.
     @param x,y each a 1-d numpy.array containing x- and y- coordinates for independent variables,
@@ -40,7 +45,10 @@ def singleGaussian2d(x, y, xc, yc, sigma_x=1., sigma_y=1., theta=0.):
     @param xc,yc each a float giving the centroid of the gaussian
     @param sigma_x,sigma_y each a float giving the sigma of the gaussian
     @param theta a float giving the rotation of the gaussian (degrees)
+    @param ampl a float giving the amplitude of the gaussian
     @return a 2-d numpy.array containing the normalized 2-d Gaussian
+
+    @Note this can be done in `astropy.modeling` but for now we have it explicitly here.
     """
     theta = (theta/180.) * np.pi
     cos_theta2, sin_theta2 = np.cos(theta)**2., np.sin(theta)**2.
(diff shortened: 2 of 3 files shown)


python/lsst/ip/diffim/imageDecorrelation.py
```diff
@@ -130,19 +129,28 @@ def __init__(self, *args, **kwargs):
                                                                      "NO_DATA", "DETECTED_NEGATIVE"]))
 
     @pipeBase.timeMethod
-    def run(self, templateExposure, exposure, subtractedExposure, psfMatchingKernel):
+    def run(self, exposure, templateExposure, subtractedExposure, psfMatchingKernel,
+            xcen=None, ycen=None, var1=None, var2=None):
         """! Perform decorrelation of an image difference exposure.
```
parejkoj: Might be handy to note in the docstring that exposure, template match equation whatever in some paper.

python/lsst/ip/diffim/imageDecorrelation.py
```diff
@@ -164,74 +172,100 @@ def run(self, templateExposure, exposure, subtractedExposure, psfMatchingKernel)
...
+                self.log.info("Using matching kernel computed at (%d, %d)" % (xcen, ycen))
                 spatialKernel.computeImage(kimg, True, xcen, ycen)
-        except:  # Not a spatially-varying kernel (usually not true)
-            kimg = psfMatchingKernel.computeImage()
 
-        # Compute the images' sigmas (sqrt of variance)
-        statObj = afwMath.makeStatistics(templateExposure.getMaskedImage().getVariance(), afwMath.MEANCLIP,
-                                         self.statsControl)
-        sig1 = np.sqrt(statObj.getValue(afwMath.MEANCLIP))
+                if False:  # debug code to save spatially varying kernel for analysis
```
parejkoj: Please do this with an explicit debugging flag instead of `if False`. `if [True|False]` is dangerous in many ways.

python/lsst/ip/diffim/imageDecorrelation.py
```diff
@@ -130,19 +129,28 @@ def __init__(self, *args, **kwargs):
                                                                      "NO_DATA", "DETECTED_NEGATIVE"]))
```
PaulPrice: The list of mask planes should be configurable rather than hard-coded.  It just needs a `ListField` in the `Config`.

python/lsst/ip/diffim/imageDecorrelation.py
```diff
@@ -164,74 +172,100 @@ def run(self, templateExposure, exposure, subtractedExposure, psfMatchingKernel)
...
-        sig1 = np.sqrt(statObj.getValue(afwMath.MEANCLIP))
+                if False:  # debug code to save spatially varying kernel for analysis
+                    import cPickle
+                    import gzip
+                    cPickle.dump(spatialKernel, gzip.GzipFile('spatialKernel.p.gz', 'wb'))
 
-        statObj = afwMath.makeStatistics(exposure.getMaskedImage().getVariance(), afwMath.MEANCLIP,
-                                         self.statsControl)
-        sig2 = np.sqrt(statObj.getValue(afwMath.MEANCLIP))
+        except:  # Not a spatially-varying kernel (usually not true)
```
parejkoj: You should only catch the errors that spatialKernel.computeImage can raise that are relevant to this way of dealing with the error. Catching all exceptions is dangerous.

python/lsst/ip/diffim/imageDecorrelation.py
```diff
@@ -130,19 +129,28 @@ def __init__(self, *args, **kwargs):
...
+        @param[in] templateExposure the template afwImage.Exposure used for PSF matching
         @param[in] subtractedExposure the subtracted exposure produced by
         `ip_diffim.ImagePsfMatchTask.subtractExposures()`
         @param[in] psfMatchingKernel an (optionally spatially-varying) PSF matching kernel produced
         by `ip_diffim.ImagePsfMatchTask.subtractExposures()`
+        @param[in] xcen X-pixel coordinate to use for computing constant matching kernel to use
+        If `None` (default), then use the center of the image.
+        @param[in] ycen Y-pixel coordinate to use for computing constant matching kernel to use
+        If `None` (default), then use the center of the image.
+        @param[in] var1 image variance for science image
```
parejkoj: If you're calling the images exposure and template, you should call the variances var_exposure and var_template (or something), or call the images var1 and var2. Having the names different like this makes them hard to keep track of.

python/lsst/ip/diffim/imageDecorrelation.py
```diff
@@ -164,74 +172,100 @@ def run(self, templateExposure, exposure, subtractedExposure, psfMatchingKernel)
...
-        statObj = afwMath.makeStatistics(exposure.getMaskedImage().getVariance(), afwMath.MEANCLIP,
-                                         self.statsControl)
-        sig2 = np.sqrt(statObj.getValue(afwMath.MEANCLIP))
+        except:  # Not a spatially-varying kernel (usually not true)
+            kimg = psfMatchingKernel.computeImage()
 
-        corrKernel = DecorrelateALKernelTask._computeDecorrelationKernel(kimg.getArray(), sig1=sig1, sig2=sig2)
-        fcorrKernel = DecorrelateALKernelTask._fixEvenKernel(corrKernel)
+        if var1 is None:
+            statObj = afwMath.makeStatistics(exposure.getMaskedImage().getVariance(),
```
parejkoj: You could make a short helper function that takes an Exposure and a statsControl and just returns the variance and call it in these two `if None` blocks, since they do the same thing (and I'm guessing you want them to always do the same thing).

python/lsst/ip/diffim/imageDecorrelation.py
```diff
@@ -164,74 +172,100 @@ def run(self, templateExposure, exposure, subtractedExposure, psfMatchingKernel)
...
+                                             exposure.getMaskedImage().getMask(),
+                                             afwMath.MEANCLIP, self.statsControl)
+            var1 = statObj.getValue(afwMath.MEANCLIP)
+        self.log.info("Variance (science): %f" % var1)
```
parejkoj: Does this actually log something useful, or is str(var1) some crazy c++ pointer nonsense? Would be good to have the log give something informative here.

python/lsst/ip/diffim/imageDecorrelation.py
```diff
@@ -164,74 +172,100 @@ def run(self, templateExposure, exposure, subtractedExposure, psfMatchingKernel)
...
 
     @staticmethod
-    def _computeDecorrelationKernel(kappa, sig1=0.2, sig2=0.2):
+    def _computeDecorrelationKernel(kappa, var1=0.04, var2=0.04):
         """! Compute the Lupton/ZOGY post-conv. kernel for decorrelating an
         image difference, based on the PSF-matching kernel.
         @param kappa  A matching kernel 2-d numpy.array derived from Alard & Lupton PSF matching
-        @param sig1   Average sqrt(variance) of template image used for PSF matching
-        @param sig2   Average sqrt(variance) of science image used for PSF matching
+        @param var1   Average variance of science image used for PSF matching
```
parejkoj: Same thing here: if var1==science, then call it science. Otherwise don't call anything science or template and just use 1 and 2 everywhere.

python/lsst/ip/diffim/imageDecorrelation.py
```diff
@@ -164,74 +172,100 @@ def run(self, templateExposure, exposure, subtractedExposure, psfMatchingKernel)
...
+        @param var2   Average variance of template image used for PSF matching
         @return a 2-d numpy.array containing the correction kernel
 
         @note As currently implemented, kappa is a static (single) kernel.
         """
+        kappa = DecorrelateALKernelTask._fixOddKernel(kappa)
         kft = fft2(kappa)
```
parejkoj: Though it's not explicitly part of this review, I would rather you didn't `from scipy import fft2`, but just `import scipy.fftpack`, so that it's clear where these fft-ers come from. You're not calling them so often that the reduction in characters is worth the "what is fft2? question.

python/lsst/ip/diffim/imageDecorrelation.py
```diff
@@ -164,74 +172,100 @@ def run(self, templateExposure, exposure, subtractedExposure, psfMatchingKernel)
...
-        kft = np.sqrt((sig1**2 + sig2**2) / (sig1**2 + sig2**2 * kft**2))
+        kft = np.sqrt((var1 + var2) / (var1 + var2 * kft**2))
         pck = ifft2(kft)
         pck = ifftshift(pck.real)
+        fkernel = DecorrelateALKernelTask._fixEvenKernel(pck)
 
         # I think we may need to "reverse" the PSF, as in the ZOGY (and Kaiser) papers...
         # This is the same as taking the complex conjugate in Fourier space before FFT-ing back to real space.
         if False:  # TBD: figure this out. For now, we are turning it off.
-            pck = pck[::-1, :]
```
parejkoj: Another if False. Please just comment out the block, with the comment as to why above it.

python/lsst/ip/diffim/imageDecorrelation.py
```diff
@@ -164,74 +172,100 @@ def run(self, templateExposure, exposure, subtractedExposure, psfMatchingKernel)
...
         @param kappa  A matching kernel array derived from Alard & Lupton PSF matching
         @param psf    The uncorrected psf array of the science image (and also of the diffim)
-        @param sig1   Average sqrt(variance) of template image used for PSF matching
-        @param sig2   Average sqrt(variance) of science image used for PSF matching
+        @param var1   Average variance of science image used for PSF matching
+        @param var2   Average variance of template image used for PSF matching
         @return a 2-d numpy.array containing the new PSF
         """
-        def post_conv_psf_ft2(psf, kernel, sig1=1., sig2=1.):
+        def post_conv_psf_ft2(psf, kernel, var1=1., var2=1.):
```
parejkoj: Why are these defaults 1 vs. 0.04 as above?

python/lsst/ip/diffim/imageDecorrelation.py
```diff
@@ -242,19 +276,34 @@ def post_conv_psf_ft2(psf, kernel, sig1=1., sig2=1.):
...
 
-        def post_conv_psf(psf, kernel, sig1=1., sig2=1.):
-            kft = post_conv_psf_ft2(psf, kernel, sig1, sig2)
+        def post_conv_psf(psf, kernel, var1=1., var2=1.):
+            kft = post_conv_psf_ft2(psf, kernel, var1, var2)
             out = ifft2(kft)
             return out
 
-        pcf = post_conv_psf(psf=psf, kernel=kappa, sig1=sig2, sig2=sig1)
+        pcf = post_conv_psf(psf=psf, kernel=kappa, var1=var1, var2=var2)
```
parejkoj: This flip-flopped before. Just checking that you don't want it flip-flopped this time.

python/lsst/ip/diffim/imageDecorrelation.py
djreiss: I agree. However, I got rid of this thanks to a prior review.

python/lsst/ip/diffim/imageDecorrelation.py
djreiss: Yes, it's extremely useful -- it's a single float. Running the unit test will show it (`python tests/testImageDecorrelation.py`).

python/lsst/ip/diffim/imageDecorrelation.py
```diff
@@ -242,19 +276,34 @@ def post_conv_psf_ft2(psf, kernel, sig1=1., sig2=1.):
...
         pcf = pcf.real / pcf.real.sum()
         return pcf
 
     @staticmethod
+    def _fixOddKernel(kernel):
+        """! Take a kernel with odd dimensions and make them even for FFT
+        @param kernel a numpy.array
+        @return a fixed kernel numpy.array
+        """
+        # Note this works best if we left-pad
```
parejkoj: Maybe say why left pad is better than right?

python/lsst/ip/diffim/imageDecorrelation.py
```diff
@@ -242,19 +276,34 @@ def post_conv_psf_ft2(psf, kernel, sig1=1., sig2=1.):
...
             out = ifft2(kft)
             return out
 
-        pcf = post_conv_psf(psf=psf, kernel=kappa, sig1=sig2, sig2=sig1)
+        pcf = post_conv_psf(psf=psf, kernel=kappa, var1=var1, var2=var2)
         pcf = pcf.real / pcf.real.sum()
         return pcf
 
     @staticmethod
+    def _fixOddKernel(kernel):
```
parejkoj: Not sure this should be a static method: it looks more like a stand-alone, top-level function.

python/lsst/ip/diffim/imageDecorrelation.py
```diff
@@ -242,19 +276,34 @@ def post_conv_psf_ft2(psf, kernel, sig1=1., sig2=1.):
...
+        """! Take a kernel with odd dimensions and make them even for FFT
+        @param kernel a numpy.array
+        @return a fixed kernel numpy.array
+        """
+        # Note this works best if we left-pad
+        out = kernel
```
parejkoj: It looks like the goal is to not modify the kernel in place, so you probably want `copy()` or `deepcopy()`here.

python/lsst/ip/diffim/imageDecorrelation.py
```diff
@@ -242,19 +276,34 @@ def post_conv_psf_ft2(psf, kernel, sig1=1., sig2=1.):
...
+        if (out.shape[0] % 2) == 1:
+            out = np.pad(out, ((1,0), (0,0)), mode='constant')
+        if (out.shape[1] % 2) == 1:
+            out = np.pad(out, ((0,0), (1,0)), mode='constant')
+        out *= (np.mean(kernel) / np.mean(out))  # need to re-scale to same mean for FFT
+        return out
+
+    @staticmethod
```
parejkoj: Another unlikely static method that should just be a top level function.

tests/testImageDecorrelation.py
```diff
@@ -54,25 +55,26 @@ def singleGaussian2d(x, y, xc, yc, sigma_x=1., sigma_y=1., theta=0.):
...
 
 
-def makeFakeImages(xim=None, yim=None, sig1=0.2, sig2=0.2, psf1=2.2, psf2=3.3, offset=None,
-                   psf_yvary_factor=0., varSourceChange=1/50., theta1=0., theta2=0., im2background=0.,
+def makeFakeImages(xim=None, yim=None, var1=0.04, var2=0.04, psf1=3.3, psf2=2.2, offset=None,
+                   psf_yvary_factor=0., varSourceChange=1/50., theta1=0., theta2=0., im1background=0.,
                    n_sources=500, seed=66, verbose=False):
     """! Make two exposures: a template and a science exposure.
     Add random sources of identical flux, with randomly-distributed fluxes and a given PSF, then add noise.
+    In all cases below, index (1) is the science image, and (2) is the template.
```
parejkoj: I'd suggest using "science" and "template" everywhere in the code to make that explicit. They're not symmetric, so don't treat them like they are.

tests/testImageDecorrelation.py
```diff
@@ -84,17 +86,17 @@ def makeFakeImages(xim=None, yim=None, sig1=0.2, sig2=0.2, psf1=2.2, psf2=3.3, o
     """
     np.random.seed(seed)
 
-    psf1 = [2.2, 2.2] if psf1 is None else psf1
+    psf1 = [3.3, 3.3] if psf1 is None else psf1
```
parejkoj: This won't do what the docstring says, since the default values are numbers instead of None.

tests/testImageDecorrelation.py
```diff
@@ -159,8 +161,8 @@ def makeExposure(imgArray, psfArray, imgSigma):
         im1ex.setPsf(psfNew)
         return im1ex
 
-    im1ex = makeExposure(im1, im1_psf, sig1)
-    im2ex = makeExposure(im2, im2_psf, sig2)
+    im1ex = makeExposure(im1, im1_psf, var1)  # Science image
```
parejkoj: I may be a broken record, but another "science/template" replacement here.

tests/testImageDecorrelation.py
```diff
@@ -54,25 +55,26 @@ def singleGaussian2d(x, y, xc, yc, sigma_x=1., sigma_y=1., theta=0.):
...
     Add random sources of identical flux, with randomly-distributed fluxes and a given PSF, then add noise.
+    In all cases below, index (1) is the science image, and (2) is the template.
     @param xim,yim image pixel coordinates on which to generate the image grid. Default is (-256:256).
-    @param sig1,sig2 std. dev. of noise to be generated on input images. Defalt is 0.2 for both.
+    @param var1,var2 variance of noise to be generated on input images. Default is 0.04 for both.
     @param psf1,psf2 std. dev. of (Gaussian) PSFs for the two images in x,y direction. Default is
-    [2.2, 2.2] and [3.3, 3.3] for im1 and im2 respectively.
+    [3.3, 3.3] and [2.2, 2.2] for im1 and im2 respectively.
     @param offset add a constant (pixel) astrometric offset between the two images
     @param psf_yvary_factor vary the PSF of the science image by this much across the image
```
parejkoj: Please note that this variation is only in the y-direction (which will make the name be clear).

tests/testImageDecorrelation.py
```diff
@@ -190,65 +182,124 @@ def setUp(self):
                                                                      "DETECTED", "BAD",
                                                                      "NO_DATA", "DETECTED_NEGATIVE"]))
 
+    def _setUpImages(self, var1=0.04, var2=0.04):
+        """!Generate a fake aligned template and science image and analyse the noise in the
+        resulting diffim. First, use a non-spatially-varying psf.
+        """
+
+        self.var1 = var1  # variance of noise in science image
```
parejkoj: science/template again.

tests/testImageDecorrelation.py
```diff
@@ -190,65 +182,124 @@ def setUp(self):
                                                                      "DETECTED", "BAD",
                                                                      "NO_DATA", "DETECTED_NEGATIVE"]))
 
+    def _setUpImages(self, var1=0.04, var2=0.04):
+        """!Generate a fake aligned template and science image and analyse the noise in the
+        resulting diffim. First, use a non-spatially-varying psf.
```
parejkoj: I feel like this docstring doesn't match what the method does.

tests/testImageDecorrelation.py
```diff
@@ -190,65 +182,124 @@ def setUp(self):
...
+    def _testImages(self):
         # Create the matching kernel. We used Gaussian PSFs for im1 and im2, so we can compute the "expected"
         # matching kernel sigma.
         psf1_sig = self.im1ex.getPsf().computeShape().getDeterminantRadius()
         psf2_sig = self.im2ex.getPsf().computeShape().getDeterminantRadius()
-        sig_match = np.sqrt((psf2_sig**2. - psf1_sig**2.))
-        self.assertClose(sig_match, np.sqrt((self.psf2_sigma**2. - self.psf1_sigma**2.)), rtol=1e-5)
-        #matchingKernel = measAlg.SingleGaussianPsf(31, 31, sig_match)
+        sig_match = np.sqrt((psf1_sig**2. - psf2_sig**2.))
+        self.assertClose(sig_match, np.sqrt((self.psf1_sigma**2. - self.psf2_sigma**2.)), rtol=1e-5)
```
parejkoj: Why would this ever fail? Can the psf in self.im1ex be different from the input psf?

tests/testImageDecorrelation.py
```diff
@@ -190,65 +182,124 @@ def setUp(self):
...
+        # mKernel = measAlg.SingleGaussianPsf(31, 31, sig_match)
         x0 = np.arange(-16, 16, 1)
         y0 = x0.copy()
         x0im, y0im = np.meshgrid(x0, y0)
-        matchingKernel = singleGaussian2d(x0im, y0im, 0., 0., sigma_x=sig_match, sigma_y=sig_match)
+        matchingKernel = singleGaussian2d(x0im, y0im, -1., -1., sigma_x=sig_match, sigma_y=sig_match)
```
parejkoj: See my forthcoming comment about astropy.modeling's Gaussian2D.

tests/testImageDecorrelation.py
```diff
@@ -190,65 +182,124 @@ def setUp(self):
...
+        expected_var = self.var1 + self.var2
         print('Expected variance:', expected_var)
 
-        im2 = self.im2ex.getMaskedImage().getImage().getArray()
-        print(np.nan_to_num(matched_im1 - im2).var())
+        statObj = afwMath.makeStatistics(self.im1ex.getMaskedImage(), afwMath.VARIANCECLIP,
+                                         self.statsControl)
+        var = statObj.getValue(afwMath.VARIANCECLIP)
+        statObj = afwMath.makeStatistics(self.im2ex.getMaskedImage(), afwMath.VARIANCECLIP,
+                                         self.statsControl)
```
parejkoj: Name collision here between statObj and var for im1 and im2, and again a few lines further below. Do you actually need to get these variances here? They aren't used.

tests/testImageDecorrelation.py
```diff
@@ -190,65 +182,124 @@ def setUp(self):
...
+        var = statObj.getValue(afwMath.VARIANCECLIP)
+
+        im1 = self.im1ex.getMaskedImage().getImage().getArray()
+        matched_im2 = matched_im2ex.getMaskedImage().getImage().getArray()
         # Uncorrected diffim - variance is wrong (too low)
-        self.assertNotClose(np.nan_to_num(matched_im1 - im2).var(), expected_var, rtol=0.1)
-        # In fact, it should have an (incorrect) variance that is close to the variance of the science image.
-        self.assertClose((matched_im1 - im2)[~np.isnan(matched_im1 - im2)].var(), self.sig2**2., rtol=0.1)
+        self.assertNotClose(np.nan_to_num(im1 - matched_im2).var(), expected_var, rtol=0.1)
```
parejkoj: Can you quantify this test better? Instead of being "not close" is it that the first value is much less than the second, or something? That would make the test more robust.

tests/testImageDecorrelation.py
```diff
@@ -190,65 +182,124 @@ def setUp(self):
...
+
+        # Create the diffim (uncorrected)
+        diffExp = self.im1ex.clone()
+        tmp = diffExp.getMaskedImage()
+        tmp -= matched_im2ex.getMaskedImage()
+        # Uncorrected diffim exposure - variance is wrong (too low) - same as above but on exposure
+        statObj = afwMath.makeStatistics(diffExp.getMaskedImage(), afwMath.VARIANCECLIP,
+                                         self.statsControl)
+        var = statObj.getValue(afwMath.VARIANCECLIP)
+        self.assertNotClose(var, expected_var, rtol=0.02)
```
parejkoj: Again: can you quantify the test here better, so it's not just that those "aren't close?"

tests/testImageDecorrelation.py
```diff
@@ -190,65 +182,124 @@ def setUp(self):
...
+        statObj = afwMath.makeStatistics(corrected_diffExp.getMaskedImage(), afwMath.VARIANCECLIP,
                                          self.statsControl)
-        var = statObj.getValue(afwMath.VARIANCE)
-        print(var, expected_var)
-        self.assertClose(var, expected_var, rtol=0.02)
+        var = statObj.getValue(afwMath.VARIANCECLIP)
+        self.assertClose(var, expected_var, rtol=0.05)
+
+        # Check statistics of variance plane in corrected diffim
+        statObj = afwMath.makeStatistics(corrected_diffExp.getMaskedImage().getVariance(),
```
parejkoj: Seeing all of these makeStatistics calls, I feel like you want a couple helper functions in ImageDecorrelation.py that take MaskedImages and return the mean, variance, etc. Would clean this up a fair bit, since there's a lot of boilerplate here.

tests/testImageDecorrelation.py
```diff
@@ -190,65 +182,124 @@ def setUp(self):
...
+                                         corrected_diffExp.getMaskedImage().getMask(),
+                                         afwMath.MEANCLIP, self.statsControl)
+        mn = statObj.getValue(afwMath.MEANCLIP)
+        print('CORRECTED VARIANCE:', var, mn)
+        self.assertClose(mn, expected_var, rtol=0.05)
```
parejkoj: These tests are good: the values should be close, that's it.

tests/testImageDecorrelation.py
```diff
@@ -190,65 +182,124 @@ def setUp(self):
...
+        self.assertClose(var, mn, rtol=0.05)
+
+    def testDiffimCorrection1(self):
+        """! Test two images with the same variance.
```
parejkoj: Do we want doxygen in our test methods? The test docstrings are used to help debug why a test failed (and aren't really part of the overall "published" documentation), so I think all you want for these are the first lines (without the leading !) of each. The other lines don't tell me anything, and belong in the _testImages() docstring.

tests/testImageDecorrelation.py
```diff
@@ -190,65 +182,124 @@ def setUp(self):
...
         del self.im1ex
         del self.im2ex
 
-    def testDiffimCorrection(self):
-        """! Check that the variance of the corrected diffim matches the theoretical value
-        (to within a 1% tolerance).
-        """
-
+    def _testImages(self):
         # Create the matching kernel. We used Gaussian PSFs for im1 and im2, so we can compute the "expected"
```
parejkoj: Though it's a "private" method, it would be good to have a short docstring here that says what gets tested.

tests/testImageDecorrelation.py
```diff
@@ -190,65 +182,124 @@ def setUp(self):
...
+    def testDiffimCorrection1(self):
+        """! Test two images with the same variance.
+
+        Check that the variance of the corrected diffim matches the theoretical value
+        (to within a 1% tolerance). Use
+        """
+        self._setUpImages(var1=0.04, var2=0.04)
+        self._testImages()
+
+    def testDiffimCorrection2(self):
```
parejkoj: Name the methods something useful so that the test output tells the user what the test does, e.g. "testImageDecorrelation_same_variance", "testImageDecorrelation_higher_science_variance", "testImageDecorrelation_higher_template_variance"

parejkoj:
Few things, besides the specific comments I had on lines (some of which are expanded on below):

* You can replace your singleGaussian2d() with astropy.modeling's Gaussian2D as follows (check that it doesn't break anything). Plus, since it's an object, you can instantiate it once and just call it in the loop as needed, which will make the intention of that loop clearer.
```
x,y = np.meshgrid(np.linspace(blah), np.linspace(blah))
someGaussian2d = astropy.modeling.functional_models.Gaussian2D(1,0,2,1,2,np.pi/4)
someGaussian2d(x,y)
```
* Please double-check your linter config: a few off-standard lines have slipped through (not necessarily in the reviewed lines).
* If you pull those staticmethods out to top level functions, you can then easily write a unittest for each, to help ensure they're e.g. padding how you want them to.
* I probably shouldn't be surprised that afw.math.makeStatistics doesn't have a non-verbose way to get a variance, mean, etc. from a maskedImage. Given that, you can cut out a lot of code duplication by writing a top level imageDecorrelation functions to return the clipped-mean variance of a MaskedImage, and maybe another for one of the other statistics you're always computing.
* Your tests would be more robust if you better quantified what the test was (e.g. `var-expected_var > 10` or `var-expected_var < 10`), since assertNotClose covers quite a broad range of possibilities.

tests/testImageDecorrelation.py
```diff
@@ -190,65 +182,124 @@ def setUp(self):
...
+        """! Test when the template variance is higher than that of the science img.
+
+        Check that the variance of the corrected diffim matches the theoretical value
+        (to within a 1% tolerance).
+        """
+        self._setUpImages(var1=0.04, var2=0.08)
+        self._testImages()
 
 
 def suite():
```
parejkoj: TestSuite is dead: long live py.test! See SQR-012 and the developer docs for more details:

https://sqr-012.lsst.io/
https://developer.lsst.io/coding/python_testing.html

python/lsst/ip/diffim/imageDecorrelation.py
PaulPrice: Standards have long been that `if False` is preferred to commenting out code.

parejkoj: Whose standards? It's very unpythonic, and can be a cause of confusion (commented out code shows up differently in an editor, whereas "if False" just looks like any other code). I couldn't find anything in the coding standards about not commenting out blocks.

I even have a community post about this:

https://community.lsst.org/t/use-of-if-true-if-false-in-python-and-if-0-in-c/455/3

PaulPrice: It's a pythonification of [a C++ standard](https://dev.lsstcorp.org/trac/wiki/C%2B%2BStandard/Statements#a5-38.NocodeSHOULDbecommentedoutuseapreprocessordirectivetoincludeorinhibitcodeuse.).  Granted that we don't actually have a corresponding python standard, but 'we' have long interpreted this as blessing `if False`.

parejkoj: I understand where it comes from in C, but I **very strongly** advocate that we stop using it in python, especially since `if False:` and `#if 0` actually have quite different behavior (interpreted vs. compiled language). See my community post above for details (and we can continue the conversation there, to not clutter David's PR).

python/lsst/ip/diffim/imageDecorrelation.py
djreiss: It's not in a paper, yet. But I will reference the (still in progress) DMTN-021 with those details.

python/lsst/ip/diffim/imageDecorrelation.py
djreiss: Keeping as-is following your discussion with Paul.

python/lsst/ip/diffim/imageDecorrelation.py
djreiss: OK, done.

python/lsst/ip/diffim/imageDecorrelation.py
djreiss: This is not to catch an exception by the method, but essentially to catch if the method `isSpatiallyVarying` exists for the kernel class. So, I will catch `AttributeError`.

python/lsst/ip/diffim/imageDecorrelation.py
djreiss: Very well, I will call the variances `svar` and `tvar` for science and template.

python/lsst/ip/diffim/imageDecorrelation.py
djreiss: Done. See above.

python/lsst/ip/diffim/imageDecorrelation.py
djreiss: OK.

python/lsst/ip/diffim/imageDecorrelation.py
djreiss: No reason. Removed the defaults.

python/lsst/ip/diffim/imageDecorrelation.py
djreiss: Thanks. It's correct.

python/lsst/ip/diffim/imageDecorrelation.py
djreiss: Um, right-padding won't move the centroid.

python/lsst/ip/diffim/imageDecorrelation.py
djreiss: I'm going to go with the philosophy of this comment: http://stackoverflow.com/a/16079946

Note that a subclass may want to override some of these methods and I want to make that an explicit choice.

tests/testImageDecorrelation.py
djreiss: Thanks for this. 
I'll go with this discussion: https://stackoverflow.com/a/19773559
Keeping the `hasattr` call and adding the `basestring` check.

tests/testImageDecorrelation.py
djreiss: OK, updated docstring.

tests/testImageDecorrelation.py
djreiss: Thanks, fixed.

tests/testImageDecorrelation.py
djreiss: Just a sanity check. If this fails then the rest of the tests will fail.

tests/testImageDecorrelation.py
djreiss: You're right, thanks. They were leftover from previously removed tests.

tests/testImageDecorrelation.py
djreiss: Yes, it is much less than. Is there an `assertMuchLessThan`?

tests/testImageDecorrelation.py
djreiss: Done.

tests/testImageDecorrelation.py
djreiss: OK.

tests/testImageDecorrelation.py
djreiss: OK

tests/testImageDecorrelation.py
djreiss: OK, converted.

python/lsst/ip/diffim/imageDecorrelation.py
parejkoj: I see the reference. Thanks. Maybe I_1 and I_2 should be I_s and I_t in DMTN-021, to make it explicit?

python/lsst/ip/diffim/imageDecorrelation.py
parejkoj: Good. That's much better now!

python/lsst/ip/diffim/imageDecorrelation.py
parejkoj: Thanks.

python/lsst/ip/diffim/imageDecorrelation.py
parejkoj: Ok, good.

tests/testImageDecorrelation.py
parejkoj: Clearer, thanks.

tests/testImageDecorrelation.py
```diff
@@ -178,13 +190,12 @@ def setUp(self):
         self.statsControl = afwMath.StatisticsControl()
         self.statsControl.setNumSigmaClip(3.)
         self.statsControl.setNumIter(3)
-        self.statsControl.setAndMask(afwImage.MaskU.getPlaneBitMask(["INTRP", "EDGE",
+        self.statsControl.setAndMask(afwImage.MaskU.getPlaneBitMask(["INTRP", "EDGE", "SAT", "CR",
                                                                      "DETECTED", "BAD",
                                                                      "NO_DATA", "DETECTED_NEGATIVE"]))
 
     def _setUpImages(self, var1=0.04, var2=0.04):
-        """!Generate a fake aligned template and science image and analyse the noise in the
```
parejkoj: Please make these svar and tvar everywhere in the test, too.

tests/testImageDecorrelation.py
parejkoj: I think that's worth an inline "sanity check" comment.

tests/testImageDecorrelation.py
```diff
@@ -194,11 +205,26 @@ def _setUpImages(self, var1=0.04, var2=0.04):
             = makeFakeImages(var1=self.var1, var2=self.var2, psf1=self.psf1_sigma, psf2=self.psf2_sigma,
                              n_sources=50, verbose=False)
 
+    def _computeVarianceMean(self, maskedIm):
```
parejkoj: You can just use computeVarianceMean from imageDecorrelation.py, no?

tests/testImageDecorrelation.py
```diff
@@ -223,100 +249,64 @@ def _testImages(self):
...
+        var = self._computePixelVariance(diffExp.getMaskedImage())
+        self.assertLess(var, expected_var)
 
         # Uncorrected diffim exposure - variance plane is wrong (too low)
-        statObj = afwMath.makeStatistics(diffExp.getMaskedImage().getVariance(),
-                                         diffExp.getMaskedImage().getMask(), afwMath.MEANCLIP,
-                                         self.statsControl)
-        mn = statObj.getValue(afwMath.MEANCLIP)
+        mn = self._computeVarianceMean(diffExp.getMaskedImage())
         self.assertNotClose(mn, expected_var, rtol=0.02)
```
parejkoj: Another "assertNotClose" that should probably be "assertLess".

tests/testImageDecorrelation.py
parejkoj: Thanks.

tests/testImageDecorrelation.py
djreiss: Done.

tests/testImageDecorrelation.py
djreiss: It uses a pre-made statsControl object, so not static.

tests/testImageDecorrelation.py
djreiss: Done.

PR #27, opened 2016-08-04
DM-6999: Update for the logging changes in the task framework

hchiang2:
Tasks' logs are switched from lsst.pex.logging to lsst.log


r-owen:
In two places you changed `self.log.logdebug` to `self.log.trace`. Why the change in logging level? Otherwise this looks great.

hchiang2:
They have been changed to `self.log.debug`. (I guess you were looking at a slightly outdated branch and I only corrected them a few days ago)

PR #29, opened 2016-08-23
Tickets/dm 7320

fred3m:


parejkoj:
Please do a quick pass and remove all the `#-=-=-=` lines.

parejkoj:
testDipole still has a bunch of assertTrue that can be cleaned up. Also, please move the np.random.seed into the class setUp.

parejkoj:
Please also remove empty setUp and tearDown (those that just have `pass`).

parejkoj:
There's a bunch of bad handling of dataDir that should be turned into `@unittest.skipIf()`, but I'll file a ticket specifically for that.

fred3m:
Makes sense, I'll make the indicated changes and update the JIRA ticket.

timj:
Proper skipping is an important part of pytest upgrades. It really helps a lot to see the skips explicitly in the results rather than a test module just exiting early.

PR #31, opened 2016-08-31
Fix NaiveDipoleCentroid logic

AstroVPK:
Only fit dipoles to footprints with a pos. and neg. peak

remove debug statements; add other default plugins

### DIFF
--- a/python/lsst/ip/diffim/dipoleFitTask.py
+++ b/python/lsst/ip/diffim/dipoleFitTask.py
@@ -46,6 +46,11 @@ class DipoleFitPluginConfig(measBase.SingleFramePluginConfig):
     """!Configuration for DipoleFitPlugin
     """
 
+    fitAllDiaSources = pexConfig.Field(
+        dtype=float, default=False,
+        doc="""Attempte dipole fit of all diaSources (otherwise just the ones consisting of overlapping
+        positive and negative footprints)""")
+
     maxSeparation = pexConfig.Field(
         dtype=float, default=5.,
         doc="Assume dipole is not separated by more than maxSeparation * psfSigma")
@@ -102,10 +107,15 @@ class DipoleFitTaskConfig(measBase.SingleFrameMeasurementConfig):
                               "base_PixelFlags",
                               "base_SkyCoord",
                               "base_PsfFlux",
+                              "base_GaussianCentroid",
+                              "base_PeakLikelihoodFlux",
+                              "base_PeakCentroid",
+                              "base_SdssCentroid",
+                              "base_NaiveCentroid",
                               "ip_diffim_NaiveDipoleCentroid",
                               "ip_diffim_NaiveDipoleFlux",
                               "ip_diffim_PsfDipoleFlux",
-                              "ip_diffim_ClassificationDipole"
+                              "ip_diffim_ClassificationDipole",
                               ]
 
         self.slots.calibFlux = None
@@ -538,10 +548,11 @@ class DipoleFitAlgorithm(object):
         cenNeg = cenPos = fpCentroid
 
         pks = fp.getPeaks()
+
         if len(pks) >= 1:
             cenPos = pks[0].getF()    # if individual (merged) peaks were detected, use those
-        if len(pks) >= 2:
-            cenNeg = pks[1].getF()
+        if len(pks) >= 2:    # peaks are already sorted by centroid flux so take the most negative one
+            cenNeg = pks[-1].getF()
 
         # For close/faint dipoles the starting locs (min/max) might be way off, let's help them a bit.
         # First assume dipole is not separated by more than 5*psfSigma.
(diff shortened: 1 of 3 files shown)


python/lsst/ip/diffim/dipoleFitTask.py
```diff
@@ -102,10 +107,16 @@ def setDefaults(self):
                               "base_PixelFlags",
                               "base_SkyCoord",
                               "base_PsfFlux",
+                              "base_GaussianCentroid",
+                              #"base_GaussianFlux",
```
AstroVPK: This should be removed if it's not required.

src/DipoleAlgorithms.cc
```diff
@@ -137,34 +137,54 @@ void NaiveDipoleCentroid::measure(
...
     afw::detection::PeakCatalog const& peaks = source.getFootprint()->getPeaks();
 
-    naiveCentroid(source, exposure, peaks[0].getI(), (peaks[0].getPeakValue() >= 0 ?
-                                                       getPositiveKeys() :
-                                                       getNegativeKeys()));
-    if (peaks.size() > 1) {
-        naiveCentroid(source, exposure, peaks[1].getI(), (peaks[1].getPeakValue() >= 0 ?
-                                                           getPositiveKeys() :
-                                                           getNegativeKeys()));
+    /*
```
AstroVPK: This block of commented out code should be removed. If you want to keep the comemnted out code, maybe make a copy in a user branch?

src/DipoleAlgorithms.cc
```diff
@@ -137,34 +137,54 @@ void NaiveDipoleCentroid::measure(
...
+    }
+    */
+
+    int posInd = 0;
+    double posValue = peaks[posInd].getPeakValue(), negValue = 0;
+    if (posValue < 0.) { /* All peaks are negative so use the *most* negative value */
+       posInd = peaks.size() - 1;
+       posValue = peaks[posInd].getPeakValue();
+    }
+    //std::cout << "   HEREXXX: " << posValue << std::endl;
```
AstroVPK: Again, this should be removed.

src/DipoleAlgorithms.cc
```diff
@@ -137,34 +137,54 @@ void NaiveDipoleCentroid::measure(
...
+    naiveCentroid(source, exposure, peaks[posInd].getI(),
+                 (posValue >= 0 ? getPositiveKeys() : getNegativeKeys()));
+
+    if (posValue > 0. && posInd == 0 && peaks.size() > 1) { /* See if there's also a negative peak */
+       int negInd = peaks.size() - 1;
+       negValue = peaks[negInd].getPeakValue();
+       if (posValue > 0. && negValue < 0.) {
+          //std::cout << "         HERE" << std::endl;
```
AstroVPK: And here.

src/DipoleAlgorithms.cc
RobertLuptonTheGood: Please don't create a branch for this.  This looks like a classic case for using Trace logging ( @hchiang2 )

python/lsst/ip/diffim/dipoleFitTask.py
djreiss: My bad. I completely forgot to remove the commented-out debug code before pushing. Thanks for noting it -- I'll remove all of the comments that you've identified.

djreiss: base_GaussianFlux was commented out because it leads to a schema set-up failure:
```
Alias for 'slot_Shape_flag' must be defined before initializing 'base_GaussianFlux' plugin. {0}
lsst::pex::exceptions::LogicError: 'Alias for 'slot_Shape_flag' must be defined before initializing 'base_GaussianFlux' plugin.'
```
I'll remove this line now, but I've filed a ticket (DM-7458), as I would like to add base_GaussianFlux to the default list of plugins.

TallJimbo: To run `base_GaussianFlux`, you'll need to run a shape algorithm (e.g. `base_SdssShape`), and set the shape slot to point to it (`...config.measurement.slots.shape = 'base_SdssShape'`).

djreiss: Thanks, @TallJimbo -- this should be an easy fix.

src/DipoleAlgorithms.cc
hchiang2: Sorry I didn't get a chance to look carefully before this is rewritten.  Converting the debug logging in C++ to use `lsst.log` is still work in progress so I don't have a good example in a stable place yet, but if you don't mind, this might be an example: https://github.com/lsst/afw/blob/4afd3c77e952ac24efa75f5ac87e5d04f2f4ebe4/src/detection/PsfFormatter.cc    (No promise this commit will stay long, sorry!)

Get a logger `LOG_LOGGER _log = LOG_GET("ip.diffim.DipoleAlgorithms");` and then log to it  `LOGL_DEBUG(_log, "messages")`     For more details, please see [RFC-203](https://jira.lsstcorp.org/browse/RFC-203?focusedCommentId=52671&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-52671) and later those descriptions will be added to some formal documentation.

PR #37, opened 2016-10-17
DM-8002: Separate display logic from algorithm logic.

hsinfang:
This split in logic is necessary to guard the call
to ds9.Buffering() with an `if display` statement,
but it's also cleaner.


python/lsst/ip/diffim/diaCatalogSourceSelector.py
```diff
@@ -201,43 +201,56 @@ def selectStars(self, exposure, sourceCat, matches=None):
         #
         starCat = SourceCatalog(sourceCat.schema)
 
+        if display:
```
hsinfang: From below it looks to me that `if display and displayExposure:` would be enough?

TallJimbo: Correct.  Will adjust accordingly.

PR #38, opened 2016-11-14
DM-7000: Convert tracing to use lsst.log and remove pex.logging

hsinfang:


hsinfang:
Most of the examples do not run, but `examples/snapPsfMatchTask.py` does. I pasted its outputs (before and after this patch) to Jira. 

Yes, the outputs are similar with the minor format differences.

PR #43, opened 2017-01-06
Wrap ip_diffim using pybind11: tickets/DM-8460

r-owen:


python/lsst/ip/diffim/detail/buildSingleKernelVisitor.cc
```diff
@@ -0,0 +1,98 @@
...
+    cls.def("reset", &BuildSingleKernelVisitor<PixelT>::reset);
+    cls.def("processCandidate", &BuildSingleKernelVisitor<PixelT>::processCandidate, "candidate"_a);
+
+    mod.def("makeBuildSingleKernelVisitor", (std::shared_ptr<BuildSingleKernelVisitor<PixelT>>(*)(
+                                                afw::math::KernelList const&, pex::policy::Policy const&)) &
+                                                makeBuildSingleKernelVisitor<PixelT>,
+            "basisList"_a, "policy"_a);
+    mod.def("makeBuildSingleKernelVisitor",
+            (std::shared_ptr<BuildSingleKernelVisitor<PixelT>>(*)(
+                afw::math::KernelList const&, pex::policy::Policy const&, std::shared_ptr<Eigen::MatrixXd>)) &
```
fred3m: I'm a little confused. I thought that shared_ptr's to Eigen matrices was removed because it was causing problems. Is it only a problem when a shared_ptr to an Eigen::Matrix is returned?

python/lsst/ip/diffim/makeRatingVector.py
```diff
@@ -48,32 +47,32 @@ def makeRatingVector(kernelCellSet, spatialKernel, spatialBg):
...
-    #                                                 sdqaVector[i].getValue(),
-    #                                                 sdqaVector[i].getErr()))
+    # for i in range(sdqaVector.size()):
+    #     pexLog.Trace("lsst.ip.diffim.makeSdqaRatingVector", 5,
+    #                  "Sdqa Rating %s : %.2f %.2f" % (sdqaVector[i].getName(),
+    #                                                  sdqaVector[i].getValue(),
+    #                                                  sdqaVector[i].getErr()))
     #
-    #return sdqaVector
+    # return sdqaVector
```
fred3m: Half of this file is commented out. Should those lines just be removed at this point?

tests/testCompareToHotpants.py
```diff
@@ -62,10 +63,10 @@ def setUp(self):
...
                 afwGeom.Point2I(xc, yc), afwGeom.Extent2I(1, 1))))
 
-        #detConfig.detThresholdType = "stdev"
-        #kcDetect = ipDiffim.KernelCandidateDetectionF(pexConfig.makePolicy(detConfig))
-        #kcDetect.apply(self.smi, self.tmi)
-        #self.footprints = kcDetect.getFootprints()
+        # detConfig.detThresholdType = "stdev"
+        # kcDetect = ipDiffim.KernelCandidateDetectionF(pexConfig.makePolicy(detConfig))
+        # kcDetect.apply(self.smi, self.tmi)
+        # self.footprints = kcDetect.getFootprints()
```
fred3m: Can this block be deleted?

tests/testImagePsfMatch.py
```diff
@@ -146,7 +146,7 @@ def testSubtractExposures(self):
 
         self.assertEqual(type(resultsAL.subtractedExposure), afwImage.ExposureF)
         self.assertEqual(type(resultsAL.psfMatchingKernel), afwMath.LinearCombinationKernel)
-        self.assertEqual(type(resultsAL.backgroundModel), afwMath.Function2D)
+        self.assertEqual(type(resultsAL.backgroundModel), afwMath.Chebyshev1Function2D)
```
fred3m: What happened here to change the return type of backgroundModel?

r-owen: It was always returning Chebyshev1Function2D but SWIG showed the base class, not the actual class.

python/lsst/ip/diffim/makeRatingVector.py
r-owen: You are likely right. I'll see how old the code is and delete it if it's old, else ask around.

tests/testCompareToHotpants.py
r-owen: As above

python/lsst/ip/diffim/detail/buildSingleKernelVisitor.cc
r-owen: What precipitated the change is that one function that returned a `std::shared_ptr<Eigen::MatrixXd>` was failing in Python with a complaint that the returned value could not be converted to a Python object:
```
ipDiffim.makeRegularizationMatrix(self.policyDF)
TypeError: Unable to convert function return value to a Python type! The signature was
    (policy: lsst.pex.policy._policy.Policy) -> Eigen::Matrix<double, -1, -1, 0, -1, -1>
```
It was never clear to us why this was failing. It may be a clue that the function signature listed in the error message does not match the actual function signature, in that the shared pointer is missing. I am also unsure if the other functions that return shared pointer to matrices were working correctly. In any case...

Is was unwise to be using shared pointers to Eigen matrices, so rather than try to fix the pybind11 issue we decided to clean up the C++ API by removing use of shared pointers to Eigen matrices and vectors. Note that it would not have sufficed to fix just this one function because it produced data required by other functions, which required a shared pointer to a matrix. This also allowed us to improve const correctness.

@TallJimbo may wish to say more.

python/lsst/ip/diffim/makeRatingVector.py
r-owen: The code is ancient. I'll remove all commented-out code here and in testCmpareToHotpants.py as you suggested below.

r-owen: The problem is that the function does nothing. The commented out parts are what actually computed and returned the rating that was desired, but `sdqa` is gone. The function is never run. I'll get rid of the file.

PR #44, opened 2017-02-08
Review for DM-9316: tests broken with NumPy 1.12

jdswinbank:
This is required as of Numpy 1.12.

### DIFF
--- a/tests/testDipole.py
+++ b/tests/testDipole.py
@@ -79,10 +79,12 @@ def createDipole(w, h, xc, yc, scaling=100.0, fracOffset=1.2):
     # Create the dipole, offset by fracOffset of the Psf FWHM (pixels)
     offset = fracOffset * psfFwhmPix // 2
     array = image.getImage().getArray()
-    xp, yp = xc - psfw//2 + offset, yc - psfh//2 + offset
+    xp = int(xc - psfw//2 + offset)
+    yp = int(yc - psfh//2 + offset)
     array[yp:yp+psfh, xp:xp+psfw] += psfim.getArray()
 
-    xn, yn = xc - psfw//2 - offset, yc - psfh//2 - offset
+    xn = int(xc - psfw//2 - offset)
+    yn = int(yc - psfh//2 - offset)
     array[yn:yn+psfh, xn:xn+psfw] -= psfim.getArray()
 
     if display:


tests/testDipole.py
```diff
@@ -80,10 +80,10 @@ def createDipole(w, h, xc, yc, scaling=100.0, fracOffset=1.2):
     offset = fracOffset * psfFwhmPix // 2
     array = image.getImage().getArray()
     xp, yp = xc - psfw//2 + offset, yc - psfh//2 + offset
-    array[yp:yp+psfh, xp:xp+psfw] += psfim.getArray()
+    array[int(yp):int(yp)+psfh, int(xp):int(xp)+psfw] += psfim.getArray()
```
r-owen: Wouldn't `int(py+psfh)` be safer and clearer? Similarly below.

timj: aren't `xp` and `yp` already integers or is the problem that `xc` and `offset` aren't?

tests/testDipole.py
```diff
@@ -80,10 +80,10 @@ def createDipole(w, h, xc, yc, scaling=100.0, fracOffset=1.2):
     offset = fracOffset * psfFwhmPix // 2
     array = image.getImage().getArray()
     xp, yp = xc - psfw//2 + offset, yc - psfh//2 + offset
```
timj: Putting these calculations on a single line seems unnecessarily perl-like...

tests/testDipole.py
jdswinbank: > Wouldn't `int(py+psfh)` be safer and clearer? Similarly below

I think "safer" depends on what threat you are protecting against. Certainly, it would cause the array indexing to be robust against `psfh` becoming non-integral. My feeling was that if something which is currently an integer suddenly becomes a float, we'd probably want the test to break to make sure we notice.

> I'd rather the the `xc` and `yc` were forced to int earlier.

That would change the logic: `int(xc) - psfw//2 + int(offset)` is not necessarily the same as `int(xc - psfw//2 + offset)`. With care it would be possible to perform the casting earlier and still be certain of ending up with the same result, but I'm not sure what it really buys us.

timj: Well, can you intify `xp` and `yp` before using them as indices then so we have less clutter in the array indexing itself?

timj: in some sense
```python
xp = int(xc - psfw//2 + offset)
yp = int(yc - psfh//2 + offset)
```
is more readable because you can easily spot the differences between `xp` and `yp` by scanning down the screen.

r-owen: I really like this solution. I find it very readable. Note that it does have the same effect as what I asked for in the first place: it is obvious that the index will be int, which is comforting to the reader but which @jdswinbank raised objections to.

jdswinbank: Ok, ok, I'll add some more line breaks to keep everybody happy. :)

PR #45, opened 2017-02-23
Add unit tests for ImageMapReduce task

djreiss:


tests/testImageMapReduce.py
```diff
@@ -0,0 +1,243 @@
...
+                                         ImageMapperSubtask, ImageMapperSubtaskConfig,
+                                         ImageReducerSubtask, ImageReducerSubtaskConfig)
+
+
+def setup_module(module):
+    lsst.utils.tests.init()
+
+
+class TestImageMapperSubtaskConfig(ImageMapperSubtaskConfig):
+    """!
```
parejkoj: Please switch to pydoc style for all python docstrings here, and below: https://developer.lsst.io/docs/py_docs.html

tests/testImageMapReduce.py
```diff
@@ -0,0 +1,243 @@
...
+        default=10.
+    )
+
+
+class TestImageMapperSubtask(ImageMapperSubtask):
+    ConfigClass = TestImageMapperSubtaskConfig
+    _DefaultName = "ip_diffim_TestImageMapperSubtask"
+
+    def __init__(self, *args, **kwargs):
+        """! Create the image gridding subTask
```
parejkoj: The docstring for `__init__` goes in the class-level docstring.

tests/testImageMapReduce.py
```diff
@@ -0,0 +1,243 @@
...
+        @param *args arguments to be passed to lsst.pipe.base.task.Task.__init__
+        @param **kwargs keyword arguments to be passed to lsst.pipe.base.task.Task.__init__
+        """
+        ImageMapperSubtask.__init__(self, *args, **kwargs)
```
parejkoj: I don't think you even need this `__init__`, if you're not doing anything besides calling super.

tests/testImageMapReduce.py
```diff
@@ -0,0 +1,243 @@
...
+
+class ImageMapReduceTest(lsst.utils.tests.TestCase):
+    """!A test case for the image gridded processing task
+    """
+
+    def setUp(self):
+        self.exposure = afwImage.ExposureF(128, 128)
+        self.exposure.setPsf(measAlg.DoubleGaussianPsf(11, 11, 2.0, 3.7))
+        mi = self.exposure.getMaskedImage()
+        mi[:, :] = 0.
```
parejkoj: Huh. I just looked, and didn't see an Exposure constructor that would assign a value to the entire maskedImage. Could you ask slack#dm if such a constructor exists, and if not file a ticket against afw?

tests/testImageMapReduce.py
```diff
@@ -0,0 +1,243 @@
...
+
+    def __init__(self, *args, **kwargs):
+        """! Create the image gridding subTask
+        @param *args arguments to be passed to lsst.pipe.base.task.Task.__init__
+        @param **kwargs keyword arguments to be passed to lsst.pipe.base.task.Task.__init__
+        """
+        ImageMapperSubtask.__init__(self, *args, **kwargs)
+
+    def run(self, subExp, expandedSubExp, fullBBox, **kwargs):
+        """! Add `addAmount` to given subExposure.
```
parejkoj: "Add `config.addAmount` to a subExposure and return the result. subExposure is not modified."

Also, might as well call it `subExposure` in the code, if you're calling it such in the docstring.

tests/testImageMapReduce.py
```diff
@@ -0,0 +1,243 @@
...
+    \brief Configuration parameters for the TestImageMapperSubtask
+    """
+    addAmount = pexConfig.Field(
+        dtype=float,
+        doc="""Amount to add to image""",
+        default=10.
+    )
+
+
+class TestImageMapperSubtask(ImageMapperSubtask):
```
parejkoj: Prefacing a non-test class with `Test` is confusing. Maybe call it `AddAmountImageMapperSubtask`, since that's what it actually does, and note in the class docstring that it's a simple `ImageMapperSubtask` to test ImageMapper functionality? Similarly for its Config, and the other MapReduceSubtasks.

Now that I've looked at the testing code, renaming them to something closer to their functionality makes even more sense: I can't tell TestStuff1 from TestStuff2, but for one you test that an addition happened, and for the other you test a mean.

tests/testImageMapReduce.py
```diff
@@ -0,0 +1,243 @@
...
+
+    \brief Configuration parameters for the TestImageMapReduceTask
+    """
+    mapperSubtask = pexConfig.ConfigurableField(
+        doc="Mapper subtask to run on each subimage",
+        target=TestImageMapperSubtask,
+    )
+
+
+class TestImageMapperSubtask2(ImageMapperSubtask):
```
parejkoj: Better name may be `MeanImageMapperSubtask`?

tests/testImageMapReduce.py
```diff
@@ -0,0 +1,243 @@
...
+    ConfigClass = TestImageMapperSubtaskConfig
+    _DefaultName = "ip_diffim_TestImageMapperSubtask2"
+
+    def __init__(self, *args, **kwargs):
```
parejkoj: Don't think you need this `__init__`, since you just call super.

tests/testImageMapReduce.py
```diff
@@ -0,0 +1,243 @@
...
+
+    def run(self, subExp, expandedSubExp, fullBBox, **kwargs):
+        """!Return the mean of the given subExposure.
+
+        @param[in] subExp the sub-exposure of `exposure` upon which to operate
+        @param[in] expandedSubExp the expanded sub-exposure of `exposure` upon which to operate
+        @param[in] fullBBox the bounding box of the original exposure
+        @return a float, the mean of `subExp.getMaskedImage().getArray()`.
+        """
+        return subExp.getMaskedImage().getImage().getArray().mean()
```
parejkoj: Ah the joys of afw Exposures...

tests/testImageMapReduce.py
```diff
@@ -0,0 +1,243 @@
...
+        """
+        config = TestImageMapReduceConfig()
+        task = ImageMapReduceTask(config)
+        config.mapperSubtask.addAmount = 5.
+        newExp = task.run(self.exposure)
+        newMI = newExp.getMaskedImage()
+        newArr = newMI.getImage().getArray()
+        mi = self.exposure.getMaskedImage()
+
+        self.assertClose(mi.getImage().getArray().mean(), newArr.mean() - 5.)
```
parejkoj: Can't you replace all of these asserts with a single call to: 

`self.assertFloatsAlmostEqual(mi.getImage().getArray()+5, newArr)`

since what you really want to compare is every array value?

tests/testImageMapReduce.py
```diff
@@ -0,0 +1,243 @@
...
+        config.gridStepX = config.gridStepY = 8
+        config.reducerSubtask.reduceOperation = 'average'
+        task = ImageMapReduceTask(config)
+        config.mapperSubtask.addAmount = 5.
+        newExp = task.run(self.exposure)
+        newMI = newExp.getMaskedImage()
+        newArr = newMI.getImage().getArray()
+        mi = self.exposure.getMaskedImage()
+
+        self.assertClose(mi.getImage().getArray().mean(), newArr.mean() - 5.)
```
parejkoj: As above, use `self.assertFloatsAlmostEqual`

tests/testImageMapReduce.py
```diff
@@ -0,0 +1,243 @@
...
+        self.assertClose(newArr.min(), 5.)
+        self.assertClose(newArr.max(), 5.)
+
+    def testExampleTaskWithOverlaps(self):
+        """Test sample grid task that adds 5.0 to input image and uses
+        'average' `reduceOperation`.
+        """
+        config = TestImageMapReduceConfig()
+        config.gridStepX = config.gridStepY = 8
+        config.reducerSubtask.reduceOperation = 'average'
```
parejkoj: I don't know what reducerSubtask.reduceOperation will do, but it looks like this and the NoOverlaps version both have identical asserts, so it looks to me that they produce identical results. Should they? If so, what's the difference between them?

tests/testImageMapReduce.py
```diff
@@ -0,0 +1,243 @@
...
+        'none' `reduceOperation`.
+        """
+        config = TestImageMapReduceConfig2()
+        config.gridStepX = config.gridStepY = 8
+        config.reducerSubtask.reduceOperation = 'none'
+        task = ImageMapReduceTask(config)
+        subMeans = task.run(self.exposure)
+
+        self.assertEqual(len(subMeans), len(task.boxes0))
+        self.assertClose(subMeans[0], 0.)
```
parejkoj: I don't know what `subMeans` should look like, but you should test all values. If it really should be all zeros, use `self.assertFloatsEqual(subMeans, 0)`. Although, I wonder if you shouldn't have some values in it, so that the mean is something more interesting than zero?

tests/testImageMapReduce.py
```diff
@@ -0,0 +1,243 @@
...
+        config.reducerSubtask.reduceOperation = 'none'
+        centroidsX = [i for i in np.linspace(0, 128, 50)]
+        centroidsY = centroidsX
+        config.gridCentroidsX = centroidsX
+        config.gridCentroidsY = centroidsY
+        task = ImageMapReduceTask(config)
+        subMeans = task.run(self.exposure)
+
+        self.assertEqual(len(subMeans), len(centroidsX))
+        self.assertClose(subMeans[0], 0.)
```
parejkoj: As above, test with `self.assertFloatsEqual(subMeans, 0)` and try to come up with a non-zero version of the test.

tests/testImageMapReduce.py
```diff
@@ -0,0 +1,243 @@
...
+
+    def testGridValidity(self):
+        """!Test sample grids with various spacings and sizes and other options.
```
parejkoj: No doxygen `!` in python docstrings.

tests/testImageMapReduce.py
```diff
@@ -0,0 +1,243 @@
...
+                    boxes = task._generateGrid(self.exposure)
+                    if len(boxes[0]) > 1000:  # bypass to prevent slow testing
+                        continue
+                    newExp = task.run(self.exposure)
+                    newMI = newExp.getMaskedImage()
+                    newArr = newMI.getImage().getArray()
+                    mi = self.exposure.getMaskedImage()
+
+                    mn = newArr.mean()
+                    self.assertClose(mi.getImage().getArray().mean(), mn - expectedVal)
```
parejkoj: If you're going to have tests in a loop, you need to pass a `msg` to the asserts so one can tell which set of parameters failed.

tests/testImageMapReduce.py
```diff
@@ -0,0 +1,243 @@
...
+                    self.assertClose(mn, expectedVal)
```
parejkoj: Same comments as above about testing the full array.

tests/testImageMapReduce.py
```diff
@@ -0,0 +1,243 @@
...
+        for scaleByFwhm in (False, True):
+            config.scaleByFwhm = scaleByFwhm
+            for gsx in range(51, 1, -10):
+                config.gridStepX = config.gridSizeX = gsx
+                for gsy in range(51, 1, -10):
+                    config.gridStepY = config.gridSizeY = gsy
+
+                    task = ImageMapReduceTask(config)
+                    boxes = task._generateGrid(self.exposure)
+                    if len(boxes[0]) > 1000:  # bypass to prevent slow testing
```
parejkoj: How much slower is it? What bit of code makes it slow? Seems like a potentially interesting possible failure mode (having many boxes).

tests/testImageMapReduce.py
```diff
@@ -0,0 +1,243 @@
...
+        """
+        config = TestImageMapReduceConfig()
+        config.reducerSubtask.reduceOperation = 'copy'
+        expectedVal = 1.
+        config.mapperSubtask.addAmount = expectedVal
+
+        for scaleByFwhm in (False, True):
+            config.scaleByFwhm = scaleByFwhm
+            for gsx in range(51, 1, -10):
+                config.gridStepX = config.gridSizeX = gsx
```
parejkoj: Should the gridSize and gridStep always be equal? If not, you should have at least one test where they're different.

tests/testImageMapReduce.py
```diff
@@ -0,0 +1,243 @@
...
+        """!Test sample grids with various spacings and sizes and other options.
+
+        Try to see if we can break it.
+        """
+        config = TestImageMapReduceConfig()
+        config.reducerSubtask.reduceOperation = 'copy'
+        expectedVal = 1.
+        config.mapperSubtask.addAmount = expectedVal
+
+        for scaleByFwhm in (False, True):
```
parejkoj: I'd put the contents of one of these loops (maybe the inner-most one?) in a separate method, say `_testGridValidity(self, scaleByFwhm, gridX, gridY)`, and then call that. The nested loops are a bit unwieldy.

parejkoj: It also concerns me that scaleByFwhm doesn't change any of the results. You should probably have a test where that value matters, otherwise it might be doing nothing and you'd never know.

tests/testImageMapReduce.py
```diff
@@ -0,0 +1,243 @@
...
+
+class TestImageMapperSubtaskConfig(ImageMapperSubtaskConfig):
+    """!
+    \anchor TestImageMapperSubtaskConfig_
+
+    \brief Configuration parameters for the TestImageMapperSubtask
+    """
+    addAmount = pexConfig.Field(
+        dtype=float,
+        doc="""Amount to add to image""",
```
parejkoj: Don't need triple quotes here.

tests/testImageMapReduce.py
djreiss: Renamed as you suggested, including changing `TestImageMapperSubtask2` to `GetMeanImageMapperSubtask`.

tests/testImageMapReduce.py
djreiss: As above, changed to `GetMeanImageMapperSubtask`.

tests/testImageMapReduce.py
djreiss: For exposures, that would be tricky (image, variance, masks, psfs, etc.) To me, that would only make sense if it explicitly sets everything to zero. I am not sure if the default empty constructor guarantees that (probably not, for speed), so I just do it explicitly here.

tests/testImageMapReduce.py
djreiss: Yes. Difference is with 'average' and when the grid elements overlap, it will take the average of the returned values for areas that contain overlapping regions. Thus for this simple case it produces the same exact result as the no overlap test.

tests/testImageMapReduce.py
djreiss: Changed to use a value of 1.234. Perplexed however that I have to adjust the rtol to 1e-6 for the test to pass. The pixels of the test image are not exactly set to 1.234 due to floating point errors and these propagate to the mean.

tests/testImageMapReduce.py
djreiss: scaleByFwhm changes the grid parameters, but here we simply want to ensure that given a set of input grid parameters, we don't fail and the final output image is not wrong. Thus the fact that they are the same is a good thing. That said, you're right that we should check the grid itself. I'll add a test for that.

tests/testImageMapReduce.py
djreiss: It is not a failure; it would just lead to a very long unit test run which I figure we want to avoid for regular CI.

tests/testImageMapReduce.py
RobertLuptonTheGood: I don't understand this.  The pixels are not 1.234 because that can't be exactly represented as a float, but all should be the same.  So the problem should be fixed by setting the correct mean (maybe as the (0,0) value of the image), not by modifying the tolerance.

If this doesn't work the problem is in the calculation of the mean (e.g. using a float accumulator)

tests/testImageMapReduce.py
parejkoj: This does suggest to me that there should be a separate test that results in different values. Otherwise, the guts of "average" and "noOverlaps" could be the same and you'd never know.

tests/testImageMapReduce.py
parejkoj: I'm uncomfortable with a "just in case" thing that disables tests. If you've tweaked the parameters so that it doesn't happen, just get rid of it. If a future person re-changes the parameters such that the test takes a long time, they can figure out what to do then.

tests/testImageMapReduce.py
djreiss: Thanks, comparing to [0,0] was the correct thing to do and the test passes now without an rtol.

tests/testImageMapReduce.py
djreiss: Or maybe a test that compares the output of the two and makes sure that they're the same? The problem is, if ever they don't give the same result on the same image, then something is broken.

tests/testImageMapReduce.py
```diff
@@ -167,71 +166,113 @@ def testExampleTaskWithOverlaps(self):
...
+            else:  # otherwise the grid is too fine and elements too small.
+                config.gridStepX = gstepx * 3.
+                config.gridSizeX = gsizex * 3.
+                config.gridStepY = gstepy * 3.
+                config.gridSizeY = gsizey * 3.
+            task = ImageMapReduceTask(config)
+            boxes = task._generateGrid(self.exposure)
+            ind = 0 if scaleByFwhm else 1
+            lenBoxes[ind] = len(boxes[0])
+            if len(boxes[0]) > 1000:  # bypass to prevent slow testing
```
parejkoj: I still don't like needing this if statement, particularly if you've reworked it to not result in such a large number of boxes.

tests/testImageMapReduce.py
```diff
@@ -167,71 +166,113 @@ def testExampleTaskWithOverlaps(self):
...
-                        continue
-                    newExp = task.run(self.exposure)
-                    newMI = newExp.getMaskedImage()
-                    newArr = newMI.getImage().getArray()
-                    mi = self.exposure.getMaskedImage()
-
-                    mn = newArr.mean()
-                    self.assertClose(mi.getImage().getArray().mean(), mn - expectedVal)
-                    self.assertClose(mn, expectedVal)
+            if scaleByFwhm:
```
parejkoj: This suggests to me that you should have two separate test methods, one for `scaleByFwhm=True` and the other for `False`.

tests/testImageMapReduce.py
```diff
@@ -167,71 +166,113 @@ def testExampleTaskWithOverlaps(self):
...
 
         self.assertEqual(len(subMeans), len(centroidsX))
-        self.assertClose(subMeans[0], 0.)
+        self.assertFloatsAlmostEqual(np.array(subMeans), 1.234, rtol=1e-6)
 
     def testGridValidity(self):
-        """!Test sample grids with various spacings and sizes and other options.
+        """Test sample grids with various spacings and sizes and other options.
 
         Try to see if we can break it.
```
parejkoj: I'm not sure this line of the docstring is actually helpful? Maybe if you defined what this could break, and how?

tests/testImageMapReduce.py
parejkoj: It's good we brought this up: https://jira.lsstcorp.org/browse/DM-9593

tests/testImageMapReduce.py
parejkoj: Ok, in that case, I think a useful more complicated test would be exactly what you say, on a more complicated input data set (say, a random field + PSFs). Then you'll have two tests that just check that each of those algorithms is sane, plus one that checks that they produce the same result on a more complicated data set. That should result in more robust testing overall.

tests/testImageMapReduce.py
djreiss: That will lead to a lot of code duplication. Plus I want to compare the results of `scaleByFwhm=True` vs.  `False`.

tests/testImageMapReduce.py
```diff
@@ -130,58 +142,122 @@ class ImageMapReduceTest(lsst.utils.tests.TestCase):
...
         config.mapperSubtask.addAmount = 5.
-        newExp = task.run(self.exposure)
+        config.reducerSubtask.reduceOperation = reduceOp
+        newExp = task.run(self.exposure, addNans=withNaNs).exposure
         newMI = newExp.getMaskedImage()
         newArr = newMI.getImage().getArray()
-        mi = self.exposure.getMaskedImage()
+        isnan = np.isnan(newArr)
+        if not withNaNs:
+            self.assertTrue(np.sum(isnan) == 0,
```
parejkoj: Don't you want `self.assertEqual(np.sum(isnan), 0)`?

tests/testImageMapReduce.py
```diff
@@ -130,58 +142,122 @@ class ImageMapReduceTest(lsst.utils.tests.TestCase):
     """A test case for the image gridded processing task
     """
     def setUp(self):
+        self._makeImage()
```
parejkoj: Since you're using `msg=` now, probably set `self.longMessage=True`, to get both your custom messages, and the default unittest ones.

tests/testImageMapReduce.py
```diff
@@ -130,58 +142,122 @@ class ImageMapReduceTest(lsst.utils.tests.TestCase):
...
+
+        mi = self.exposure.getMaskedImage().getImage().getArray()
+        self.assertFloatsAlmostEqual(mi[~isnan], newArr[~isnan] - 5.,
+                                     msg='Failed on withNaNs: %s' % str(withNaNs))
+
+    def _testAverageVersusCopy(self):
+        self.testAverageVersusCopy(withNaNs=False)
+        self.testAverageVersusCopy(withNaNs=True)
+
+    def testAverageVersusCopy(self, withNaNs=False):
```
parejkoj: This should have a preceeding `_`, otherwise it'll get run by the test suite on its own.

tests/testImageMapReduce.py
```diff
@@ -130,58 +142,122 @@ class ImageMapReduceTest(lsst.utils.tests.TestCase):
...
         task = ImageMapReduceTask(config)
         config.mapperSubtask.addAmount = 5.
-        newExp = task.run(self.exposure)
+        newExp = task.run(self.exposure, addNans=withNaNs).exposure
         newMI = newExp.getMaskedImage()
         newArr = newMI.getImage().getArray()
         mi = self.exposure.getMaskedImage()
+        isnan = np.isnan(newArr)
+        if not withNaNs:
+            self.assertTrue(np.sum(isnan) == 0,
```
parejkoj: Another `assertEqual` here.

tests/testImageMapReduce.py
```diff
@@ -130,58 +142,122 @@ class ImageMapReduceTest(lsst.utils.tests.TestCase):
...
+        task = ImageMapReduceTask(config)
+        config.mapperSubtask.addAmount = 5.
+        newExp = task.run(exposure2, addNans=withNaNs).exposure
+        newMI2 = newExp.getMaskedImage()
 
-    def testExampleTaskMean(self):
+        newMA1 = newMI1.getImage().getArray()
+        isnan = np.isnan(newMA1)
+        if not withNaNs:
+            self.assertTrue(np.sum(isnan) == 0)
```
parejkoj: `assertEqual`

tests/testImageMapReduce.py
```diff
@@ -130,58 +142,122 @@ class ImageMapReduceTest(lsst.utils.tests.TestCase):
...
+        newMA2 = newMI2.getImage().getArray()
+
+        self.assertFloatsAlmostEqual(newMA1[~isnan], newMA2[~isnan])
```
parejkoj: :+1:

tests/testImageMapReduce.py
```diff
@@ -258,21 +364,24 @@ def _testGridValidity(self, config, gstepx, gsizex, gstepy, gsizey, expectedVal=
...
-            if len(boxes[0]) > 1000:  # bypass to prevent slow testing
-                continue
-            newExp = task.run(self.exposure)
+            lenBoxes[ind] = len(task.boxes0)
+            newExp = task.run(self.exposure).exposure
             newMI = newExp.getMaskedImage()
             newArr = newMI.getImage().getArray()
-            mi = self.exposure.getMaskedImage()
+            isnan = np.isnan(newArr)
+            self.assertTrue(np.sum(isnan) == 0, msg='Failed NaN (%d), on config: %s' %
```
parejkoj: `assertEqual` again

tests/testImageMapReduce.py
```diff
@@ -258,21 +364,24 @@ def _testGridValidity(self, config, gstepx, gsizex, gstepy, gsizey, expectedVal=
...
+                            (np.sum(isnan), str(config)))
 
-            self.assertFloatsAlmostEqual(mi.getImage().getArray(), newArr - expectedVal,
+            mi = self.exposure.getMaskedImage().getImage().getArray()
+            self.assertFloatsAlmostEqual(mi[~isnan], newArr[~isnan] - expectedVal,
                                          msg='Failed on config: %s' % str(config))
 
-        self.assertTrue(lenBoxes[0] < lenBoxes[1], msg='Failed on config: %s' % str(config))
+        self.assertTrue(lenBoxes[0] < lenBoxes[1], msg='Failed lengths on config: %s' %
```
parejkoj: `self.assertLess(lenBoxes[0], lenBoxes[1], ...)`

tests/testImageMapReduce.py
djreiss: Apparently True is the default, but I'll set it explicitly anyway.

tests/testImageMapReduce.py
djreiss: Oops, missed that.

djreiss:
Final commit of the unit test in this PR was instead moved to PR #46, to be merged alongside ImageMapReduceTask.

PR #46, opened 2017-02-23
Add imageMapReduce task

djreiss:
Unit tests are in PR #45.


python/lsst/ip/diffim/imageMapReduce.py
```diff
@@ -0,0 +1,575 @@
...
+        """Generate two lists of bounding boxes that evenly grid `exposure`
+
+        Unless the config was provided with `centroidCoordsX` and
+        `centroidCoordsY`, grid (subimage) centers are spaced
+        evenly by gridStepX/Y. Then the grid is adjusted as
+        little as possible to evenly cover the input exposure (if
+        adjustGridOption is not 'none'). Then the second set of
+        bounding boxes is expanded by borderSizeX/Y. The expanded
+        bounding boxes are adjusted to ensure that they intersect
+        the exposure's bounding box. The resulting lists of bounding
```
parejkoj: Do you need to both return the lists and save them as instance variables? I think the latter is more than sufficient, especially since this is a "private" method.

python/lsst/ip/diffim/imageMapReduce.py
```diff
@@ -0,0 +1,575 @@
...
+        boxes and corresponding expanded bounding boxes are
+        returned, and also set to `self.boxes0`, `self.boxes1`.
+
+        Parameters
+        ----------
+        exposure : afwImage.Exposure`
```
parejkoj: unneeded backtic at the end?

python/lsst/ip/diffim/imageMapReduce.py
```diff
@@ -0,0 +1,575 @@
...
+            input exposure whose full bounding box is to be evenly gridded.
+        forceEvenSized : boolean
+            force grid elements to have even-valued x- and y- dimensions? (for ZOGY)
```
parejkoj: Spell out ZOGY more?

python/lsst/ip/diffim/imageMapReduce.py
```diff
@@ -0,0 +1,575 @@
...
+                bb0, bb1 = offsetAndClipBoxes(bbox0, bbox1, xoff, yoff, bbox)
+                yoff += gridStepY
+                if bb0.getArea() > 0 and bb1.getArea() > 0:
+                    self.boxes0.append(bb0)
+                    self.boxes1.append(bb1)
+            xoff += gridStepX
+
+        return self.boxes0, self.boxes1
+
+    def _plotBoxes(self, exposure):
```
parejkoj: Should this plotting code live outside the task somewhere?

python/lsst/ip/diffim/imageMapReduce.py
```diff
@@ -0,0 +1,575 @@
...
+        gridSizeY = self.config.gridSizeY
+        gridStepX = self.config.gridStepX
+        gridStepY = self.config.gridStepY
+        borderSizeX = self.config.borderSizeX
+        borderSizeY = self.config.borderSizeY
+        adjustGridOption = self.config.adjustGridOption
+        scaleByFwhm = self.config.scaleByFwhm
+        bbox = exposure.getBBox()
+
+        psfFwhm = (exposure.getPsf().computeShape().getDeterminantRadius() *
```
parejkoj: Technically, or coding standards don't like spaces around `*`, but I don't really care.

python/lsst/ip/diffim/imageMapReduce.py
```diff
@@ -0,0 +1,575 @@
...
+                   2. * np.sqrt(2. * np.log(2.)))
+
+        def rescaleValue(val):
+            if scaleByFwhm:
+                return np.rint(val * psfFwhm).astype(int)
```
parejkoj: What happens if the psf has different dimensions in X vs. Y?

python/lsst/ip/diffim/imageMapReduce.py
```diff
@@ -0,0 +1,575 @@
...
+        borderSizeY = rescaleValue(borderSizeY)
+
+        nGridX = bbox.getWidth() // gridStepX
+        nGridY = bbox.getHeight() // gridStepY
+
+        if adjustGridOption == 'spacing':
+            nGridX = bbox.getWidth() / gridStepX
+            # Readjust gridStepX so that it fits perfectly in the image.
+            gridStepX = float(bbox.getWidth() - gridSizeX) / float(nGridX)
+            if gridStepX <= 0:
```
parejkoj: `<=0` here, or `<=1`? (same below)

python/lsst/ip/diffim/imageMapReduce.py
```diff
@@ -0,0 +1,575 @@
...
+        self.boxes0 = []  # "main" boxes; store in task so can be extracted if needed
+        self.boxes1 = []  # "expanded" boxes
+
+        # use given centroids as centers for bounding boxes
+        if self.config.gridCentroidsX is not None and len(self.config.gridCentroidsX) > 0:
+            for i, centroidX in enumerate(self.config.gridCentroidsX):
+                centroidY = self.config.gridCentroidsY[i]
+                centroid = afwGeom.Point2D(centroidX, centroidY)
+                bb0 = afwGeom.Box2I(bbox0)
+                xoff = int(np.floor(centroid.getX())) - bb0.getWidth()//2
```
parejkoj: Good use of explicit integer division.

python/lsst/ip/diffim/imageMapReduce.py
```diff
@@ -0,0 +1,575 @@
...
+            bb1 = afwGeom.Box2I(bbox1)
+            bb1.shift(afwGeom.Extent2I(xoff, yoff))
+            bb1.clip(bbox)
+            if forceEvenSized:
+                bb0.grow(afwGeom.Extent2I(bb0.getWidth() % 2, bb0.getHeight() % 2))
+                bb1.grow(afwGeom.Extent2I(bb1.getWidth() % 2, bb1.getHeight() % 2))
+            return bb0, bb1
+
+        xoff = 0
+        while(xoff <= bbox.getWidth()):
```
parejkoj: I've missed `while` loops!

djreiss:
> The overall design is reasonable. Several comments about docs, new unittests and returning Structs.
> 
> I think it would be good for you to run coverage (either via pytest-cov, or coverage.py), to help find where your current tests aren't getting to. I can help you with that, if you like.
> 
> Although I have some comments to that effect above, you do need some more docs about what happens when Map doesn't return an Exposure (e.g., moving average as you suggested), both at the file/class level, and for the method args/return values.
> 
> I do worry about about implementing MapReduce in the stack in ip_diffim, instead of either taking an off-the-shelf python MR code, or putting it somewhere higher (e.g. lsst.utils?). The implementation here is somewhat specific to your use case, though, so I don't know how feasible either of those would be. At the very least, you should think about how one would make this parallel via multiprocessing or mpi, since MR lends it self perfectly to that.

Thanks for the excellent review. I have no issues with nearly all of your comments. I agree with the potential for moving this out of `ip_diffim` if there is a use-case, since it is not specific to image differencing. For now, I have plans for two separate uses of ImageMapReduce in `ip_diffim`, which is why I placed it there.

python/lsst/ip/diffim/imageMapReduce.py
```diff
@@ -0,0 +1,575 @@
...
+import abc
+from future.utils import with_metaclass
+
+import lsst.afw.image as afwImage
+import lsst.afw.math as afwMath
+import lsst.afw.geom as afwGeom
+import lsst.pex.config as pexConfig
+import lsst.pipe.base as pipeBase
+
+__all__ = ("ImageMapReduceTask", "ImageMapReduceConfig",
```
djreiss: Thanks for pointing out, but I'll leave it as-is until there's an implemented RFC.

python/lsst/ip/diffim/imageMapReduce.py
```diff
@@ -0,0 +1,575 @@
...
+
+        Parameters
+        ----------
+        subExp : afw.Exposure
+            the sub-exposure upon which to operate
+        expandedSubExp : afw.Exposure
+            the expanded sub-exposure upon which to operate
+        fullBBox : afwGeom.BoundingBox
+            the bounding box of the original exposure
+        kwargs :
```
djreiss: They get passed to this method from above. Allows for flexibility in defining a `run` method for a subclass, and any additional data it may require.

python/lsst/ip/diffim/imageMapReduce.py
```diff
@@ -0,0 +1,575 @@
...
+            additional keyword arguments
+
+        Returns
+        -------
+        anything, including an a `afw.Exposure`
```
djreiss: OK. This requires coordination with ImageReducerSubtask. I will make it explicit.

python/lsst/ip/diffim/imageMapReduce.py
```diff
@@ -0,0 +1,575 @@
...
+        """
+        pass
+
+
+class ImageReducerSubtaskConfig(pexConfig.Config):
+    """Configuration parameters for the ImageReducerSubtask
+    """
+    reduceOperation = pexConfig.ChoiceField(
+        dtype=str,
+        doc="""Operation to use for reducing subimages into new image.""",
```
djreiss: But is there anything wrong with having em?

python/lsst/ip/diffim/imageMapReduce.py
```diff
@@ -0,0 +1,575 @@
...
+        default="copy",
```
djreiss: Well the default grid configuration doesn't have overlaps, so this is not dangerous for that default. The non-deterministic bit is just a warning.

python/lsst/ip/diffim/imageMapReduce.py
```diff
@@ -0,0 +1,575 @@
...
+
+        Returns
+        -------
+        an `afw.Exposure` or a list, depending on `config.reduceOperation`.
+
+        Notes
+        -----
+        And currently known issues:
+        1. This currently should correctly handle overlapping patches.
+           For overlapping patches, use `config.reduceOperation='average'`.
```
djreiss: OK, sure.

python/lsst/ip/diffim/imageMapReduce.py
```diff
@@ -0,0 +1,575 @@
...
+    reducerSubtask = pexConfig.ConfigurableField(
+        doc="Subtask to combine results of mapperSubTask",
+        target=ImageReducerSubtask,
+    )
+
+    # Separate gridCentroidsX and gridCentroidsY since pexConfig.ListField accepts limited dtypes
+    #  (i.e., no Point2D)
+    gridCentroidsX = pexConfig.ListField(
+        dtype=float,
+        doc="""Input X centroids around which to place subimages.
```
djreiss: Good point, it's the latter. However, the former might be a nice option.

python/lsst/ip/diffim/imageMapReduce.py
```diff
@@ -0,0 +1,575 @@
...
+        """
+        boxes0, boxes1 = self.boxes0, self.boxes1
+        if boxes0 is None:
+            boxes0, boxes1 = self._generateGrid(exposure)
+        if len(boxes0) != len(boxes1):
+            raise Exception('Uh oh!')   # TBD: define a specific exception to raise
+
+        self.log.info("Processing %d sub-exposures" % len(boxes0))
+        patches = []
+        for i in range(len(boxes0)):
```
djreiss: Anticipated this one, already done.

python/lsst/ip/diffim/imageMapReduce.py
```diff
@@ -0,0 +1,575 @@
...
+        patches : list
+            list of subExposures of `exposure` to merge
+        exposure : afwImage.Exposure
+            the original exposure
+        **kwargs :
+            additional keyword arguments
+
+        Returns
+        -------
+        Output of `reducerSubtask.run` which may be an `afw.Exposure` or
```
djreiss: Upon further thought, I'm not sure if pipeBase.Struct is the right thing here. In all cases, if I were to use `pipeBase.Struct` it would contain a single item -- either an exposure or a list. I suppose I could enforce that it either return a `pipeBase.Struct(exposure=result)` or `pipeBase.Struct(items=result)`. Thoughts?

python/lsst/ip/diffim/imageMapReduce.py
djreiss: It uses the average.

python/lsst/ip/diffim/imageMapReduce.py
djreiss: This is just to catch a case which shouldn't happen. I'll raise an exception instead.

python/lsst/ip/diffim/imageMapReduce.py
djreiss: Not sure. Why?

python/lsst/ip/diffim/imageMapReduce.py
djreiss: I take it back, everything now returns pipeBase.Struct's.

python/lsst/ip/diffim/imageMapReduce.py
parejkoj: Not really, I guess, but it's surprising to see them on single-line strings.

python/lsst/ip/diffim/imageMapReduce.py
parejkoj: I guess that's ok? I just worry that someone would change the other to have overlaps and not notice this comment.

python/lsst/ip/diffim/imageMapReduce.py
parejkoj: I guess the question is, would you plot as part of the Task, or after the task is done?

python/lsst/ip/diffim/imageMapReduce.py
```diff
@@ -175,7 +246,8 @@ class ImageMapReduceConfig(pexConfig.Config):
     )
 
     # Separate gridCentroidsX and gridCentroidsY since pexConfig.ListField accepts limited dtypes
-    #  (i.e., no Point2D)
+    #  (i.e., no Point2D). The resulting set of centroids is the "vertical stack" of
+    #  `gridCentroidsX` and `gridCentroidsY`.
```
parejkoj: Please give an example here, showing what you get for, e.g., `(1,2 ), (3,4)`

djreiss:
Waited until pybind11 merge. Confirmed build on Jenkins.

PR #53, opened 2017-03-29
DM-7611: Implement spatially varying decorrelation kernel in ip_diffim

djreiss:
…tion

Additional necessary improvements:

- Update ImageMapReduceTask to use Exposures other than ExposureF
- Update ImageReducerSubtask to correctly propagate masks
- Docstring fixes in imageMapReduce.py
- Additional unit tests and refactoring of testImageDecorrelation.py
  to test decorrelation via two different tasks

### DIFF
--- a/python/lsst/ip/diffim/imageDecorrelation.py
+++ b/python/lsst/ip/diffim/imageDecorrelation.py
@@ -27,12 +27,17 @@ import numpy as np
 import scipy.fftpack
 
 import lsst.afw.image as afwImage
+import lsst.afw.geom as afwGeom
 import lsst.meas.algorithms as measAlg
 import lsst.afw.math as afwMath
 import lsst.pex.config as pexConfig
 import lsst.pipe.base as pipeBase
+import lsst.log
 
-__all__ = ("DecorrelateALKernelTask", "DecorrelateALKernelConfig")
+from .imageMapReduce import (ImageMapReduceConfig, ImageMapperSubtask)
+
+__all__ = ("DecorrelateALKernelTask", "DecorrelateALKernelConfig",
+           "DecorrelateALKernelMapperSubtask", "DecorrelateALKernelMapReduceConfig")
 
 
 class DecorrelateALKernelConfig(pexConfig.Config):
@@ -180,9 +185,6 @@ class DecorrelateALKernelTask(pipeBase.Task):
         @note Still TBD (ticket DM-6580): understand whether the convolution is correctly modifying
         the variance plane of the new subtractedExposure.
         """
-        self.log.info("Starting.")
-        kimg = None
-
         spatialKernel = psfMatchingKernel
         kimg = afwImage.ImageD(spatialKernel.getDimensions())
         bbox = subtractedExposure.getBBox()
@@ -190,41 +192,32 @@ class DecorrelateALKernelTask(pipeBase.Task):
             xcen = (bbox.getBeginX() + bbox.getEndX()) / 2.
         if ycen is None:
             ycen = (bbox.getBeginY() + bbox.getEndY()) / 2.
-        self.log.info("Using matching kernel computed at (%d, %d)" % (xcen, ycen))
+        self.log.info("Using matching kernel computed at (%d, %d)", xcen, ycen)
         spatialKernel.computeImage(kimg, True, xcen, ycen)
-
-        if False:  # debug code to save spatially varying kernel for analysis
-            import pickle
-            import gzip
-            pickle.dump(spatialKernel, gzip.GzipFile('spatialKernel.p.gz', 'wb'))
 
         if svar is None:
             svar = self.computeVarianceMean(exposure)
-        self.log.info("Variance (science): %f" % svar)
         if tvar is None:
             tvar = self.computeVarianceMean(templateExposure)
-        self.log.info("Variance (template): %f" % tvar)
+        self.log.info("Variance (science, template): (%f, %f)", svar, tvar)
 
         var = self.computeVarianceMean(subtractedExposure)
-        self.log.info("Variance (uncorrected diffim): %f" % var)
+        self.log.info("Variance (uncorrected diffim): %f", var)
 
         corrKernel = DecorrelateALKernelTask._computeDecorrelationKernel(kimg.getArray(), svar, tvar)
-        self.log.info("Convolving.")
         correctedExposure, corrKern = DecorrelateALKernelTask._doConvolve(subtractedExposure, corrKernel)
-        self.log.info("Updating correctedExposure and its PSF.")
 
         # Compute the subtracted exposure's updated psf
-        psf = subtractedExposure.getPsf().computeImage().getArray()
+        psf = subtractedExposure.getPsf().computeImage(afwGeom.Point2D(xcen, ycen)).getArray()
         psfc = DecorrelateALKernelTask.computeCorrectedDiffimPsf(corrKernel, psf, svar=svar, tvar=tvar)
         psfcI = afwImage.ImageD(psfc.shape[0], psfc.shape[1])
         psfcI.getArray()[:, :] = psfc
         psfcK = afwMath.FixedKernel(psfcI)
         psfNew = measAlg.KernelPsf(psfcK)
         correctedExposure.setPsf(psfNew)
-        self.log.info("Complete.")
 
         var = self.computeVarianceMean(correctedExposure)
-        self.log.info("Variance (corrected diffim): %f" % var)
+        self.log.info("Variance (corrected diffim): %f", var)
 
         return pipeBase.Struct(correctedExposure=correctedExposure, correctionKernel=corrKern)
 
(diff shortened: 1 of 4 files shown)


python/lsst/ip/diffim/imageDecorrelation.py
```diff
@@ -353,3 +353,95 @@ def _doConvolve(exposure, kernel):
...
+        subExp2 = scienceExposure.Factory(scienceExposure, expandedSubExposure.getBBox())
+        subExp1 = templateExposure.Factory(templateExposure, expandedSubExposure.getBBox())
+
+        psfMatchingKernel = alTaskResult.psfMatchingKernel if alTaskResult is not None else psfMatchingKernel
+
+        res = DecorrelateALKernelTask.run(self, subExp2, subExp1, expandedSubExposure,
+                                          psfMatchingKernel)
+        diffim = res.correctedExposure.Factory(res.correctedExposure, subExposure.getBBox())
+        out = pipeBase.Struct(subExposure=diffim, decorrelationKernel=res.correctionKernel)
+        self.log.setLevel(logLevel)
```
parejkoj: also remove this bit of the logging.

python/lsst/ip/diffim/imageMapReduce.py
```diff
@@ -270,6 +272,7 @@ def run(self, mapperResults, exposure, **kwargs):
             wts = weights.getArray().astype(np.float)
             newMI.getImage().getArray()[:, :] /= wts
             newMI.getVariance().getArray()[:, :] /= wts
+            # TBD: set mask to something for pixels where wts == 0. Shouldn't happen.
```
parejkoj: TODOs should always have a linked ticket, otherwise they'll get lost.

python/lsst/ip/diffim/imageMapReduce.py
```diff
@@ -311,7 +314,7 @@ def _constructPsf(self, mapperResults, exposure):
         for i, res in enumerate(mapperResults):
             subExp = res.subExposure
             if subExp.getWcs() != wcsref:
-                raise Exception('Wcs of subExposure is different from exposure')
```
parejkoj: Good change.

python/lsst/ip/diffim/imageMapReduce.py
```diff
@@ -532,10 +535,12 @@ def _runMapper(self, exposure, doClone=False, **kwargs):
             raise ValueError('Bounding boxes list and expanded bounding boxes list are of different lengths')
 
         self.log.info("Processing %d sub-exposures", len(self.boxes0))
+        self.log.info("Mapper sub-task: %s", self.mapperSubtask._DefaultName)
```
parejkoj: These subtasks should be written as part of the Config output: do we need them to be logged, also?

tests/testImageDecorrelation.py
```diff
@@ -136,15 +138,31 @@ def makeFakeImages(xim=None, yim=None, svar=0.04, tvar=0.04, psf1=3.3, psf2=2.2,
...
-    if im1background != 0.:  # im1background = 10.
-        if verbose:
-            print('Background:', im1background)
-        im1 += im1background
-
     im1_psf = singleGaussian2d(x0im, y0im, 0, 0, psf1[0], psf1[1], theta=theta1)
     im2_psf = singleGaussian2d(x0im, y0im, offset[0], offset[1], psf2[0], psf2[1], theta=theta2)
 
+    def makeWcs(offset=0):
+        # taken from $AFW_DIR/tests/testMakeWcs.py
```
parejkoj: At least have a docstring that says what "offset" is?

tests/testImageDecorrelation.py
```diff
@@ -245,19 +265,19 @@ def _testImages(self):
 
         # Expected (ideal) variance of difference image
         expected_var = self.svar + self.tvar
-        print('Expected variance:', expected_var)
+        print('EXPECTED VARIANCE:', expected_var)
```
parejkoj: Since you have an assert about expected_var below, do you really need the print statement?

tests/testImageDecorrelation.py
```diff
@@ -266,10 +286,19 @@ def _testImages(self):
         self.assertLess(mn, expected_var)
         print('UNCORRECTED VARIANCE:', var, mn)
```
parejkoj: And again: if there's an assert for it, do we need to print it?

tests/testImageDecorrelation.py
```diff
@@ -281,23 +310,66 @@ def _testImages(self):
...
-        self._testImages()
+        self._setUpImages(svar=svar, tvar=tvar)
+        diffExp, mKernel, expected_var = self._makeAndTestUncorrectedDiffim()
+        corrected_diffExp = self._runDecorrelationTask(diffExp, mKernel)
+        self._testDecorrelation(expected_var, corrected_diffExp)
 
-    def testDiffimCorrection_higher_science_variance(self):
-        """Test decorrelated diffim when the science image variance is higher than that of the template.
+    def testDiffimCorrection_same_variance(self):
+        """Test decorrelated diffim from images with different variances.
```
parejkoj: docstring and method name don't match.

tests/testImageDecorrelation.py
```diff
@@ -281,23 +310,66 @@ def _testImages(self):
...
+        # Also compare the diffim generated here vs. the non-ImageMapReduce one
+        corrected_diffExp_OLD = self._runDecorrelationTask(diffExp, mKernel)
+        self.assertImagesNearlyEqual(corrected_diffExp.getMaskedImage().getImage(),
+                                     corrected_diffExp_OLD.getMaskedImage().getImage())
+        self.assertImagesNearlyEqual(corrected_diffExp.getMaskedImage().getVariance(),
+                                     corrected_diffExp_OLD.getMaskedImage().getVariance())
+        self.assertMasksEqual(corrected_diffExp.getMaskedImage().getMask(),
+                              corrected_diffExp_OLD.getMaskedImage().getMask())
+        # Is this the same as the above? Let's just be complete.
+        self.assertMaskedImagesNearlyEqual(corrected_diffExp.getMaskedImage(),
```
parejkoj: I think you only need the last line of these: see the docstring, which says "The mask planes must match exactly. The image and variance planes are nearly equal if all pixels obey: |val1 - val0| <= rtol*|val1| + atol or, for float types, if nan/inf/-inf pixels match.":

https://lsst-web.ncsa.illinois.edu/doxygen/x_masterDoxyDoc/namespacelsst_1_1afw_1_1image_1_1test_utils.html#abf24e24b9b36b6432943b513f2739822

python/lsst/ip/diffim/imageDecorrelation.py
```diff
@@ -353,3 +353,95 @@ def _doConvolve(exposure, kernel):
...
+        'decorrelationKernel', although that currently is not used.
+
+        Notes
+        -----
+        This `run` method accepts parameters identical to those of
+        `ImageMapperSubtask.run`, since it is called from the
+        `ImageMapperTask`.  See that class for more information.
+        """
+        logLevel = self.log.getLevel()
+        self.log.setLevel(lsst.log.ERROR)  # Prevent too much log verbosity from DecorrelateALKernelTask.run
```
djreiss: Problem is when the subtask is verbose, then each of the ~60 calls to it gets logged. Without this change, it slowed down processing of one image for me from ~1s to ~45s.

parejkoj: In that case, either set the logging level in the test that makes the call, configure it on the commandline, or file a ticket/bring up on slack#dm about logging being slow.

python/lsst/ip/diffim/imageDecorrelation.py
```diff
@@ -353,3 +353,95 @@ def _doConvolve(exposure, kernel):
...
+
+
+class DecorrelateALKernelMapperTask(DecorrelateALKernelTask):
+    """Task to be used as an ImageMapperSubtask for computing
+    A&L decorrelation on subimages on a grid across a A&L difference image.
+
+    This task subclasses DecorrelateALKernelTask in order to implement
+    all of that task's configuration parameters, as well as its `run` method.
+    """
+    ConfigClass = DecorrelateALKernelConfig
```
djreiss: No, that config is passed to the ImageMapReduceClass to set its mapperSubtask to this class.

Confusing, right?

python/lsst/ip/diffim/imageDecorrelation.py
```diff
@@ -353,3 +353,95 @@ def _doConvolve(exposure, kernel):
...
+        alTaskResult = kwargs.get('alTaskResult', None)
+        psfMatchingKernel = kwargs.get('psfMatchingKernel', None)
+        preConvKernel = kwargs.get('preConvKernel', None)
+
+        # subExp and expandedSubExp are subimages of the (un-decorrelated) diffim!
+        # So here we compute corresponding subimages of im1, and im2
+        subExp2 = scienceExposure.Factory(scienceExposure, expandedSubExposure.getBBox())
+        subExp1 = templateExposure.Factory(templateExposure, expandedSubExposure.getBBox())
+
+        psfMatchingKernel = alTaskResult.psfMatchingKernel if alTaskResult is not None else psfMatchingKernel
```
djreiss: or ValueError?

python/lsst/ip/diffim/imageDecorrelation.py
djreiss: See above...

python/lsst/ip/diffim/imageMapReduce.py
djreiss: I find it useful.

python/lsst/ip/diffim/imageDecorrelation.py
djreiss: Well, I only silence the INFO, not the ERROR messages. I guess it is not overly slow (my initial slowness claim was apparent in an IPython notebook, not from a unit test) -- it's just very verbose.

djreiss:
Thanks for the review! I think I responded to all your comments.

python/lsst/ip/diffim/imageDecorrelation.py
parejkoj: Ah. Should `DecorrelateALKernelMapperConfig` then be called `DecorrelateALKernelMapReduceConfig`? The naming here is definitely confusing.

python/lsst/ip/diffim/imageDecorrelation.py
parejkoj: I debated about that, but I think ValueError is for a different situation (e.g, received `-1` when only positive numbers are allowed): 

https://docs.python.org/2/library/exceptions.html#exceptions.ValueError
https://docs.python.org/2/library/exceptions.html#exceptions.RuntimeError

tests/testImageDecorrelation.py
parejkoj: Could you please say why you think the print statement is useful here? It would get gobbled up during jenkins runs, and not normally seen by a use, unless they have reason to look at the test output (e.g., the test failed). If there's some information not being captured by the current asserts, maybe we need more asserts?

python/lsst/ip/diffim/imageDecorrelation.py
parejkoj: Is the verbosity a problem in the unittests? That could be handled by setting the log level in the unittest itself. For actual production, if the logs are too verbose, maybe some of the INFOs need to be DEBUGs?

Alternately, @r-owen 's comment on slack here is a better approach: https://lsstc.slack.com/archives/C2JPL2DGD/p1490830986422452

python/lsst/ip/diffim/imageDecorrelation.py
djreiss: OK, I suppose that's better. I also updated the docstring to clarify a bit. I also renamed `DecorrelateALKernelMapperTask` to `DecorrelateALKernelMapperSubtask`.

tests/testImageDecorrelation.py
djreiss: As long as I'm still developing these tasks, I find it reassuring to look at the numbers and make sure they make sense. I cannot explain it better than that. I count 1422 'print' statements in unit tests across the lsst stack.

parejkoj: "The stack currently does something" is not normally a good justification, considering a lot of the cruft in there.

The thing I'm worried about is when someone else is working on this code, what are they supposed to do with that printed number? If you know what the rough value should be, maybe an `assertAlmostEqual(expected_var, SOMENUMBER)` is in order, as a sanity check?

python/lsst/ip/diffim/imageDecorrelation.py
djreiss: This is a unique situation. The task could be run once (in which case it is not too chatty -- about 4 INFO messages -- and the output is useful). Or in this case it was subclassed to be run as a mapperSubtask (and then you get ~100 x as many messages). I don't know if we have an available way to handle this situation.

tests/testImageDecorrelation.py
djreiss: OK wrapped the print statements in a verbose flag check (default False).

tests/testImageDecorrelation.py
djreiss: OK wrapped the print statements in a verbose flag check (default False).

python/lsst/ip/diffim/imageDecorrelation.py
parejkoj: I suspect we're going to have to deal with the situation soon enough (given the work on SuperTask and parallelization): Please file a ticket on how to manage logging in a "call this task 100 times" environment, and I'll be happy-ish.

djreiss: OK, I (1) removed some INFO verbosity in the parent task and (2) will filed a ticket.

SimonKrughoff: I'm a little worried that the questions raised by @djreiss will be lost in the noise in Jira.  I think it might be worth having a broader discussion.  Is a community post reasonable?

python/lsst/ip/diffim/imageDecorrelation.py
r-owen: I encourage you to undo that change. I am not aware of any code in the stack that uses "Subtask" in the name and it seems unnecessary. Any task can be treated as a subtask. Only subclasses of `CmdLineTask` can be top-level tasks, but all other tasks are still considered tasks.

PR #55, opened 2017-03-29
ModelPsfMatchTask: Set output exposure PSF to model PSF used

yalsayyad:


python/lsst/ip/diffim/modelPsfMatch.py
```diff
@@ -300,6 +300,7 @@ def run(self, exposure, referencePsfModel, kernelSum=1.0):
         psfMatchedExposure = afwImage.ExposureF(exposure.getBBox(), exposure.getWcs())
         psfMatchedExposure.setFilter(exposure.getFilter())
         psfMatchedExposure.setCalib(exposure.getCalib())
+        psfMatchedExposure.setPsf(referencePsfModel)
```
laurenam: I don't pretend to have looked at the downstream code & potential consequences, so I'll just ask and accept your answer to the question: does the refererecePsfModel still need to be returned as a separate component of the Struct given that it is now explicitly attached to the returned psfMatchedExposure?

yalsayyad: It doesn't still need to be there.  Because it has only been there for 4 months, it is *extremely* unlikely that anyone's non-stack scripts depend on it now, so I'm comfortable making the interface change of removing it.

PR #61, opened 2017-04-28
Tickets/dm 8145

djreiss:


python/lsst/ip/diffim/zogy.py
```diff
@@ -0,0 +1,839 @@
...
+        S = np.fft.ifft2(D_hat * Pd_bar)
+
+        # Adjust the variance planes of the two images to contribute to the final detection
+        # (eq's 26-29).
+        Pn_hat2 = np.conj(preqs.Pn_hat) * preqs.Pn_hat  # np.abs(preqs.Pn_hat)**2.
+        Kr_hat = self.Fr * self.Fn**2. * np.conj(preqs.Pr_hat) * Pn_hat2 / preqs.denom**2.
+        Pr_hat2 = np.conj(preqs.Pr_hat) * preqs.Pr_hat  # np.abs(preqs.Pr_hat)**2.
+        Kn_hat = self.Fn * self.Fr**2. * np.conj(preqs.Pn_hat) * Pr_hat2 / preqs.denom**2.
+
+        Kr_hat2 = np.fft.fft2(np.fft.ifft2(Kr_hat)**2.)
```
isullivan: Should you take the absolute value before squaring here?

python/lsst/ip/diffim/zogy.py
```diff
@@ -0,0 +1,839 @@
...
+
+        # Psf and image for science img (index 2)
+        subExp2 = expandedSubExposure
+
+        # Psf and image for template img (index 1)
+        subExp1 = template.Factory(template, expandedSubExposure.getBBox())
+
+        if sigmas is None:
+            sig1 = sig2 = None
+        else:  # for testing, can use the input sigma (e.g., global value for entire exposure)
```
isullivan: Please put this comment and the following ones after `if` or `else` statements on their own line.

python/lsst/ip/diffim/zogy.py
```diff
@@ -0,0 +1,839 @@
...
+            psf1 = np.pad(psf1, ((1, 1), (0, 0)), mode='constant')
+        elif psf1.shape[0] > psf1.shape[1]:
+            psf1 = np.pad(psf1, ((0, 0), (1, 1)), mode='constant')
+
+        # from diffimTests.diffimTests ...
+        if subExp1.getDimensions()[0] < psf1.shape[0] or subExp1.getDimensions()[1] < psf1.shape[1]:
+            return pipeBase.Struct(subExposure=subExposure)
+
+        psf1b = psf2b = None
+        if True and psf1.shape[0] == 41:  # it's a measured psf (hack!) Note this *really* helps for measured psfs.
```
isullivan: Since you're using this code again, it *really* should be a separate `kernelFilter` function. In that function, the size of the zero-padding should be configurable, but could have a default of 10 as you use here.
I don't think testing against the size of the psf equalling 41 is safe. Measured psfs will have a wide range of sizes.

python/lsst/ip/diffim/imageMapReduce.py
```diff
@@ -165,7 +165,10 @@ class ImageReducerSubtaskConfig(pexConfig.Config):
             "sum": """add pixels from overlaps (probably never wanted; used for testing)
                        into correct location in new exposure""",
             "average": """same as copy, but also average pixels from overlapped regions
-                       (NaNs ignored)"""
+                       (NaNs ignored)""",
```
isullivan: On line 31 above, afwMath is imported but not used.

tests/testImageDecorrelation.py
```diff
@@ -115,19 +115,19 @@ def makeFakeImages(xim=None, yim=None, svar=0.04, tvar=0.04, psf1=3.3, psf2=2.2,
         print(np.sqrt(psf1[0]**2 - psf2[0]**2))
         print('Offset:', offset)
 
-    xim = np.arange(-256, 256, 1) if xim is None else xim
+    xim = np.arange(-128, 128, 1) if xim is None else xim
```
isullivan: Please define a variable to store the value 128, and reference that here (and line 195)

tests/testZogy.py
```diff
@@ -0,0 +1,209 @@
...
+
+import lsst.utils.tests
+import lsst.afw.image as afwImage
+#import lsst.afw.geom as afwGeom
+import lsst.afw.math as afwMath
+#import lsst.meas.algorithms as measAlg
+import lsst.daf.base as dafBase
+import testImageDecorrelation as testID  # for makeFakeImages
+
+from lsst.ip.diffim.zogy import (ZogyTask, ZogyConfig, ZogyMapReduceConfig)
```
isullivan: Does this import need to be a tuple?

tests/testZogy.py
```diff
@@ -0,0 +1,209 @@
...
+        self.svar = svar  # variance of noise in science image
+        self.tvar = tvar  # variance of noise in template image
+
+        self.im1ex, self.im2ex \
+            = testID.makeFakeImages(svar=self.svar, tvar=self.tvar,
+                                    psf1=self.psf1_sigma, psf2=self.psf2_sigma,
+                                    n_sources=10, psf_yvary_factor=varyPsf,
+                                    seed=666, verbose=False)
+        # Create an array corresponding to the "expected" subtraction (noise only)
+        np.random.seed(666)
```
isullivan: Since you're setting this seed and the keyword seed on line 74 above to the same value, you should define `seed = 666` as a variable and reference that in both places.

python/lsst/ip/diffim/zogy.py
```diff
@@ -0,0 +1,839 @@
...
+
+All methods in `ZogyTask` assume template and science images are
+already accurately photometrically and astrometrically registered.
+
+`ZogyMapperSubtask` is a wrapper which runs `ZogyTask` in the
+`ImageMapReduce` framework, computing of ZOGY diffim's on small,
+overlapping sub-images, thereby enabling complete ZOGY diffim's which
+account for spatially-varying noise and PSFs across the two input
+exposures.
+"""
```
djreiss: A good one is in the unit test. I added a comment pointing to that.

python/lsst/ip/diffim/zogy.py
```diff
@@ -0,0 +1,839 @@
...
+        - D : 2D numpy.array, the proper image difference
+        - D_var : 2D numpy.array, the variance image for `D`
+        """
+        # Do all in fourier space (needs image-sized PSFs)
+        psf1 = self._padPsfToImageSize(self.im1_psf, self.im1)
+        psf2 = self._padPsfToImageSize(self.im2_psf, self.im2)
+
+        preqs = self.computePrereqs(psf1, psf2, padSize=0)  # already padded the PSFs
+
+        # First do the image
```
djreiss: Ah but they're not the same (I almost miss that) -- the diffim is a subtraction, the variance is an addition.

isullivan: Yes, which is why you have the flag, to switch between the + and - version. I am a strong proponent of using as much of the same code as possible to process the image and weights (in this case variance) planes, so in your function you could calculate term_1 and term_2 of the equation, and then based on the flag either add or subtract those two terms. This is just a preference on my part, so don't feel you have to re-write this if you think a function would be awkward here.

python/lsst/ip/diffim/zogy.py
```diff
@@ -0,0 +1,839 @@
...
+        Kr_hat = (preqs.Pr_hat + delta) / (preqs.denom + delta)
+        Kn_hat = (preqs.Pn_hat + delta) / (preqs.denom + delta)
+        Kr = np.fft.ifft2(Kr_hat).real
+        Kr = np.roll(np.roll(Kr, -1, 0), -1, 1)
+        Kn = np.fft.ifft2(Kn_hat).real
+        Kn = np.roll(np.roll(Kn, -1, 0), -1, 1)
+
+        # Using this block makes it slightly faster, but ~25% worse artifacts:
+        padSize = self.padSize if padSize is None else padSize
+        if debug and padSize > 0:
```
djreiss: If you're referring to the "if debug..." block, this doesn't set self.padSize to zero. Neither does the line above it (if that's the one you're actually talking about). 

But as you suggest below I will move it and the next block to a "filterPsf" function.

python/lsst/ip/diffim/zogy.py
```diff
@@ -0,0 +1,839 @@
...
+            ps = padSize
+            Kn = Kn[ps:-ps, ps:-ps]
+            Kr = Kr[ps:-ps, ps:-ps]
+
+        # Uncommenting this block makes it also slightly faster (zeros are ignored in convolution)
```
djreiss: The ability to have a "debug" option was discussed in an RFC and accepted. That's why it's here, but I agree that I will move the psf filtering as a separate method, and create a ticket to investigate the best way to move forward on this.

python/lsst/ip/diffim/zogy.py
isullivan: Sorry, I meant that this seems to achieve the same effect as setting `self.padSize=0`, so the block of code might not be necessary if you could get the same effect by doing that.

python/lsst/ip/diffim/zogy.py
djreiss: No, this squaring is actually performed in real space.

PR #63, opened 2017-06-05
DM-10009: Ensure masks are valid from ImageMapReduceTask

djreiss:
Ensure all pixels that are not touched by the reducer are flagged in masks

These pixels are likely to be the result of a bug in the mapper.

Also revent divide by zero when averaging results in weights of 0.

### DIFF
--- a/python/lsst/ip/diffim/imageMapReduce.py
+++ b/python/lsst/ip/diffim/imageMapReduce.py
@@ -170,6 +170,11 @@ class ImageReducerSubtaskConfig(pexConfig.Config):
                        entire input exposure""",
         }
     )
+    badMaskPlanes = pexConfig.ListField(
+        dtype=str,
+        doc="""Mask planes to set for invalid pixels""",
+        default=('INVALID_MAPREDUCE', 'BAD', 'NO_DATA')
+    )
 
 
 class ImageReducerSubtask(pipeBase.Task):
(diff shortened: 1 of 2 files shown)


python/lsst/ip/diffim/imageMapReduce.py
```diff
@@ -275,15 +276,36 @@ def run(self, mapperResults, exposure, **kwargs):
...
                     wtsView.getArray()[isNotNan] += 1
 
+        isNan = (np.isnan(newMI.getImage().getArray()) |
+                  np.isnan(newMI.getVariance().getArray()))
+        if np.sum(isNan) > 0:
+            # set mask to INVALID for pixels where produced exposure is NaN
+            mask = newMI.getMask()
+            # New mask plane - for debugging map-reduced images
+            mask.addMaskPlane('INVALID_MAPREDUCE')
+            bad = mask.getPlaneBitMask(['INVALID_MAPREDUCE', 'BAD', 'NO_DATA'])
```
natelust: These mask planes should not be hard coded in the routine, but configurable from the config class with these as defaults. In the future there may be some reason to set more.

python/lsst/ip/diffim/imageMapReduce.py
```diff
@@ -275,15 +276,36 @@ def run(self, mapperResults, exposure, **kwargs):
...
+
         if reduceOp == 'average':
             wts = weights.getArray().astype(np.float)
             self.log.info('AVERAGE: Maximum overlap: %f', wts.max())
             self.log.info('AVERAGE: Average overlap: %f', np.nanmean(wts))
+            self.log.info('AVERAGE: Minimum overlap: %f', wts.min())
+            wtsZero = wts == 0.
+            self.log.info('AVERAGE: Number of zero pixels: %f (%f%%)', np.sum(wtsZero),
+                          np.mean(wtsZero) * 100.)
+            wts[wtsZero] = 1e-29  # avoid division by zero warnings -- will set to nan and flag them below
```
natelust: This line is not needed and loops over pixels needlessly, see below

python/lsst/ip/diffim/imageMapReduce.py
```diff
@@ -275,15 +276,36 @@ def run(self, mapperResults, exposure, **kwargs):
...
             newMI.getImage().getArray()[:, :] /= wts
             newMI.getVariance().getArray()[:, :] /= wts
```
natelust: numpy ufuncs save us here, you can just do:
np.divide(newMI.getImage().getArray(), wts, out=newMI.getImage().getArray(), where=~wtsZero)
This will do inplace operation (output is the same as input) and only loop over and do a calculation where wtsZero is not true, all in one step.

python/lsst/ip/diffim/imageMapReduce.py
```diff
@@ -275,15 +276,36 @@ def run(self, mapperResults, exposure, **kwargs):
...
-            wtsZero = wts == 0.
-            newMI.getImage().getArray()[wtsZero] = newMI.getVariance().getArray()[wtsZero] = np.nan
-            # TBD: set mask to something for pixels where wts == 0. Shouldn't happen. (DM-10009)
+            if np.sum(wtsZero) > 0:
+                newMI.getImage().getArray()[wtsZero] = newMI.getVariance().getArray()[wtsZero] = np.nan
```
natelust: why are you setting these values to nan here? is that not redundant info with the mask plane bits set? Do we ever want nans in an image? Doesn't that make things like afw statistics need to do more work?

python/lsst/ip/diffim/imageMapReduce.py
```diff
@@ -275,15 +276,36 @@ def run(self, mapperResults, exposure, **kwargs):
...
+            mask.getArray()[isNan] |= bad
+
         if reduceOp == 'average':
             wts = weights.getArray().astype(np.float)
             self.log.info('AVERAGE: Maximum overlap: %f', wts.max())
             self.log.info('AVERAGE: Average overlap: %f', np.nanmean(wts))
+            self.log.info('AVERAGE: Minimum overlap: %f', wts.min())
+            wtsZero = wts == 0.
+            self.log.info('AVERAGE: Number of zero pixels: %f (%f%%)', np.sum(wtsZero),
+                          np.mean(wtsZero) * 100.)
```
natelust: Also just consider using afw statistics directly, I think you can have it build and return multiple statistics all at once.

tests/testImageMapReduce.py
```diff
@@ -377,6 +377,35 @@ def testGridCentroidsWrongLength(self):
...
+        isnan = np.isnan(newArr)
+        self.assertGreater(np.sum(isnan), 1000)
+
+        mi = self.exposure.getMaskedImage().getImage().getArray()
+        self.assertFloatsAlmostEqual(mi[~isnan], newArr[~isnan] - 5.)
+
+        mask = newMI.getMask()  # Now check the mask
+        self.assertGreater(mask.getMaskPlane('INVALID_MAPREDUCE'), 0)
+        maskBit = mask.getPlaneBitMask('INVALID_MAPREDUCE')
+        nMasked = np.sum((mask.getArray()[:, :] & maskBit) != 0)
```
natelust: Im pretty sure here the operation you want is numpy.bitwise_and, you can avoid the [:,:] in that case.

python/lsst/ip/diffim/imageMapReduce.py
```diff
@@ -261,6 +261,7 @@ def run(self, mapperResults, exposure, **kwargs):
             patchMI = item.getMaskedImage()
             isNotNan = ~(np.isnan(patchMI.getImage().getArray()) |
                          np.isnan(patchMI.getVariance().getArray()))
```
natelust: See my comment below about multiplying the arrays together to prevent needless loops. If you used np.isfinite instead of np.isnan you can save the inverting loop (if inf is also not a valid value).

djreiss: At risk of premature optimization, I will implement everything you've recommended here.

python/lsst/ip/diffim/imageMapReduce.py
```diff
@@ -275,15 +276,36 @@ def run(self, mapperResults, exposure, **kwargs):
...
+            wtsZero = wts == 0.
+            self.log.info('AVERAGE: Number of zero pixels: %f (%f%%)', np.sum(wtsZero),
+                          np.mean(wtsZero) * 100.)
+            wts[wtsZero] = 1e-29  # avoid division by zero warnings -- will set to nan and flag them below
             newMI.getImage().getArray()[:, :] /= wts
             newMI.getVariance().getArray()[:, :] /= wts
-            wtsZero = wts == 0.
-            newMI.getImage().getArray()[wtsZero] = newMI.getVariance().getArray()[wtsZero] = np.nan
-            # TBD: set mask to something for pixels where wts == 0. Shouldn't happen. (DM-10009)
+            if np.sum(wtsZero) > 0:
```
djreiss: thought about moving it higher, but the logic doesn't work. But your second comment is valid.

python/lsst/ip/diffim/imageMapReduce.py
djreiss: Only works as:
```
tmp = newMI.getImage().getArray()
np.divide(tmp, wts, out=tmp, where=~wtsZero)
```
will do it this way.

python/lsst/ip/diffim/imageMapReduce.py
djreiss: Not sure about your last question. This was for the case where I used `wts[wtsZero] = 1e-29` . Since you helped me get rid of that, this line can go now too.

PR #64, opened 2017-06-05
DM-10803: Refactor grid generation in ImageMapReduceTask

djreiss:
- Rename 'gridCentroidsX/Y' to 'cellCentroidsX/Y'
- Rename 'gridSizeX/Y' to 'cellSizeX/Y'
- Refactor grid generation to iterate over centroids rather than
   offsetting bounding boxes

### DIFF
--- a/python/lsst/ip/diffim/imageMapReduce.py
+++ b/python/lsst/ip/diffim/imageMapReduce.py
@@ -57,7 +57,7 @@ two subtasks, `ImageMapperSubtask` and `ImageReducerSubtask`.
 dimensions to be processed, and then calls the `run` methods of the
 `ImageMapperSubtask` and `ImageReducerSubtask` on those sub-exposures.
 `ImageMapReduceTask` may be configured with a list of sub-exposure
-centroids (`config.gridCentroidsX` and `config.gridCentroidsY`) and a
+centroids (`config.cellCentroidsX` and `config.cellCentroidsY`) and a
 single pair of bounding boxes defining their dimensions, or a set of
 parameters defining a regular grid of centroids (`config.gridStepX`
 and `config.gridStepY`).
@@ -357,11 +357,11 @@ class ImageMapReduceConfig(pexConfig.Config):
         target=ImageReducerSubtask,
     )
 
-    # Separate gridCentroidsX and gridCentroidsY since pexConfig.ListField accepts limited dtypes
+    # Separate cellCentroidsX and cellCentroidsY since pexConfig.ListField accepts limited dtypes
     #  (i.e., no Point2D). The resulting set of centroids is the "vertical stack" of
-    #  `gridCentroidsX` and `gridCentroidsY`, i.e. for (1,2), (3,4) respectively, the
+    #  `cellCentroidsX` and `cellCentroidsY`, i.e. for (1,2), (3,4) respectively, the
     #   resulting centroids are ((1,3), (2,4)).
-    gridCentroidsX = pexConfig.ListField(
+    cellCentroidsX = pexConfig.ListField(
         dtype=float,
         doc="""Input X centroids around which to place subimages.
                If None, use grid config options below.""",
@@ -369,7 +369,7 @@ class ImageMapReduceConfig(pexConfig.Config):
         default=None
     )
 
-    gridCentroidsY = pexConfig.ListField(
+    cellCentroidsY = pexConfig.ListField(
         dtype=float,
         doc="""Input Y centroids around which to place subimages.
                If None, use grid config options below.""",
@@ -377,14 +377,14 @@ class ImageMapReduceConfig(pexConfig.Config):
         default=None
     )
 
-    gridSizeX = pexConfig.Field(
+    cellSizeX = pexConfig.Field(
         dtype=float,
         doc="""Dimensions of each grid cell in x direction""",
         default=10.,
         check=lambda x: x > 0.
     )
 
-    gridSizeY = pexConfig.Field(
+    cellSizeY = pexConfig.Field(
         dtype=float,
         doc="""Dimensions of each grid cell in y direction""",
         default=10.,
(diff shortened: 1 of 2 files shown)


python/lsst/ip/diffim/imageMapReduce.py
```diff
@@ -622,146 +622,121 @@ def _generateGrid(self, exposure, forceEvenSized=False, **kwargs):
...
         adjustGridOption = self.config.adjustGridOption
         scaleByFwhm = self.config.scaleByFwhm
-        bbox = exposure.getBBox()
 
-        psfFwhm = (exposure.getPsf().computeShape().getDeterminantRadius() *
-                   2. * np.sqrt(2. * np.log(2.)))
-        if scaleByFwhm:
-            self.log.info("Scaling grid parameters by %f" % psfFwhm)
+        if cellCentroidsX is None or len(cellCentroidsX) <= 0:
+            # No given centroids; construct them from cellSize/gridStep
```
parejkoj: "Not given centroids"?

python/lsst/ip/diffim/imageMapReduce.py
```diff
@@ -622,146 +622,121 @@ def _generateGrid(self, exposure, forceEvenSized=False, **kwargs):
...
 
-        def rescaleValue(val):
+            psfFwhm = (exposure.getPsf().computeShape().getDeterminantRadius() *
+                       2. * np.sqrt(2. * np.log(2.)))
```
parejkoj: No spaces around multiplication.

python/lsst/ip/diffim/imageMapReduce.py
```diff
@@ -622,146 +622,121 @@ def _generateGrid(self, exposure, forceEvenSized=False, **kwargs):
...
+                    return np.rint(val).astype(int)
+
+            cellSizeX = rescaleValue(cellSizeX)
+            cellSizeY = rescaleValue(cellSizeY)
+            gridStepX = rescaleValue(gridStepX)
+            gridStepY = rescaleValue(gridStepY)
+            borderSizeX = rescaleValue(borderSizeX)
+            borderSizeY = rescaleValue(borderSizeY)
+
+            nGridX = bbox.getWidth() // gridStepX
```
parejkoj: No spaces around division (floor, or otherwise). Also below.

python/lsst/ip/diffim/imageMapReduce.py
```diff
@@ -622,146 +622,121 @@ def _generateGrid(self, exposure, forceEvenSized=False, **kwargs):
...
+            nGridY = bbox.getHeight() // gridStepY
+
+            if adjustGridOption == 'spacing':
+                # Readjust spacings so that they fit perfectly in the image.
+                nGridX = bbox.getWidth() // cellSizeX + 1
+                nGridY = bbox.getHeight() // cellSizeY + 1
+                cellCentroids = [(x, y) for x in np.linspace(cellSizeX//2, bbox.getWidth() - cellSizeX//2,
```
parejkoj: Double-nested list comprehensions are a bit unpleasant. I think what you want to do is use `np.meshgrid` here. Possibly:

```python
np.meshgrid(np.linspace(...), np.linspace(...))
```

See: https://docs.scipy.org/doc/numpy/reference/generated/numpy.meshgrid.html

python/lsst/ip/diffim/imageMapReduce.py
```diff
@@ -622,146 +622,121 @@ def _generateGrid(self, exposure, forceEvenSized=False, **kwargs):
...
-        if self.config.gridCentroidsX is not None and len(self.config.gridCentroidsX) > 0:
-            for i, centroidX in enumerate(self.config.gridCentroidsX):
-                centroidY = self.config.gridCentroidsY[i]
+        def _makeBoxEvenSized(bb):
+            """Force a bounding-box to have dimensions that are modulo 2."""
+
+            if bb.getWidth() % 2 == 1:  # grow to the right
+                bb.include(afwGeom.Point2I(bb.getMaxX()+1, bb.getMaxY()))  # Expand by 1 pixel!
+                bb.clip(bbox)
+                if bb.getWidth() % 2 == 1:  # clipped at right -- so grow to the left
```
parejkoj: Could it ever be clipped on both sides, such that this this expansion also fails?

python/lsst/ip/diffim/imageMapReduce.py
```diff
@@ -622,146 +622,121 @@ def _generateGrid(self, exposure, forceEvenSized=False, **kwargs):
...
+                    bb.include(afwGeom.Point2I(bb.getMaxX(), bb.getMinY()-1))
+                    bb.clip(bbox)
+
+            return bb
+
+        # Use given or grid-parameterized centroids as centers for bounding boxes
+        if cellCentroidsX is not None and len(cellCentroidsX) > 0:
+            for i, centroidX in enumerate(cellCentroidsX):
+                centroidY = cellCentroidsY[i]
                 centroid = afwGeom.Point2D(centroidX, centroidY)
```
parejkoj: If you make a meshgrid above, you should be able to do:

```
for xy in cellCentroids.flatten():
   centroid = afwGeom.Point2D(xy[0], xy[1])
...
```

djreiss: This doesn't work as meshgrid returns a list of arrays.

djreiss: Plus, the grid may not actually be a grid (if the config-ed `cellCentroidsX` and `cellCentroidsY` are not None). So I'm leaving as-is but refactoring a bit more to reduce code duplication.

python/lsst/ip/diffim/imageMapReduce.py
djreiss: See my comment below for why.

python/lsst/ip/diffim/imageMapReduce.py
djreiss: Hmm, it's only possible if the cell size is > the image size. This shouldn't happen but I suppose if it does I can shrink the `bb` instead of growing it. 

Ugh, `Box2D`s are not very easily manipulatable. I'm just going to raise an exception if it is not even-sized after these two attempts, let's see how often it actually happens.

PR #66, opened 2017-06-06
DM-10805: Spatially-varying ZOGY option

djreiss:
- Include unit tests for new task


python/lsst/ip/diffim/zogy.py
```diff
@@ -288,8 +319,8 @@ def computePrereqs(self, psf1=None, psf2=None, padSize=0):
         padSize = self.padSize if padSize is None else padSize
         Pr, Pn = psf1, psf2
         if padSize > 0:
-            Pr = ZogyTask._padPsfToSize(psf1, (padSize, padSize))
-            Pn = ZogyTask._padPsfToSize(psf2, (padSize, padSize))
+            Pr = ZogyTask._padPsfToSize(psf1, (psf1.shape[0]+padSize, psf1.shape[0]+padSize))
```
yalsayyad: Did you mean: `(psf1.shape[0]+padSize, psf1.shape[1]+padSize)` or is there something special about the `x` dimension. Please comment if so.

python/lsst/ip/diffim/zogy.py
```diff
@@ -898,3 +949,133 @@ class ZogyMapReduceConfig(ImageMapReduceConfig):
...
+        def gv(exp):
+            return exp.getMaskedImage().getImage().getArray()
+
+        if spatiallyVarying:
+            config = self.config.zogyMapReduceConfig
+            task = ImageMapReduceTask(config=config)
+            results = task.run(scienceExposure, template=templateExposure, inImageSpace=inImageSpace,
+                               doScorr=doPreConvolve, forceEvenSized=True)
+            results.D = results.exposure
+            # The CoaddPsf apparently cannot be used for detection as it doesn't have a
```
yalsayyad: Is this comment about CoaddPsf still true?

PaulPrice: I don't think it was ever true. We have used `CoaddPsf` for detection from its genesis.
If I'm wrong, please file a ticket on meas_algorithms.

djreiss: My understanding from looking at the code was that it computes a single psf from the `averagePosition` and uses that for detection. But it is still a good point that the comment is untrue.

python/lsst/ip/diffim/zogy.py
```diff
@@ -898,3 +949,133 @@ class ZogyMapReduceConfig(ImageMapReduceConfig):
...
+        self.zogyMapReduceConfig.reducerSubtask.reduceOperation = 'average'
+
+
+class ZogyImagePsfMatchTask(ImagePsfMatchTask):
+    ConfigClass = ZogyImagePsfMatchConfig
+
+    def __init__(self, *args, **kwargs):
+        ImagePsfMatchTask.__init__(self, *args, **kwargs)
+
+    def setDefaults(self):
```
yalsayyad: Does this function do anything? I thought only Configs run `setDefaults`.

python/lsst/ip/diffim/zogy.py
```diff
@@ -898,3 +949,133 @@ class ZogyMapReduceConfig(ImageMapReduceConfig):
...
+            if not doPreConvolve:
+                results = task.computeDiffim(inImageSpace=inImageSpace)
+                results.matchedExposure = results.R
+            else:
+                results = task.computeScorr(inImageSpace=inImageSpace)
+                results.D = results.S
+
+        # Make sure masks of input images are propagated to diffim
+        mask = results.D.getMaskedImage().getMask()
+        badBits = mask.getPlaneBitMask(['UNMASKEDNAN', 'NO_DATA', 'BAD', 'EDGE', 'SUSPECT', 'CR', 'SAT'])
```
yalsayyad: badBits not used.

python/lsst/ip/diffim/zogy.py
```diff
@@ -898,3 +949,133 @@ class ZogyMapReduceConfig(ImageMapReduceConfig):
...
+        badBitsNan = mask.getPlaneBitMask(['UNMASKEDNAN'])
```
yalsayyad: See comment on jira ticket about UNMASKEDNAN. Not sure if calexps/exposures are guaranteed to have this. Check to see if this plane is present first before using it

python/lsst/ip/diffim/zogy.py
```diff
@@ -898,3 +949,133 @@ class ZogyMapReduceConfig(ImageMapReduceConfig):
...
+                                         exposure.getMaskedImage().getMask(),
+                                         afwMath.MEANCLIP|afwMath.MEDIAN, statsControl)
+        mn = statObj.getValue(afwMath.MEANCLIP)
+        med = statObj.getValue(afwMath.MEDIAN)
+        return mn, med
+
+    def subtractExposures(self, templateExposure, scienceExposure,
+                          doWarping=True, spatiallyVarying=True, inImageSpace=False,
+                          doPreConvolve=False):
+
```
yalsayyad: Needs docstring (esp. to make it clear that this function modifies the input exposures)

python/lsst/ip/diffim/zogy.py
```diff
@@ -898,3 +949,133 @@ class ZogyMapReduceConfig(ImageMapReduceConfig):
...
+                # Also warp the PSF
+                xyTransform = afwImage.XYTransformFromWcsPair(scienceExposure.getWcs(),
+                                                              templateExposure.getWcs())
+                psfWarped = measAlg.WarpedPsf(templateExposure.getPsf(), xyTransform)
+                templateExposure = self._warper.warpExposure(scienceExposure.getWcs(),
+                                                             templateExposure,
+                                                             destBBox=scienceExposure.getBBox())
+
+                templateExposure.setPsf(psfWarped)
+                templateExposure.writeFits('WARPEDTEMPLATE_ZOGY.fits')
```
yalsayyad: Tasks shouldn't write exposures to the current working directory.  If you really need this, use the lsstDebug framework to tuck it in an if statement so that it only writes out if requested.

python/lsst/ip/diffim/zogy.py
```diff
@@ -898,3 +949,133 @@ class ZogyMapReduceConfig(ImageMapReduceConfig):
...
+        badBits = mask.getPlaneBitMask(['UNMASKEDNAN', 'NO_DATA', 'BAD', 'EDGE', 'SUSPECT', 'CR', 'SAT'])
+        badBitsNan = mask.getPlaneBitMask(['UNMASKEDNAN'])
+        mask |= gm(scienceExposure)
+        mask |= gm(templateExposure)
+        gm(results.D)[:, :] = mask
+        gm(results.D).getArray()[np.isnan(ga(results.D))] |= badBitsNan
+        gm(results.D).getArray()[np.isnan(ga(scienceExposure))] |= badBitsNan
+        gm(results.D).getArray()[np.isnan(ga(templateExposure))] |= badBitsNan
+
+        #results = pipeBase.Struct(exposure=D)
```
yalsayyad: Remove

python/lsst/ip/diffim/zogy.py
```diff
@@ -898,3 +949,133 @@ class ZogyMapReduceConfig(ImageMapReduceConfig):
...
+
+                templateExposure.setPsf(psfWarped)
+                templateExposure.writeFits('WARPEDTEMPLATE_ZOGY.fits')
+            else:
+                self.log.error("ERROR: Input images not registered")
+                raise RuntimeError("Input images not registered")
+
+        def gm(exp):
+            return exp.getMaskedImage().getMask()
+        def ga(exp):
```
yalsayyad: + line between functions (even inner functions).

python/lsst/ip/diffim/zogy.py
```diff
@@ -898,3 +949,133 @@ class ZogyMapReduceConfig(ImageMapReduceConfig):
...
+            return exp.getMaskedImage().getImage().getArray()
+        def gv(exp):
```
yalsayyad: gv is not used and is exactly the same as ga.

RobertLuptonTheGood: Also, user convenience functions like these make the code harder to read.  Please don't do this.

Also, with pybind11, you can write this as e.g. `exp.maskedImage.mask` which might make this more palatable.

python/lsst/ip/diffim/zogy.py
```diff
@@ -898,3 +949,133 @@ class ZogyMapReduceConfig(ImageMapReduceConfig):
...
+                results = task.computeScorr(inImageSpace=inImageSpace)
+                results.D = results.S
+
+        # Make sure masks of input images are propagated to diffim
+        mask = results.D.getMaskedImage().getMask()
+        badBits = mask.getPlaneBitMask(['UNMASKEDNAN', 'NO_DATA', 'BAD', 'EDGE', 'SUSPECT', 'CR', 'SAT'])
+        badBitsNan = mask.getPlaneBitMask(['UNMASKEDNAN'])
+        mask |= gm(scienceExposure)
+        mask |= gm(templateExposure)
+        gm(results.D)[:, :] = mask
```
yalsayyad: `mask` here is a view (pointer) of `results.D`'s mask  (i.e. the two lines before have already modified it)  This line is unnecessary as it just copies itself to itself.

python/lsst/ip/diffim/zogy.py
```diff
@@ -478,7 +516,8 @@ def _setNewPsf(self, exposure, psfArr):
         exposure.setPsf(psfNew)
         return exposure
 
-    def computeDiffim(self, inImageSpace=None, padSize=None, **kwargs):
+    def computeDiffim(self, inImageSpace=None, padSize=None,
+                      returnMatchedTemplate=False, **kwargs):
```
yalsayyad: Add returnMatchedTemplate to docstring.

python/lsst/ip/diffim/zogy.py
```diff
@@ -898,3 +949,133 @@ class ZogyMapReduceConfig(ImageMapReduceConfig):
...
+        #results = pipeBase.Struct(exposure=D)
+        results.subtractedExposure = results.D
+        #results.matchedExposure = results.R
+        results.warpedExposure = templateExposure
+        return results
+
+    def subtractMaskedImages(self, templateExposure, scienceExposure,
+                             doWarping=True, spatiallyVarying=True, inImageSpace=False,
+                             doPreConvolve=False):
+        pass  # not implemented
```
yalsayyad: Consider if `raise NotImplementedError` would be better here.

python/lsst/ip/diffim/zogy.py
djreiss: Unless I'm missing something that I did (wrong), I don't *think* it modifies the input exposures. I will surely add the docstring.

python/lsst/ip/diffim/zogy.py
djreiss: Also, redundant, so removed.

python/lsst/ip/diffim/zogy.py
djreiss: Whoops, meant to remove that, thanks for catching it.

python/lsst/ip/diffim/zogy.py
djreiss: This arose from a desire to force it to be square. However, that is enforced below, so I will update it here to avoid confusion.

PR #71, opened 2017-08-29
DM-11620: Make test_dipoleFitter.py more robust

laurenam:
This includes setting the random seed explicitly in the test dataset
realization code and allowing for more configs to be specified in the
call to the detectDipoleSources() function in utils.py.  In particular,
the re-estimation of the background fails in test_dipoleFitter.py when
weighting is turned on, so wether to weight when approximating the
background is now controlled by an input parameter to the function (set
to True by default, but False for the test case here).

I also allowed for the default binSize for background estimation to be
adjusted as the current default of 128 is not appropriate for test
images typically on the order of 100x100 pixels.  To avoid decreasing
the binSize by too much, a minimum binSize input parameter is provided.

### DIFF
--- a/python/lsst/ip/diffim/utils.py
+++ b/python/lsst/ip/diffim/utils.py
@@ -768,11 +768,12 @@ class DipoleTestImage(object):
     def _makeDipoleImage(self):
         """!Generate an exposure and catalog with the given dipole source(s)"""
 
+        # Must seed the pos/neg images with different values to ensure they get different noise realizations
         posImage, posCatalog = self._makeStarImage(
-            xc=self.xcenPos, yc=self.ycenPos, flux=self.flux)
+            xc=self.xcenPos, yc=self.ycenPos, flux=self.flux, randomSeed=111)
 
         negImage, negCatalog = self._makeStarImage(
-            xc=self.xcenNeg, yc=self.ycenNeg, flux=self.fluxNeg)
+            xc=self.xcenNeg, yc=self.ycenNeg, flux=self.fluxNeg, randomSeed=222)
 
         dipole = posImage.clone()
         di = dipole.getMaskedImage()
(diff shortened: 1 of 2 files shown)


python/lsst/ip/diffim/utils.py
```diff
@@ -820,14 +820,40 @@ def fitDipoleSource(self, source, **kwds):
...
 
         Detect pos/neg sources in the diffim, then merge them. A
         bigger "grow" parameter leads to a larger footprint which
         helps with dipole measurement for faint dipoles.
-        """
 
+        Parameters
+        ----------
+        doMerge : `bool`
+           Wether to merge the positive and negagive detections into a single source table
```
jdswinbank: "Whether" (and below).

python/lsst/ip/diffim/utils.py
```diff
@@ -820,14 +820,40 @@ def fitDipoleSource(self, source, **kwds):
...
+        diffim : `lsst.afw.image.exposure.exposure.ExposureF`
+           Difference image on which to perform detection
+        detectSigma : `float`
+           Threshold for object detection
+        grow : `int`
+           Number of pixels to grow the footprints before merging
+        weighting : `bool`
+           Wether to use weighting in the background (re)estimation (only applies is useApprox=True)
```
jdswinbank: What is `useApprox`? I'd naïvely expect it to be another argument to `detectDipoleSources`, but it isn't.

(In fact, I bet I can find it buried in the detection task config, but that's not clear from this docstring.)

python/lsst/ip/diffim/utils.py
```diff
@@ -820,14 +820,40 @@ def fitDipoleSource(self, source, **kwds):
...
+        minBinSize : `int`
+           Minimum bin size for the background (re)estimation (only applies if the default leads to
+           nBinX, nBinY < fit order so the default config parameter needs to be decreased, but not
```
jdswinbank: I think it's clearer to actually use "min" in this docstring, ie "min(nBinX, nBinY) < fit order", or "the smaller of nBinX and nBinY is less than the fit order". (I think that's what you're actually testing, right?)

python/lsst/ip/diffim/utils.py
laurenam: Sorry, that's a background config:
https://github.com/lsst/meas_algorithms/blob/master/python/lsst/meas/algorithms/subtractBackground.py#L98
        doc="Use Approximate (Chebyshev) to model background."
This might be clearer (at least tells you where to look without being too verbose)?:
(only applies if detectConfig.background.useApprox=True)

PR #73, opened 2017-09-19
DM-11951: Fixes in AL Decorrelation and Zogy

djreiss:


python/lsst/ip/diffim/imageDecorrelation.py
```diff
@@ -223,21 +233,32 @@ def run(self, exposure, templateExposure, subtractedExposure, psfMatchingKernel,
...
-        kft = scipy.fftpack.fft2(kappa)
-        kft = np.sqrt((svar + tvar) / (svar + tvar * kft**2))
-        pck = scipy.fftpack.ifft2(kft)
-        pck = scipy.fftpack.ifftshift(pck.real)
+        kft = np.fft.fft2(kappa)
+        kft2 = np.conj(kft) * kft
+        denom = svar + tvar * kft2
+        if preConvKernel is not None:
+            mk = DecorrelateALKernelTask._fixOddKernel(kft)
+            mk[mk < 1e-3] = 1e-3  # Psf should not be < 0; and messes up denominator
```
r-owen: Please hard-coding the limit as a constant (e.g. `double MIN_KERNEL = 1.0e-3`) and document it where you define it. That keeps the code consistent in case you decide to change the limit -- the number only appears once.

Is 0.001 sufficiently small to be negligible? Some kernels can easily have 1000-ish pixels.

Given the code, I'm guessing kernel <= 0 messes up the denominator? Because if only negative values were a problem I would have expected the test to replace negative values with 0. If my supposition is correct, please update the comment, e.g. `...should not be <= 0...`.

python/lsst/ip/diffim/imageDecorrelation.py
```diff
@@ -223,21 +233,32 @@ def run(self, exposure, templateExposure, subtractedExposure, psfMatchingKernel,
...
+            mk = np.fft.fft2(mk)
+            mk2 = np.conj(mk) * mk
+            denom = svar * mk2 + tvar * kft2
+            denom[np.abs(denom) < 1e-3] = 1e-3  # fixes some aliasing issues
```
r-owen: Is this `1e-3` intrinsically the same as the `1e-3` above for `mk` or an unrelated constant? If different, please define a new constant with documentation for why you chose the value.

python/lsst/ip/diffim/imageDecorrelation.py
```diff
@@ -589,8 +609,8 @@ def run(self, scienceExposure, templateExposure, subtractedExposure, psfMatching
            by `ip_diffim.ImagePsfMatchTask.subtractExposures()`
         spatiallyVarying : bool
            if True, perform the spatially-varying operation
-        doPreConvolve : bool
-           if True, the scienceExposure has been pre-filtered with its PSF. (Currently
+        preConvKernel : bool
```
r-owen: Please correct the type (it's not `bool`)

python/lsst/ip/diffim/imageDecorrelation.py
```diff
@@ -601,7 +621,11 @@ def run(self, scienceExposure, templateExposure, subtractedExposure, psfMatching
         self.log.info('Running A&L decorrelation: spatiallyVarying=%r' % spatiallyVarying)
 
         svar = self.computeVarianceMean(scienceExposure)
+        if np.isnan(svar):  # Should not happen unless entire image has been masked.
+            svar = 1e-9
```
r-owen: As before, please make sure the entire image really has been masked.

python/lsst/ip/diffim/zogy.py
```diff
@@ -231,23 +230,50 @@ def selectPsf(psf, exposure):
...
+                psf1[1:, :] = psf1[:-1, :]
+            if maxLoc1[1] > maxLoc2[1]:
+                psf2[:, 1:] = psf2[:, :-1]
+            elif maxLoc1[1] < maxLoc2[1]:
+                psf1[:, 1:] = psf1[:, :-1]
+            maxLoc1 = np.unravel_index(np.argmax(psf1), psf1.shape)
+            maxLoc2 = np.unravel_index(np.argmax(psf2), psf2.shape)
+
+        # Make sure there are no div-by-zeros
+        psf1[psf1 < 1e-4] = 1e-4
```
r-owen: This is different than the `1e-3` you used in other code. Using a named constant would help, and it may be worth getting the constant that you use here from the value you set in the C++ code.

python/lsst/ip/diffim/zogy.py
```diff
@@ -687,6 +723,20 @@ def computeScorrFourierSpace(self, xVarAst=0., yVarAst=0., **kwargs):
         - S_var : the corrected variance image (denominator of eq. 25 of ZOGY (2016))
         - Dpsf : the PSF of the diffim D, likely never to be used.
         """
+        # Some masked regions are NaN or infinite!, and FFTs no likey.
+        def fix_nans(im):
+            """Replace any NaNs or Infs with the mean of the image."""
+            isbad = np.isinf(im) | np.isnan(im)
```
r-owen: Consider `isbad = not np.isfinite(im)` as it handles a few more cases

python/lsst/ip/diffim/imageDecorrelation.py
```diff
@@ -476,9 +496,9 @@ class DecorrelateALKernelSpatialConfig(pexConfig.Config):
     )
 
     def setDefaults(self):
-        self.decorrelateMapReduceConfig.gridStepX = self.decorrelateMapReduceConfig.gridStepY = 19
-        self.decorrelateMapReduceConfig.cellSizeX = self.decorrelateMapReduceConfig.cellSizeY = 20
-        self.decorrelateMapReduceConfig.borderSizeX = self.decorrelateMapReduceConfig.borderSizeY = 6
+        self.decorrelateMapReduceConfig.gridStepX = self.decorrelateMapReduceConfig.gridStepY = 40
+        self.decorrelateMapReduceConfig.cellSizeX = self.decorrelateMapReduceConfig.cellSizeY = 41
+        self.decorrelateMapReduceConfig.borderSizeX = self.decorrelateMapReduceConfig.borderSizeY = 8
```
r-owen: This code appears to be repeated below (line 1042). Can it be centralized?

python/lsst/ip/diffim/imageDecorrelation.py
```diff
@@ -198,14 +198,24 @@ def run(self, exposure, templateExposure, subtractedExposure, psfMatchingKernel,
 
         if svar is None:
             svar = self.computeVarianceMean(exposure)
+        if np.isnan(svar):  # Should not happen unless entire image has been masked.
+            svar = 1e-9
```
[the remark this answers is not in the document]
djreiss: I understand your point. The cases where I have had this happen is when using the ImageMapReduce framework (which does these computations on a grid of subimages) when a subimage is covered by a mask. Now that you mention it, it can be short-circuited and just return a copy of the input subtractedExposure, which is all NaNs and/or masked. I'll do that.

python/lsst/ip/diffim/imageDecorrelation.py
djreiss: All good points. Yes, it seems that 1e-3 is small enough -- smaller and strangeness can happen in the FFTs.

python/lsst/ip/diffim/imageDecorrelation.py
djreiss: I think you mean it's also in the `zogy.py` code. I can centralize it as the default in the `imageMapReduce.ImageMapReduceConfig`. However, there might be a reason to have these be different in the future. Right now I'm just finding that the same values work for both tasks. Do you have a suggestion about where to centralize this?

python/lsst/ip/diffim/zogy.py
djreiss: You're right, there's no logical reason for them to be different so I'll set them all to 1.0e-4. But there's not technically a reason for why they should be the same (the algorithms are different, they just suffer from related issues if the PSF is noisy).
Also I'm confused - there is no C++ code here, so can I set a global python variable and use it in both places?

python/lsst/ip/diffim/zogy.py
djreiss: Good advice! Should that be ~np.isfinite(im)  (element-wise negation)?

python/lsst/ip/diffim/imageDecorrelation.py
```diff
@@ -24,7 +24,6 @@
 #
 
 import numpy as np
-import scipy.fftpack
```
r-owen: Please also remove the lines
```
from future import standard_library
standard_library.install_aliases()
```
from near the top of the file -- these are no longer needed and make a hash of flake8 linting.

python/lsst/ip/diffim/imageDecorrelation.py
r-owen: It's a good question. It *looks* like the defaults here actually belong to the config `self.decorrelateMapReduceConfig`, which is  a `DecorrelateALKernelMapReduceConfig`, rather than this config. If so, and if that is true for `zogy.py` then I suggest those two map-reduce config classes inherit from a new common base class (e.g. `BaseMapReduceConfig`) that defines these common fields and assigns suitable defaults. Then in the future if one wishes to depart from any of these defaults, one can do that in the appropriate subclass or here.

For this to work `setDefaults` will have to call setDefaults on contained configs -- which is a good thing to do in any case.

However, such a change could be done as part of the refactoring. It's just the since you had to change the values in 2 places, that's a red flag.

python/lsst/ip/diffim/zogy.py
r-owen: Yes. Pity it looks like bitwise negation. `np.logical_not(...)` is more verbose but unambiguous. You decide.

python/lsst/ip/diffim/zogy.py
r-owen: Sorry, Python, not C++. Yes you could set one global python variable and use it in both places -- that makes sense if you can come up with a name that seems usable in the global API. If not, then a constant per module that is not exported would also be fine.

PR #74, opened 2017-09-21
DM-3704: Refactor ImageDifferenceTask

djreiss:
- Fix imports, configs, schemas
- Get individual command-line tasks to work
- Refactor runMatchSources in new function
- Start fixing docstrings
- Fix match-sources to correctly load ref-cat if avail
- Make tasks more robust
- Ensure saving of masked/subtracted exposures and diaSrces
  when run from imageDifference.py
- Robustify Zogy in spatially varying mode
- Allow no matchedExposure for use in dipole fitting (it is not
  produced by Zogy(spatiallyVarying)
- Improve logging messages

### DIFF
--- a/python/lsst/ip/diffim/imageDifference.py
+++ b/python/lsst/ip/diffim/imageDifference.py
@@ -1,0 +1,327 @@
+#
+# LSST Data Management System
+# Copyright 2012 LSST Corporation.
+#
+# This product includes software developed by the
+# LSST Project (http://www.lsst.org/).
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.    See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the LSST License Statement and
+# the GNU General Public License along with this program.  If not,
+# see <http://www.lsstcorp.org/LegalNotices/>.
+#
+from __future__ import absolute_import, division, print_function
+
+import lsst.pex.config as pexConfig
+import lsst.pipe.base as pipeBase
+import lsst.daf.base as dafBase
+import lsst.afw.table as afwTable
+from . import MakeDiffimTask, ProcessDiffimTask, GetCoaddAsTemplateTask, \
+    KernelCandidateF, SourceFlagChecker, DipoleAnalysis
+from . import diffimTools
+from . import utils as diUtils
+
+
+class ImageDifferenceConfig(pexConfig.Config):
+    """Config for ImageDifferenceTask
+    """
+    doAddCalexpBackground = pexConfig.Field(dtype=bool, default=True,
+                                            doc="Add background to calexp before processing it.  "
+                                            "Useful as ipDiffim does background matching.")
+
+    doDetection = pexConfig.Field(dtype=bool, default=True, doc="Detect sources?")
+
+    coaddName = pexConfig.Field(
+        doc="coadd name: typically one of deep or goodSeeing",
+        dtype=str,
+        default="deep",
+    )
+
+    makeDiffim = pexConfig.ConfigurableField(
+        target=MakeDiffimTask,
+        doc="Perform image subtraction"
+    )
+
+    processDiffim = pexConfig.ConfigurableField(
+        target=ProcessDiffimTask,
+        doc="Process subtracted image: detect and measure DIASources"
+    )
+
+    getTemplate = pexConfig.ConfigurableField(
+        target=GetCoaddAsTemplateTask,
+        doc="Subtask to retrieve template exposure and sources",
+    )
+
+    def setDefaults(self):
+        # Add filtered flux measurement, the correct measurement for pre-convolved images.
+        # Enable all measurements, regardless of doPreConvolve, as it makes data harvesting easier.
+        # To change that you must modify algorithms.names in the task's applyOverrides method,
+        # after the user has set doPreConvolve.
+        self.processDiffim.measurement.algorithms.names.add('base_PeakLikelihoodFlux')
+
+    def validate(self):
+        pexConfig.Config.validate(self)
+        if self.processDiffim.doMeasurement and not self.doDetection:
+            raise ValueError("Cannot run source measurement without source detection.")
+        if self.processDiffim.doMerge and not self.doDetection:
+            raise ValueError("Cannot run source merging without source detection.")
+
+
+class ImageDifferenceTaskRunner(pipeBase.ButlerInitializedTaskRunner):
+
+    @staticmethod
+    def getTargetList(parsedCmd, **kwargs):
+        return pipeBase.TaskRunner.getTargetList(parsedCmd, templateIdList=parsedCmd.templateId.idList,
+                                                 **kwargs)
+
+
+class ImageDifferenceTask(pipeBase.CmdLineTask):
+    """Subtract an image from a template and measure the result
+    """
+    ConfigClass = ImageDifferenceConfig
+    RunnerClass = ImageDifferenceTaskRunner
+    _DefaultName = "imageDifference"
+
+    def __init__(self, butler=None, **kwargs):
+        """!Construct an ImageDifference Task
+
+        @param[in] butler  Butler object to use in constructing reference object loaders
+        """
+        pipeBase.CmdLineTask.__init__(self, **kwargs)
+        self.makeSubtask("getTemplate")
+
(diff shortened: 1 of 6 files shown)


python/lsst/ip/diffim/registerImage.py
```diff
@@ -0,0 +1,192 @@
...
+        inCentroidKey = copyMatches[0].second.getTable().getCentroidKey()
+        for i in range(self.config.sipIter):
+            sipFit = makeCreateWcsWithSip(copyMatches, inputWcs, self.config.sipOrder, inputBBox)
+            self.log.debug("Registration WCS RMS iteration %d: %f pixels",
+                           i, sipFit.getScatterInPixels())
+            wcs = sipFit.getNewWcs()
+            dr = [m.first.get(refCoordKey).angularSeparation(
+                wcs.pixelToSky(m.second.get(inCentroidKey))).asArcseconds() for
+                m in copyMatches]
+            dr = numpy.array(dr)
```
parejkoj: Make the list comp a generator and put it inside `np.array` to avoid the temporary.

python/lsst/ip/diffim/registerImage.py
```diff
@@ -0,0 +1,192 @@
...
+            rms = math.sqrt((dr*dr).mean())  # RMS from zero
+            rms = max(rms, 1.0e-9)  # Don't believe any RMS smaller than this
+            self.log.debug("Registration iteration %d: rms=%f", i, rms)
+            good = numpy.where(dr < self.config.sipRej*rms)[0]
+            numBad = len(copyMatches) - len(good)
+            self.log.debug("Registration iteration %d: rejected %d", i, numBad)
+            if numBad == 0:
+                break
+            copyMatches = type(matches)(copyMatches[i] for i in good)
```
parejkoj: This should be much faster: replace `np.where` above with `good = dr < self.config.sipRej*rms`, and the creation of the new `copyMatches` with `copyMatches = list(itertools.compress(x,a))` (I believe type(matches) is a list in this instance, but you can check that: if it's more like a np array you can just slice it directly with the bool array).

You also would replace `len(good)` with `good.sum()`, since it's a bool array now.

python/lsst/ip/diffim/registerImage.py
```diff
@@ -0,0 +1,192 @@
...
+        @param templateBBox: Target bounding box
+        @return Warped exposure
+        """
+        warper = Warper.fromConfig(self.config.warper)
+        copyExp = inputExp.Factory(inputExp.getMaskedImage(), newWcs)
+        alignedExp = warper.warpExposure(templateWcs, copyExp, destBBox=templateBBox)
+        return alignedExp
+
+    def warpSources(self, inputSources, newWcs, templateWcs, templateBBox):
+        """Warp sources to the new frame
```
parejkoj: You should note that only points that fall inside templateBBox are included in the output sources. It took a bit of staring at the code to realize that.

tests/test_Register.py
```diff
@@ -0,0 +1,231 @@
+from __future__ import division, print_function, absolute_import
+from builtins import zip
+from builtins import range
+#!/usr/bin/env python
+#
+# LSST Data Management System
+# Copyright 2008-2013 LSST Corporation.
```
parejkoj: 2017

tests/test_Register.py
```diff
@@ -0,0 +1,231 @@
...
+import numpy as np
+
+import lsst.utils.tests
+import lsst.afw.image as afwImage
+import lsst.afw.table as afwTable
+import lsst.afw.coord as afwCoord
+import lsst.afw.geom as afwGeom
+import lsst.afw.display.ds9 as ds9
+from lsst.pipe.base import Struct
+#from lsst.pipe.tasks.registerImage import RegisterConfig, RegisterTask
```
parejkoj: Line not needed?

tests/test_Register.py
```diff
@@ -0,0 +1,231 @@
...
+from lsst.ip.diffim.registerImage import RegisterConfig, RegisterTask
+
+try:
+    display
+except NameError:
+    display = False
+
+
+class RegisterTestCase(unittest.TestCase):
```
parejkoj: Derive from `lsst.utils.tests.TestCase`.

tests/test_Register.py
```diff
@@ -0,0 +1,231 @@
...
+
+    """A test case for RegisterTask."""
```
parejkoj: No need for redundant docstrings.

tests/test_Register.py
```diff
@@ -0,0 +1,231 @@
...
+
+    def setUp(self):
+        self.dx = -5
+        self.dy = +3
+        self.numSources = 123
+        self.border = 10  # Must be larger than dx,dy
```
parejkoj: Useful comment.

tests/test_Register.py
```diff
@@ -0,0 +1,231 @@
...
+                # The density of points is such that I can assume that no point is next to another.
+                # The values are not quite zero because the "image" is undersampled, so we get ringing.
+                self.assertTrue((alignedArray[yTemplate+dy, xTemplate+dx] < 0.1).all())
+
+        xAligned = np.array([x for i, x in enumerate(outData.warpedSources["center_x"]) if i not in bad])
+        yAligned = np.array([y for i, y in enumerate(outData.warpedSources["center_y"]) if i not in bad])
+        self.assertAlmostEqual((xAligned - xTemplate).mean(), 0, 8)
+        self.assertAlmostEqual((xAligned - xTemplate).std(), 0, 8)
+        self.assertAlmostEqual((yAligned - yTemplate).mean(), 0, 8)
+        self.assertAlmostEqual((yAligned - yTemplate).std(), 0, 8)
```
parejkoj: Instead of testing mean and std here, if you derive from `utils.tests.TestCase`, you can use `assertImagesNearlyEqual` to test all the points. And I think you can simplify the `assertTrue(stuff.all())` tests above with `assertFloatsNearlyEqual` too.

tests/test_Register.py
```diff
@@ -0,0 +1,231 @@
...
+    def testRejection(self):
+        """Test image registration with rejection"""
+        inData = self.create()
+
+        # Tweak a source to have a bad offset
+        badIndex = 111
+
+        coordKey = inData.inputSources[badIndex].getTable().getCoordKey()
+        centroidKey = inData.inputSources[badIndex].getTable().getCentroidKey()
+        x, y = float(inData.xInput[badIndex] + 0.01), float(inData.yInput[badIndex] - 0.01)
```
parejkoj: Should the `0.01` here be some multiple of `self.pixelscale`? Otherwise, it's not clear why `0.01` is the right value.

tests/test_Register.py
```diff
@@ -0,0 +1,231 @@
...
+                                         inData.templateExp.getBBox(afwImage.LOCAL))
+
+        self.showImage(warpedExp, warpedSources, "Aligned", 3)
+        return Struct(warpedExp=warpedExp, warpedSources=warpedSources, matches=results.matches,
+                      wcs=results.wcs, task=task)
+
+    def assertRegistered(self, inData, outData, bad=set()):
+        """Assert that the registration task is registering images"""
+        xTemplate = np.array([x for i, x in enumerate(inData.xTemplate) if i not in bad])
+        yTemplate = np.array([y for i, y in enumerate(inData.yTemplate) if i not in bad])
```
parejkoj: Don't need `[]` list comp here, can just use generator.

python/lsst/ip/diffim/imageDifference.py
```diff
@@ -0,0 +1,327 @@
...
+        diaSrc = afwTable.SourceCatalog(self.schema)
+        diaSrc.getTable().setMetadata(self.algMetadata)
+        return {self.config.coaddName + "Diff_diaSrc": diaSrc}
+
+    @classmethod
+    def _makeArgumentParser(cls):
+        """Create an argument parser
+        """
+        parser = pipeBase.ArgumentParser(name=cls._DefaultName)
+        parser.add_id_argument("--id", "calexp", help="data ID, e.g. --id visit=12345 ccd=1,2")
```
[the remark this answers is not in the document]
mrawls: They are different because the default datasetType is calexp for diffim and raw in pipeBase. (The templateId doesn't exist in pipeBase.)

python/lsst/ip/diffim/makeDiffim.py
```diff
@@ -0,0 +1,874 @@
...
+
+This assumes the following makeDiffimConfig.py file is in your working directory:
+
+\code
+config.doWriteSubtractedExp=True
+config.doWriteMatchedExp=True
+config.doDecorrelation=True
+config.subtract='al'
+
+config.subtract['zogy'].zogyConfig.inImageSpace=False
```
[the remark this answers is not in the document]
mrawls: Resolved, but additional config examples are forthcoming from @djreiss in the diffimconfig.py file he is going to dig up :)

isullivan:
This PR is obsolete.

PR #75, opened 2017-09-22
Make coadd template warp type configurable

yalsayyad:
Use `--config getTemplate.warpType=psfMatched` to get
PSF-matched coadds as templates.

### DIFF
--- a/python/lsst/ip/diffim/getTemplate.py
+++ b/python/lsst/ip/diffim/getTemplate.py
@@ -43,6 +43,11 @@ class GetCoaddAsTemplateConfig(pexConfig.Config):
         dtype=str,
         default="deep",
     )
+    warpType = pexConfig.Field(
+        doc="Warp type of the coadd template: one of 'direct' or 'psfMatched'",
+        dtype=str,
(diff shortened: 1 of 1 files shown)


python/lsst/ip/diffim/getTemplate.py
```diff
@@ -135,6 +140,19 @@ def run(self, exposure, sensorRef, templateIdList=None):
...
+        """Return coadd name for given task config
+
+        Returns
+        -------
+        CoaddDatasetName : `string`
+
+        TODO: This nearly duplicates a method in CoaddBaseTask (DM-11985)
+        """
+        warpType = self.config.warpType
+        suffix = "" if warpType == "direct" else warpType[0].upper() + warpType[1:]
```
isullivan: This is probably the most efficient way of doing this, but using `warpType .title()` would be more clear.

isullivan: Oops. Yes, I didn't realize that would un-capitalize letters in the rest of the word. Good catch.

PR #84, opened 2018-04-25
DM-14204: Modernize python in ip_* packages

r-owen:


tests/test_imageDecorrelation.py
```diff
@@ -18,11 +18,8 @@
 # You should have received a copy of the LSST License Statement and
 # the GNU General Public License along with this program.  If not,
 # see <https://www.lsstcorp.org/LegalNotices/>.
-from __future__ import absolute_import, division, print_function
 import unittest
 
-from builtins import range
-from past.builtins import basestring
```
natelust: As you are doing more than future, you may consider changing the commit message to something about dropping py2 support and syntax, but it is just a suggestion

python/lsst/ip/diffim/imageDecorrelation.py
```diff
@@ -526,51 +525,51 @@ def setDefaults(self):
...
-    See \ref DecorrelateALKernelSpatialConfig
+    See @ref DecorrelateALKernelSpatialConfig
 
-    \section ip_diffim_imageDecorrelation_DecorrelateALKernelSpatialTask_Debug		Debug variables
+    @section ip_diffim_imageDecorrelation_DecorrelateALKernelSpatialTask_Debug		Debug variables
 
     This task has no debug variables
 
-    \section ip_diffim_imageDecorrelation_DecorrelateALKernelSpatialTask_Example	Example of using DecorrelateALKernelSpatialTask
+    @section ip_diffim_imageDecorrelation_DecorrelateALKernelSpatialTask_Example  Example of using DecorrelateALKernelSpatialTask  # noqa E501
```
natelust: Should there be two spaces before Example?

r-owen: I don't think it makes any difference. Unfortunately the `# noqa 501` is going to show up as part of the title so I have to fix that (in two places). Doxygen and pep8 don't get along very well. While I'm fixing that I'll make it one space in the files I fix.

PR #86, opened 2018-05-10
DM-10797 Update to use new SourceSelector API

parejkoj:


python/lsst/ip/diffim/diaCatalogSourceSelector.py
```diff
@@ -166,16 +166,27 @@ def DebugInfo(name):
...
+    def selectSources(self, sourceCat, matches=None, exposure=None):
+        """Return a selection of sources for Kernel candidates.
+
+        Parameters:
+        -----------
+        sourceCat : `lsst.afw.table.SourceCatalog`
+            Catalog of sources to select from.
+            This catalog must be contiguous in memory.
+        matches : `list` of `lsst.afw.table.ReferenceMatch` or None
+             A match vector as produced by meas_astrom.
```
TallJimbo: `None` is not actually permitted in this case, right?

PR #88, opened 2018-06-07
DM-14725: Eliminate explicit use of ndarray::EigenView in C++ code

r-owen:


src/KernelSolution.cc
```diff
@@ -559,22 +559,22 @@ namespace diffim {
         ndarray::Array<int, 1, 1> maskArray =
             ndarray::allocate(ndarray::makeVector(fullFp->getArea()));
         fullFp->getSpans()->flatten(maskArray, finalMask.getArray(), templateImage.getXY0());
-        ndarray::EigenView<int, 1, 1> maskEigen(maskArray);
+        auto maskEigen = maskArray.asEigen();
```
kfindeisen: Can't say I approve of this use of `auto`, but I gather it's for forward-compatibility with DM-14728.

r-owen: Yes

PR #93, opened 2018-07-25
DM-14738: Support DCR-matched templates for image differencing

isullivan:


python/lsst/ip/diffim/getTemplate.py
```diff
@@ -107,19 +113,32 @@ def run(self, exposure, sensorRef, templateIdList=None):
...
             )
             if patchSubBBox.isEmpty():
                 self.log.info("skip tract=%(tract)s, patch=%(patch)s; no overlapping pixels" % patchArgDict)
                 continue
-            if not sensorRef.datasetExists(**patchArgDict):
-                self.log.warn("%(datasetType)s, tract=%(tract)s, patch=%(patch)s does not exist"
-                              % patchArgDict)
-                continue
 
             nPatchesFound += 1
```
yalsayyad: nPatchesFound was incremented if the patch was found by the butler. This way if there was no data to make a template, it would raise.  I think you meant to put your block that reads or builds a patch above nPatchedFound += 0.

python/lsst/ip/diffim/getTemplate.py
```diff
@@ -107,19 +113,32 @@ def run(self, exposure, sensorRef, templateIdList=None):
...
-            self.log.info("Reading patch %s" % patchArgDict)
-            coaddPatch = sensorRef.get(**patchArgDict)
-            coaddExposure.getMaskedImage().assign(coaddPatch.getMaskedImage(), coaddPatch.getBBox())
+            if self.config.coaddName == 'dcr':
```
yalsayyad: I not crazy about the "if dcr do these three things else do these other three things."   It's fine for now, but will you help me refactor it per RFC-352 soon?

python/lsst/ip/diffim/dcrModel.py
```diff
@@ -0,0 +1,619 @@
...
+            Best fit model of the true sky after correcting chromatic effects.
+        """
+        # NANs will potentially contaminate the entire image,
+        # depending on the shift or convolution type used.
+        model = maskedImage.clone()
+        badPixels = np.isnan(model.image.array) | np.isnan(model.variance.array)
+        model.image.array[badPixels] = 0.
+        model.variance.array[badPixels] = 0.
+        model.image.array /= dcrNumSubfilters
+        model.variance.array /= dcrNumSubfilters
```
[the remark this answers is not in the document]
isullivan: I believe so, yes. Since the subfilter images are treated as independent, when we divide an image into N subfilters I think each should be treated as having less information, and sqrt(N) worse signal to noise. I will add unit tests in DM-15342 and DM-15343 that will test this and hopefull clarify the expected behavior.

python/lsst/ip/diffim/getTemplate.py
isullivan: I quite agree that the "if dcr..." block is ugly, and I would strongly welcome any suggestions for how to change it. One possibility would be to modify the butler so that `coaddPatch = sensorRef.get(**patchArgDict)` would know to calculate and return the same thing as the DcrModel calculations, but I worry that would look surprising and magic.

isullivan: I am also happy to help implement RFC-352 if that is useful, though `DcrModel` here is not a `Task`, so I'm not sure what needs to be done.

python/lsst/ip/diffim/getTemplate.py
```diff
@@ -107,19 +113,32 @@ def run(self, exposure, sensorRef, templateIdList=None):
...
+                                                           wcs=coaddWcs,
+                                                           visitInfo=exposure.getInfo().getVisitInfo())
+            else:
+                if not sensorRef.datasetExists(**patchArgDict):
+                    self.log.warn("%(datasetType)s, tract=%(tract)s, patch=%(patch)s does not exist"
+                                  % patchArgDict)
+                    continue
+                self.log.info("Reading patch %s" % patchArgDict)
+                coaddPatch = sensorRef.get(**patchArgDict)
+            coaddExposure.maskedImage.assign(coaddPatch.maskedImage, coaddPatch.getBBox())
```
RobertLuptonTheGood: Is this equivalent to `coaddExposure.maskedImage[coaddPatch.getBBox()] = coaddPatch.maskedImage`? If so, that'd be clearer

python/lsst/ip/diffim/getTemplate.py
yalsayyad: I meant that I plan to refactor `GetCoaddAsTemplateTask`.  I might need to create a DcrModel with your `fromImage` interface rather than your `fromDataRef` interface, and might need advice.

python/lsst/ip/diffim/getTemplate.py
isullivan: I believe it is equivalent in this instance, but I'd like to leave this as it is for consistency with `assembleCoadd`. If you'd like that changed, I (or someone else) could do that on a separate ticket for both `ip_diffim` and `pipe_tasks`, and possibly other packages as well.

PR #98, opened 2018-09-18
Tickets/dm-15636: Improve DcrModel regularization

isullivan:
These changes provide greater flexibility in applying regularization to DcrModels between iterations and between frequencies, and reduce artifacts in the final images.

### DIFF
--- a/python/lsst/ip/diffim/dcrModel.py
+++ b/python/lsst/ip/diffim/dcrModel.py
@@ -377,34 +377,32 @@ class DcrModel:
         templateExposure.setFilter(self.filter)
         return templateExposure
 
-    def conditionDcrModel(self, subfilter, newModel, bbox, gain=1.):
+    def conditionDcrModel(self, modelImages, bbox, gain=1.):
         """Average two iterations' solutions to reduce oscillations.
 
         Parameters
         ----------
-        subfilter : `int`
-            Index of the current subfilter within the full band.
-        newModel : `lsst.afw.image.MaskedImage`
-            The new DCR model for one subfilter from the current iteration.
-            Values in ``newModel`` that are extreme compared with the last
-            iteration are modified in place.
+        modelImages : `list` of `lsst.afw.image.MaskedImage`
+            The new DCR model images from the current iteration.
+            The values will be modified in place.
         bbox : `lsst.afw.geom.Box2I`
             Sub-region of the coadd
         gain : `float`, optional
-            Additional weight to apply to the model from the current iteration.
+            Relative weight to give the new solution when updating the model.
             Defaults to 1.0, which gives equal weight to both solutions.
         """
         # Calculate weighted averages of the image and variance planes.
         # Note that ``newModel *= gain`` would multiply the variance by ``gain**2``
-        newModel.image *= gain
-        newModel.image += self[subfilter][bbox].image
-        newModel.image /= 1. + gain
-        newModel.variance *= gain
-        newModel.variance += self[subfilter][bbox].variance
-        newModel.variance /= 1. + gain
-
     def clampModel(self, subfilter, newModel, bbox, statsCtrl, regularizeSigma, modelClampFactor,
                    convergenceMaskPlanes="DETECTED"):
+        for model, newModel in zip(self, modelImages):
+            newModel.image *= gain
+            newModel.image += model[bbox].image
+            newModel.image /= 1. + gain
+            newModel.variance *= gain
+            newModel.variance += model[bbox].variance
+            newModel.variance /= 1. + gain
+
         """Restrict large variations in the model between iterations.
 
         Parameters
--- a/python/lsst/ip/diffim/dcrModel.py
+++ b/python/lsst/ip/diffim/dcrModel.py
@@ -461,7 +461,7 @@ class DcrModel:
         maskedImage : `lsst.afw.image.MaskedImage`
             The input image to evaluate the background noise properties.
         statsCtrl : `lsst.afw.math.StatisticsControl`
-            Statistics control object for coadd
+            Statistics control object for coaddition.
         bufferSize : `int`
             Number of additional pixels to exclude
             from the edges of the bounding box.
@@ -519,13 +519,13 @@ class DcrModel:
         image = maskedImage.image.array
         filterStructure = ndimage.iterate_structure(ndimage.generate_binary_structure(2, 1),
                                                     regularizationWidth)
-        if highThreshold is not None:
+        if highThreshold:
             highPixels = image > highThreshold
             if regularizationWidth > 0:
                 # Erode and dilate ``highPixels`` to exclude noisy pixels.
                 highPixels = ndimage.morphology.binary_opening(highPixels, structure=filterStructure)
             image[highPixels] = highThreshold[highPixels]
-        if lowThreshold is not None:
+        if lowThreshold:
             lowPixels = image < lowThreshold
             if regularizationWidth > 0:
                 # Erode and dilate ``lowPixels`` to exclude noisy pixels.
--- a/python/lsst/ip/diffim/dcrModel.py
+++ b/python/lsst/ip/diffim/dcrModel.py
@@ -21,6 +21,7 @@
 #
 
 import numpy as np
+from scipy import ndimage
 from lsst.afw.coord.refraction import differentialRefraction
 import lsst.afw.geom as afwGeom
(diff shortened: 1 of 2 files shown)


python/lsst/ip/diffim/dcrModel.py
```diff
@@ -406,34 +418,36 @@ def clampModel(self, subfilter, newModel, bbox, statsCtrl, regularizeSigma, mode
             Sub-region to coadd
         statsCtrl : `lsst.afw.math.StatisticsControl`
             Statistics control object for coadd
-        regularizeSigma : `float`
-            Threshold to exclude noise-like pixels from regularization.
-        modelClampFactor : `float`
+        clamp : `float`
```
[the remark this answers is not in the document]
isullivan: I've used `clamp` in this context for a long time, but it's not defined that way in a paper. I'm happy to change it to make the code more readable. Would `maxDelta` be an improvement?
Also, should I change the name of the method to something like `regularizeIterations` or `restrictModelDelta`?

python/lsst/ip/diffim/dcrModel.py
```diff
@@ -406,34 +418,36 @@ def clampModel(self, subfilter, newModel, bbox, statsCtrl, regularizeSigma, mode
...
+        bufferSize : `int`
+            Number of additional pixels to exclude
+            from the edges of the bounding box.
         convergenceMaskPlanes : `list` of `str`, or `str`, optional
             Mask planes to use to calculate convergence.
-            Default value is set in ``calculateNoiseCutoff`` if not supplied.
+        regularizationWidth : int, optional
+            Minimum radius of a region to include in regularization, in pixels.
         """
+        denoiseStructure = ndimage.iterate_structure(ndimage.generate_binary_structure(2, 1),
```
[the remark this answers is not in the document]
isullivan: I'll move this code to a separate function, and try to make the usage clear in the comments there. 
This does smooth out the edges, but only of region of pixels being selected. The smoothing actually improves the overall results, since the artifacts should not have ragged edges.

python/lsst/ip/diffim/dcrModel.py
```diff
@@ -442,39 +456,61 @@ def regularizeModel(self, bbox, mask, statsCtrl, regularizeSigma, clampFrequency
 
         Parameters
         ----------
+        modelImages : `list` of `lsst.afw.image.MaskedImage`
```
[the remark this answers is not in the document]
isullivan: Yes, I think the new multiband images are a good fit for the dcr images. I've filed a ticket to look into switching over to them.

python/lsst/ip/diffim/dcrModel.py
```diff
@@ -442,39 +456,61 @@ def regularizeModel(self, bbox, mask, statsCtrl, regularizeSigma, clampFrequency
...
+            highPixels = ndimage.morphology.binary_opening(highPixels, structure=denoiseStructure)
+            excess[highPixels] += modelVals[highPixels] - refImage[highPixels]*clampToRef
+            modelVals[highPixels] -= excess[highPixels]
+            lowPixels = modelVals < lowThreshold
+            lowPixels = ndimage.morphology.binary_opening(lowPixels, structure=denoiseStructure)
+            excess[lowPixels] += modelVals[lowPixels] - refImage[lowPixels]/clampToRef
+            modelVals[lowPixels] -= excess[lowPixels]
+            excesses.append(excess)
+        # We want to preserve flux while reducing extreme variations between subfilters.
+        # So, add the mean clipped flux back to all subfilters equally.
```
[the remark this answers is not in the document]
isullivan: After quite a bit of thought and testing, I have removed these lines of code. The concern it addressed has become obsolete, and it doesn't fit easily with the reorganization of the above filtering code into a separate method.

python/lsst/ip/diffim/dcrModel.py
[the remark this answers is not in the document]
natelust: Sorry I somehow missed you writing this comment, I think either of your suggestions are fine, or something like dampenVariations (though I guess that implies something other than hard clipping, so maybe thresholdVariations). If you go with thresholdVariations maybe the parameter could be called deltaThreshold. I am fine with your suggestions, or if you want to take one of mine, or if you come up with something even better.

python/lsst/ip/diffim/dcrModel.py
```diff
@@ -545,9 +486,52 @@ def calculateNoiseCutoff(self, maskedImage, statsCtrl, sigma, bufferSize,
         convergeMask = mask.getPlaneBitMask(convergenceMaskPlanes)
 
         backgroundPixels = mask[bboxShrink].array & (statsCtrl.getAndMask() | convergeMask) == 0
-        noiseCutoff = sigma*np.std(maskedImage[bboxShrink].image.array[backgroundPixels])
+        noiseCutoff = np.std(maskedImage[bboxShrink].image.array[backgroundPixels])
```
natelust: This seems like it would me more natural (and probably faster) if you just used afwMath.makeStatistics setting the appropriate bit you want andMask | convergeMask in your own control object. This would save a few loops over the array.

python/lsst/ip/diffim/dcrModel.py
```diff
@@ -545,9 +486,52 @@ def calculateNoiseCutoff(self, maskedImage, statsCtrl, sigma, bufferSize,
...
+        regularizationWidth : `int`, optional
+            Minimum radius of a region to include in regularization, in pixels.
+        """
+        # Generate the structure for binary erosion and dilation, which is used to remove noise-like pixels.
+        # Groups of pixels with a radius smaller than ``regularizationWidth``
+        # will be excluded from regularization.
+        image = maskedImage.image.array
+        filterStructure = ndimage.iterate_structure(ndimage.generate_binary_structure(2, 1),
+                                                    regularizationWidth)
+        if highThreshold is not None:
```
natelust: This can be just `if highThreshold:` as None is always False-y, and you are only checking that it is not None anyway, not that it is something specific.

python/lsst/ip/diffim/dcrModel.py
```diff
@@ -545,9 +486,52 @@ def calculateNoiseCutoff(self, maskedImage, statsCtrl, sigma, bufferSize,
...
+            highPixels = image > highThreshold
+            if regularizationWidth > 0:
+                # Erode and dilate ``highPixels`` to exclude noisy pixels.
+                highPixels = ndimage.morphology.binary_opening(highPixels, structure=filterStructure)
+            image[highPixels] = highThreshold[highPixels]
+        if lowThreshold is not None:
```
natelust: same as above

python/lsst/ip/diffim/dcrModel.py
```diff
@@ -545,9 +486,52 @@ def calculateNoiseCutoff(self, maskedImage, statsCtrl, sigma, bufferSize,
         convergeMask = mask.getPlaneBitMask(convergenceMaskPlanes)
 
         backgroundPixels = mask[bboxShrink].array & (statsCtrl.getAndMask() | convergeMask) == 0
-        noiseCutoff = sigma*np.std(maskedImage[bboxShrink].image.array[backgroundPixels])
+        noiseCutoff = np.std(maskedImage[bboxShrink].image.array[backgroundPixels])
         return noiseCutoff
 
+    def applyImageThresholds(self, maskedImage, highThreshold=None, lowThreshold=None,
```
natelust: This makes things much clearer, thank you for making this change.

python/lsst/ip/diffim/dcrModel.py
isullivan: I agree it would be cleaner with all the bits set in the statsCtrl object, but I'm going to punt on this change for now. I started writing it, but then realized that I would have to pass around several additional statsCtrl objects, because sometimes I want to exclude pixels with either a set of bad masks or "convergence" masks set, and sometimes I want to exclude pixels with the set of bad masks but NOT the convergence masks set.

python/lsst/ip/diffim/dcrModel.py
isullivan: Actually, I had to revert this change. Since `highThreshold` is an array, `if highThreshold` throws an error.

python/lsst/ip/diffim/dcrModel.py
isullivan: Also changed and then reverted, as above.

PR #102, opened 2018-10-19
DM-15635: Approximate filter throughput for DcrCoadds

isullivan:


python/lsst/ip/diffim/dcrModel.py
```diff
@@ -549,23 +552,37 @@ def applyDcr(maskedImage, dcr, warpCtrl, bbox=None, useInverse=False):
         Shifts the entire image if None (Default).
     useInverse : `bool`, optional
         Use the reverse of ``dcr`` for the shift. Default: False
+    useMidpoint : `bool`, optional
+        Use the midpoint of each subfilter to calculate DCR. Default: True
```
mrawls: Can you clarify in the doctoring what the behavior is if this is False? I think it uses the start and endpoint of each filter instead, but saying that here would be helpful.

python/lsst/ip/diffim/dcrModel.py
```diff
@@ -578,14 +595,17 @@ def calculateDcr(visitInfo, wcs, filterInfo, dcrNumSubfilters):
         The filter definition, set in the current instruments' obs package.
     dcrNumSubfilters : `int`
         Number of sub-filters used to model chromatic effects within a band.
+    useMidpoint : `bool`, optional
+        Use the midpoint of each subfilter to calculate DCR. Default: True
```
mrawls: Again, please briefly say what happens if useMidpoint is set to False.

python/lsst/ip/diffim/dcrModel.py
isullivan: I'm going to actually rename `useMidpoint` to `splitSubfilters`, with a default of False. That will make the documentation more clear.

PR #110, opened 2019-02-07
DM-17564: Convert DcrModel to image type from maskedImage

isullivan:
This simpler data model reduces execution time by a factor of ~3.

### DIFF
--- a/python/lsst/ip/diffim/dcrModel.py
+++ b/python/lsst/ip/diffim/dcrModel.py
@@ -24,10 +24,7 @@ import numpy as np
 from scipy import ndimage
 from lsst.afw.coord.refraction import differentialRefraction
 import lsst.afw.geom as afwGeom
-from lsst.afw.geom import AffineTransform
-from lsst.afw.geom import makeTransform
 import lsst.afw.image as afwImage
-import lsst.afw.math as afwMath
 from lsst.geom import radians
 
 __all__ = ["DcrModel", "applyDcr", "calculateDcr", "calculateImageParallacticAngle"]
@@ -40,7 +37,7 @@ class DcrModel:
     ----------
     dcrNumSubfilters : `int`
         Number of sub-filters used to model chromatic effects within a band.
-    modelImages : `list` of `lsst.afw.image.MaskedImage`
+    modelImages : `list` of `lsst.afw.image.Image`
         A list of masked images, each containing the model for one subfilter
 
     Notes
@@ -53,11 +50,13 @@ class DcrModel:
     iterations of forward modeling or between the subfilters of the model.
     """
 
-    def __init__(self, modelImages, filterInfo=None, psf=None):
+    def __init__(self, modelImages, filterInfo=None, psf=None, mask=None, variance=None):
         self.dcrNumSubfilters = len(modelImages)
         self.modelImages = modelImages
         self._filter = filterInfo
         self._psf = psf
+        self._mask = mask
+        self._variance = variance
 
     @classmethod
     def fromImage(cls, maskedImage, dcrNumSubfilters, filterInfo=None, psf=None):
@@ -80,25 +79,28 @@ class DcrModel:
         -------
         dcrModel : `lsst.pipe.tasks.DcrModel`
             Best fit model of the true sky after correcting chromatic effects.
+
+        Raises
+        ------
+        ValueError
+            If there are any unmasked NAN values in ``maskedImage``.
         """
         # NANs will potentially contaminate the entire image,
         # depending on the shift or convolution type used.
-        model = maskedImage.clone()
-        badPixels = np.isnan(model.image.array) | np.isnan(model.variance.array)
-        model.image.array[badPixels] = 0.
-        model.variance.array[badPixels] = 0.
-        model.image.array /= dcrNumSubfilters
+        model = maskedImage.image.clone()
+        mask = maskedImage.mask.clone()
         # We divide the variance by N and not N**2 because we will assume each
         # subfilter is independent. That means that the significance of
         # detected sources will be lower by a factor of sqrt(N) in the
         # subfilter images, but we will recover it when we combine the
         # subfilter images to construct matched templates.
-        model.variance.array /= dcrNumSubfilters
-        model.mask.array[badPixels] = model.mask.getPlaneBitMask("NO_DATA")
+        variance = maskedImage.variance.clone()
+        variance /= dcrNumSubfilters
+        model /= dcrNumSubfilters
         modelImages = [model, ]
         for subfilter in range(1, dcrNumSubfilters):
             modelImages.append(model.clone())
-        return cls(modelImages, filterInfo, psf)
+        return cls(modelImages, filterInfo, psf, mask, variance)
 
     @classmethod
     def fromDataRef(cls, dataRef, datasetType="dcrCoadd", numSubfilters=None, **kwargs):
@@ -126,6 +128,8 @@ class DcrModel:
         modelImages = []
         filterInfo = None
         psf = None
+        mask = None
+        variance = None
         for subfilter in range(numSubfilters):
             dcrCoadd = dataRef.get(datasetType, subfilter=subfilter,
                                    numSubfilters=numSubfilters, **kwargs)
@@ -133,8 +137,12 @@ class DcrModel:
                 filterInfo = dcrCoadd.getFilter()
             if psf is None:
                 psf = dcrCoadd.getPsf()
-            modelImages.append(dcrCoadd.maskedImage)
-        return cls(modelImages, filterInfo, psf)
+            if mask is None:
+                mask = dcrCoadd.mask
+            if variance is None:
+                variance = dcrCoadd.variance
+            modelImages.append(dcrCoadd.image)
+        return cls(modelImages, filterInfo, psf, mask, variance)
 
     def __len__(self):
         """Return the number of subfilters.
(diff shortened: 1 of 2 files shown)


python/lsst/ip/diffim/dcrModel.py
```diff
@@ -83,22 +82,23 @@ def fromImage(cls, maskedImage, dcrNumSubfilters, filterInfo=None, psf=None):
...
         # We divide the variance by N and not N**2 because we will assume each
         # subfilter is independent. That means that the significance of
         # detected sources will be lower by a factor of sqrt(N) in the
         # subfilter images, but we will recover it when we combine the
         # subfilter images to construct matched templates.
-        model.variance.array /= dcrNumSubfilters
-        model.mask.array[badPixels] = model.mask.getPlaneBitMask("NO_DATA")
+        variance = maskedImage.variance.clone()
+        variance /= dcrNumSubfilters
+        badPixels = np.isnan(model.array) | np.isnan(variance.array)
```
parejkoj: Shouldn't these pixels be flagged by the pipeline?

python/lsst/ip/diffim/dcrModel.py
```diff
@@ -126,15 +126,21 @@ def fromDataRef(cls, dataRef, datasetType="dcrCoadd", numSubfilters=None, **kwar
...
             if filterInfo is None:
                 filterInfo = dcrCoadd.getFilter()
             if psf is None:
                 psf = dcrCoadd.getPsf()
-            modelImages.append(dcrCoadd.maskedImage)
-        return cls(modelImages, filterInfo, psf)
+            if mask is None:
+                mask = dcrCoadd.mask
+            if variance is None:
+                variance = dcrCoadd.mask
```
parejkoj: I'm confused by these lines: the filter, psf, mask and variance always get set to the first subfilter?

python/lsst/ip/diffim/dcrModel.py
```diff
@@ -558,50 +562,46 @@ def applyImageThresholds(self, image, highThreshold=None, lowThreshold=None, reg
             image[lowPixels] = lowThreshold[lowPixels]
 
 
-def applyDcr(maskedImage, dcr, warpCtrl, bbox=None, useInverse=False, splitSubfilters=False):
+def applyDcr(image, dcr, useInverse=False, splitSubfilters=False, **kwargs):
     """Shift a masked image.
```
parejkoj: It's no longer a masked image, and I think this docstring could be expanded: what direction is the shift, by how much, why is `dcr` a tuple?

python/lsst/ip/diffim/dcrModel.py
```diff
@@ -558,50 +562,46 @@ def applyImageThresholds(self, image, highThreshold=None, lowThreshold=None, reg
...
+    dcr : `tuple` of two `float`
         Shift calculated with ``calculateDcr``.
-    warpCtrl : `lsst.afw.math.WarpingControl`
-        Configuration settings for warping an image
-    bbox : `lsst.afw.geom.Box2I`, optional
-        Sub-region of the masked image to shift.
-        Shifts the entire image if None (Default).
+        Uses numpy axes ordering (Y, X).
     useInverse : `bool`, optional
         Use the reverse of ``dcr`` for the shift. Default: False
```
parejkoj: Is this really an "inverse", or a "reverse" or some other name? It's shifting by `-dcr`, no?

python/lsst/ip/diffim/dcrModel.py
```diff
@@ -558,50 +562,46 @@ def applyImageThresholds(self, image, highThreshold=None, lowThreshold=None, reg
...
-        afwMath.warpImage(shiftedImage, maskedImage[bbox],
-                          transform0, warpCtrl, padValue=padValue)
-        shiftedImage1 = afwImage.MaskedImageF(bbox)
-        transform1 = makeTransform(AffineTransform((-1.0 if useInverse else 1.0)*dcr[1]))
-        afwMath.warpImage(shiftedImage1, maskedImage[bbox],
-                          transform1, warpCtrl, padValue=padValue)
-        shiftedImage += shiftedImage1
+        if useInverse:
+            shift = [-1.*s for s in dcr[0]]
+            shift1 = [-1.*s for s in dcr[1]]
```
parejkoj: The docstring says `dcr` is a `tuple of two float`, so how can this work?

tests/test_dcrModel.py
```diff
@@ -359,39 +346,36 @@ def testRotationFlipped(self):
...
-        This additionally tests that the variance and mask planes do not change.
         """
-        dcrModels = DcrModel(modelImages=self.makeTestImages())
+        modelImages = self.makeTestImages()
+        dcrModels = DcrModel(modelImages=modelImages, mask=self.mask)
         newModels = [model.clone() for model in dcrModels]
         dcrModels.conditionDcrModel(newModels, self.bbox, gain=1.)
         for refModel, newModel in zip(dcrModels, newModels):
-            self.assertMaskedImagesEqual(refModel, newModel)
+            self.assertFloatsAlmostEqual(refModel.array, newModel.array, rtol=1e-4, atol=1e-4)
```
parejkoj: That's a surprisingly low tolerance: why so loose?

tests/test_dcrModel.py
```diff
@@ -359,39 +346,36 @@ def testRotationFlipped(self):
...
         newModels = [model.clone() for model in dcrModels]
         for model in newModels:
-            model.image.array[:] *= 3.
+            model.array[:] *= 3.
         dcrModels.conditionDcrModel(newModels, self.bbox, gain=1.)
         for refModel, newModel in zip(dcrModels, newModels):
-            refModel.image.array[:] *= 2.
-            self.assertMaskedImagesAlmostEqual(refModel, newModel)
+            refModel.array[:] *= 2.
+            self.assertFloatsAlmostEqual(refModel.array, newModel.array, rtol=1e-4, atol=1e-4)
```
parejkoj: Another very loose tolerance here.

python/lsst/ip/diffim/dcrModel.py
isullivan: Yes, because those must all be the same for every subfilter. But, I notice there's a pretty bad typo here, with the variance being set to equal the mask. I'll fix that.

python/lsst/ip/diffim/dcrModel.py
isullivan: Since it's a linear shift, a shift by `dcr` in one direction followed by a shift by `-dcr` along the same direction should recover the original image. So, I think it is appropriate to call it an inverse.

python/lsst/ip/diffim/dcrModel.py
isullivan: I'll fix the docstring. `dcr` has a different form if `splitSubfilters` is set.

python/lsst/ip/diffim/dcrModel.py
parejkoj: The docstring says "reverse"?

isullivan: OK, I'll edit the docstring to make that consistent.

tests/test_dcrModel.py
isullivan: I put the loose tolerances in while tracking down a bug, but they're not needed now. I'll remove them.

python/lsst/ip/diffim/dcrModel.py
isullivan: They should be, but I am paranoid since everything will fail if there is a single NAN pixel that slips through. I could add a warning for if that ever happens.

RobertLuptonTheGood:
There should be no NaNs;  you assume that there are none, so I think that an assert would be appropriate

python/lsst/ip/diffim/dcrModel.py
isullivan: Based on Robert's comment, I'll replace this with raising a ValueError should there ever somehow be a NAN value in the maskedImage.

parejkoj: In the unmasked part of the image, no?

isullivan: Yes, of course, and that reminds me why I originally just set any NAN pixel to zero. For simplicity, I perform some operations on the entire array, and that throws `RuntimeWarning`s if there are any NAN values. Any objection if I simply set "NO_DATA" pixels to zero, then raise a `ValueError` if there are any remaining NANs?

parejkoj: Where does the RuntimeWarning come from? We should be able to process data with NaNs in pixels that we don't actually care about (i.e. masked).

How does it change the DCR results if a masked pixel is set to `0` or some other random value?

RobertLuptonTheGood: We interpolate over all NaNs, as otherwise e.g. smoothing kernels spread out the damage.  The stack consistently assumes that bad things like NaNs and dropouts are interpolated with a bit set.

isullivan: Actually I was wrong about that warning. The RuntimeWarning was caused by `dcrWeights` being very small, which was fixed in the commit earlier this afternoon. It looks as though everything would actually work without the NAN test, since it does get masked later. So, I could remove these lines if you would prefer.

parejkoj: Yes, lets try removing this. If we do end up with NaN-related failures later on, we can try to do a better fix.

PR #113, opened 2019-02-20
DM-17522  set of notes on the ip_diffim codebase

kgabor:


doc/lsst.ip.diffim/code_notes.rst
```diff
@@ -0,0 +1,183 @@
...
+specified for the subtasks directly.
+
+Specifying a coadd as template
+------------------------------
+
+Using the ``GetCoaddAsTemplate`` subtask, we specify one or more
+science exposures by the ``--id`` general data specification
+option. Tract or patch specification as part of the ``--id`` option
+are silently ignored. The given (``coaddName``, defaults to "deep")
+type coadd is looked up in the same repository automatically. The
```
mrawls: Please clarify what you mean by "the same repository." The input repository or the output/rerun repository?

doc/lsst.ip.diffim/code_notes.rst
```diff
@@ -0,0 +1,183 @@
...
+Enabling processing steps
+-------------------------
+
+There is a sequence of pre-subtraction and post-subtraction processing
+steps included in ``ImageDifferenceTask`` around the actual
+subtraction operation of the images. Including the subtraction
+operation itself, these steps can be enabled or disabled by top level
+``do<ACTION>`` configuration options. These top-level configuration
+options are summarised in the `subtasks flowchart
+<https://github.com/lsst-dm/diffimTests/tree/master/figure_subtasks>`_
```
mrawls: Will this flowchart image appear in-line in the docs when rendered? (I hope so!) It is a very nice visual. Can you clarify - perhaps in a caption - what comes immediately before and after this flowchart? I think I understand that you put a calexp + template in at the top and get a diffim out at the bottom, but saying this explicitly would be great.

doc/lsst.ip.diffim/code_notes.rst
```diff
@@ -0,0 +1,183 @@
...
+(``templateBorderSize``). The subtask selects all patches of the coadd
+skymap that overlap with the science exposure and cuts them
+accordingly to assemble the template coadd image. The template coadd
+image inherits the coadd WCS. It is an error if there is no
+overlapping region. The ``--templateId`` option is silently
+ignored. The following example specifies one science exposure and
+implicitly the deep coadd residing in the same repository for image
+differencing:
+
+.. code-block:: python
```
mrawls: I think these code blocks should be bash rather than python, since they are things you type on the command line rather than into a python interpreter.

doc/lsst.ip.diffim/code_notes.rst
```diff
@@ -0,0 +1,183 @@
...
+produce a color corrected template matching the ``filter`` of the
+science exposure. 
+	
+Specifying a calexp as template
+-------------------------------
+
+Using the ``GetCalexpAsTemplate`` subtask, we can select two calexp
+data products for subtraction.  The ``--id`` option specifies a
+science exposure.  The ``--templateId`` specifies a new *visit id*
+only. All additional field values are silently ignored; instead, the
```
mrawls: This is consistent with my experience with imageDifferenceTask, but this appears to be a decam-specific example. Other cameras don't necessarily have a "visit id" as you describe here; what would the user need to specify for a templateId in that situation?

doc/lsst.ip.diffim/code_notes.rst
```diff
@@ -0,0 +1,183 @@
...
+then the template is subtracted from the science image.
+
+The performance of the AL algorithm was studied in details in the
+[Becker_LDM-227]_ report. This study forms the basis of the AL
+algorithm default values; the the degree of the polynomial
+multiplicator of the Gaussian kernel basis functions (``degGauss``),
+the degree of the polynomial that is fitted to the spatial variaton of
+the solution coefficients accross the image (``spatialKernelOrder``)
+and the default detection thresholds (5.5 sigma). As a legacy of this
+study, the ``Winter2013ImageDifferenceTask`` is still available in
```
mrawls: Yikes... 😨

doc/lsst.ip.diffim/code_notes.rst
```diff
@@ -0,0 +1,183 @@
...
+assumed to be background dominated i.e. uncorrelated between pixels
+and independent of the pixel values. Also we assume that the noise has
+zero expectation value i.e. the expectation value of the random noise
+is already removed. ZOGY shows that if these assumptions hold, the
+difference image noise is also independent and identically distributed
+over its pixels (white) by construction, there is no need to
+decorrelate the noise in the difference image.
+
+With the exposure variance planes, exactly the same steps are repeated than on
+the data planes, only the subtraction step is replaced by addition.
```
mrawls: Can you please elaborate on this sentence? I don't understand why addition is used instead of subtraction.

doc/lsst.ip.diffim/code_notes.rst
```diff
@@ -0,0 +1,183 @@
...
+difference result is *nan*.
+
+Pre-convolution is not implemented in the ZOGY algorithm. If
+``doPreConvolve==True`` the ``S`` the detection likelihood image
+(eq. 12 in [ZOGY2016]_) and its *corrected variance* ``S_var`` (the
+*denominator* of eq. 25 [ZOGY2016]_) are calculated following the
+*simple* correction steps presented in the paper Section 3.3. This
+signal correction is introduced to account for the source noise
+(bright sources) and also for other systematic noise sources. The
+iterative approach of section 3.5 is not implemented.
```
mrawls: This paragraph is a little confusing, because you say pre-convolution is not implemented and then say what happens if doPreConvolve is True. Would it be correct to say that only a simple/linear pre-convolution is implemented and that the recommended/more accurate iterative version is not? Are the limitations of the current implementation understood very well?

python/lsst/ip/diffim/dipoleFitTask.py
```diff
@@ -791,12 +791,12 @@ def fitDipole(self, source, tol=1e-7, rel_weight=0.1,
...
-                 of the dipole fitting optimization
-            - 2: pre-fit the background using linear least squares (as in 1), and use the parameter
-                 estimates from that fit as starting parameters for an integrated "re-fit" of the background
-                 as part of the overall dipole fitting optimization.
+                - 0: do not fit background at all
+                - 1 (default): pre-fit the background using linear least squares and then do not fit it
+                  as part of the dipole fitting optimization
+                - 2: pre-fit the background using linear least squares (as in 1), and use the parameter
+                  estimates from that fit as starting parameters for an integrated "re-fit" of the
+                  background as part of the overall dipole fitting optimization.
```
mrawls: Can you please explain why these small docstring formatting changes are part of this PR? (Here as well as imagePsfMatch.py and zogy.py.)

kgabor: These are just typos that break the docstring rendering. According to John, minor changes that don't deserve their standalone ticket can be simply added to issues as they emerge.

timj: Without this allowance the overhead to doing minor cleanups as you find them whilst working on something else would be too high. It's always more efficient to fix the little docstring problem whilst you notice it than have to file a ticket and delay the fix. So long as the fixups are on distinct commits (as here) then all is fine.

doc/lsst.ip.diffim/code_notes.rst
kgabor: Text updated.

doc/lsst.ip.diffim/code_notes.rst
kgabor: Reference to a "visit id" field originated from an old piece of command line argument description. Updated. The code in ``GetCalexpAsTemplateTask`` did not explicitly use it, though contained a hard-wired reference to 'visit' in the log messages. Removed.

doc/lsst.ip.diffim/code_notes.rst
kgabor: Done.

doc/lsst.ip.diffim/code_notes.rst
kgabor: It's not pre-convolution but a functionality with similar statistical meaning. Clarified.

doc/lsst.ip.diffim/code_notes.rst
kgabor: I think this was to follow the variance addition formula of the difference of uncorrelated random variables in the code. Thinking a bit more about it, this might not be even fully appropriate here, I have to re-check algorithm assumptions in the paper. Commented.

doc/lsst.ip.diffim/code_notes.rst
kgabor: Lots of refactoring could be done to ip_diffim...

doc/lsst.ip.diffim/code_notes.rst
kgabor: Added flowchart as a figure.

doc/lsst.ip.diffim/code_notes.rst
mrawls: OK, I trust you on this one!

PR #114, opened 2019-03-01
DM-16536: Migrate all metrics from ap.verify.measurements

kfindeisen:
This PR defines `MetricTask` subclasses for the `ip_diffim.numSciSources` and `ip_diffim.fracDiaSourcesToSciSources` metrics, which were previously implemented as functions in `ap_verify` (see lsst/ap_verify#63). This change requires that `ip_diffim` depend on `lsst.verify`.

Design note: most of the complexity of the task code comes from `MetricTask` instances not being allowed to assume a particular granularity for their metric. This, in turn, is partly because of Gen 2 (i.e., `MetricsControllerTask`) limitations, but largely because it's not yet known who will be responsible for metrics aggregation.

### DIFF
--- a/python/lsst/ip/diffim/metrics.py
+++ b/python/lsst/ip/diffim/metrics.py
@@ -1,0 +1,161 @@
+# This file is part of ip_diffim.
+#
+# Developed for the LSST Data Management System.
+# This product includes software developed by the LSST Project
+# (http://www.lsst.org).
+# See the COPYRIGHT file at the top-level directory of this distribution
+# for details of code ownership.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+
+__all__ = [
+    "NumberSciSourcesMetricTask", "NumberSciSourcesMetricConfig",
+    "FractionDiaSourcesToSciSourcesMetricTask", "FractionDiaSourcesToSciSourcesMetricConfig",
+]
+
+
+import astropy.units as u
+
+from lsst.pipe.base import Struct, InputDatasetField
+from lsst.verify import Measurement
+from lsst.verify.gen2tasks import MetricTask, register
+from lsst.verify.tasks import MetricComputationError
+
+
+class NumberSciSourcesMetricConfig(MetricTask.ConfigClass):
+    sources = InputDatasetField(
+        doc="The catalog of science sources.",
+        name="src",
+        storageClass="SourceCatalog",
+        dimensions={"Instrument", "Exposure", "Detector"},
+    )
+
+
+@register("numSciSources")
+class NumberSciSourcesMetricTask(MetricTask):
+    """Task that computes the number of cataloged science sources.
+    """
+    _DefaultName = "numSciSources"
+    ConfigClass = NumberSciSourcesMetricConfig
+
+    def run(self, sources):
+        """Count the number of science sources.
+
+        Parameters
+        ----------
+        sources : iterable of `lsst.afw.table.SourceCatalog`
+            A collection of science source catalogs, one for each unit of
+            processing to be incorporated into this metric. Its elements may
+            be `None` to represent missing data.
+
+        Returns
+        -------
+        result : `lsst.pipe.base.Struct`
+            A `~lsst.pipe.base.Struct` containing the following component:
+
+            ``measurement``
+                the total number of science sources (`lsst.verify.Measurement`
+                or `None`)
+        """
+        nSciSources = 0
+        inputData = False
+        for catalog in sources:
+            if catalog is not None:
+                nSciSources += len(catalog)
+                inputData = True
+
+        if inputData:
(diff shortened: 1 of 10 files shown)


python/lsst/ip/diffim/metrics.py
```diff
@@ -0,0 +1,158 @@
...
+        nameTemplate="{coaddName}Diff_diaSrc",
+        storageClass="SourceCatalog",
+        dimensions={"Instrument", "Exposure", "Detector"},
+    )
+
+
+@register("fracDiaSourcesToSciSources")
+class FractionDiaSourcesToSciSourcesMetricTask(MetricTask):
+    """Task that computes the ratio of difference image sources to science
+    sources in an image, visit, etc..
```
ebellm: extra period

tests/test_metrics.py
```diff
@@ -0,0 +1,203 @@
...
+
+    def testNoDataExpected(self):
+        result = self.task.run([])
+        meas = result.measurement
+        self.assertIsNone(meas)
+
+    def testGetInputDatasetTypes(self):
+        config = self.taskClass.ConfigClass()
+        types = self.taskClass.getInputDatasetTypes(config)
+        # dict.keys() is a collections.abc.Set, which has a narrower interface than __builtins__.set... WTF???
```
ebellm: Let's remove the "WTF" here ;-)

python/lsst/ip/diffim/metrics.py
```diff
@@ -0,0 +1,158 @@
...
+
+        for sciCatalog, diaCatalog in zip(sciSources, diaSources):
+            if diaCatalog is not None and sciCatalog is not None:
+                nSciSources += len(sciCatalog)
+                nDiaSources += len(diaCatalog)
+                inputData = True
+
+        if inputData:
+            metricName = self.getOutputMetricName(self.config)
+            if nSciSources <= 0.0 or nDiaSources <= 0.0:
```
ebellm: I'm not sure what the right behavior here should be if there are zero science sources, but returning 0 seems not quite right from a purely mathematical perspective.  Maybe return `None`?  I'm not sure `NaN` is more informative.

tests/test_metrics.py
```diff
@@ -0,0 +1,203 @@
...
+        self.assertEqual(meas.quantity, 0.0 * u.dimensionless_unscaled)
+
+    def testEmptySciCatalog(self):
+        sciCatalog = _makeDummyCatalog(0)
+        diaCatalog = _makeDummyCatalog(3)
+        result = self.task.run([sciCatalog], [diaCatalog])
+        meas = result.measurement
+
+        self.assertEqual(meas.metric_name, Name(metric="ip_diffim.fracDiaSourcesToSciSources"))
+        self.assertEqual(meas.quantity, 0.0 * u.dimensionless_unscaled)
```
ebellm: see above about whether this should be 0, `None`, or `NaN`

python/lsst/ip/diffim/metrics.py
```diff
@@ -0,0 +1,158 @@
...
+        doc="The catalog of science sources.",
+        name="src",
+        storageClass="SourceCatalog",
+        dimensions={"Instrument", "Exposure", "Detector"},
+    )
+
+
+@register("numSciSources")
+class NumberSciSourcesMetricTask(MetricTask):
+    """Task that computes the number of cataloged science sources.
```
ebellm: I'm suddenly worried we shouldn't call these "science" sources--see the ticket in JIRA.

python/lsst/ip/diffim/metrics.py
kfindeisen: FWIW, I copied the behavior from the original [fracUpdatedDiaObjects metric](https://github.com/lsst/ap_verify/blob/master/python/lsst/ap/verify/measurements/association.py#L132-L133).

The current `MetricTask` convention is that `None` means "not enough data" (e.g., the pipeline didn't run difference imaging); having both source catalogs, and them simply being empty, is not the same thing.

On the other hand, maybe `NaN` or `Inf` are acceptable values, since we've seen that it's also perfectly valid to measure `nSciSources == 0` and `nDiaSources > 0`. Though we should check with @afausti to make sure SQuaSH won't choke on it...

kfindeisen: Or I suppose we could raise `MetricComputationError`, though I'm not sure division by 0 counts as an "algorithmic error preventing calculation of the metric".

ebellm: I definitely don't think this is an error.  It's not clear to me that we need to preserve the semantic distinction between "no diffim" and "nothing in the source catalog", but I agree that if SQuaSH can handle `NaN` let's try that.  `Inf` just seems to be asking for trouble in the plots.

afausti: @ebellm @kfindeisen SQuaSH will store that and filter out the `NaN`s when displaying things, so it will be fine. However, it would be nice to sanitize the JSON. As pointed out in DM-13179 according to JSON [RFC 4627](https://www.ietf.org/rfc/rfc4627.txt) `NaN`s are not permitted.

RobertLuptonTheGood: It sounds as if you want us to use {{null}} instead, is that right?

kfindeisen: I think sanitizing the JSON must not be the responsibility of the `MetricTask` framework, since users can (and almost certainly will) create and store metrics without it. But it sounds like DM-13179 plans to tackle the root cause anyway.

@ebellm to me the distinction seems very clear: in the former case you don't have the data for the metric to be applicable, in the latter case you *do* have the data, you just don't like what it's telling you. But maybe this is a design issue we should discuss more generally, since I've already found the current policy hard to document.

kfindeisen: Following offline discussion with @ebellm, this and the tasks for other fractional metrics will raise `MetricComputationError` when the denominator is 0. The higher-level framework will handle things gracefully.

PR #116, opened 2019-03-19
DM-10156: Replace all uses of Calib with PhotoCalib

parejkoj:


parejkoj:
The build was failing because `lsst.afw.image.utils` is not imported. The way I changed it to is how it is used everywhere else: I'm somewhat surprised it worked before.

PR #120, opened 2019-04-30
Tickets/DM-18742: speed up DcrModel convergence

isullivan:
Add an option that amplifies the differences between DCR model planes to accelerate convergence. With default settings this is only applied within the footprints of detected sources, and only during the iterative forward modeling step of constructing the DcrModel.

Work on this ticket involved in-depth investigation of frequency regularization, and includes some changes to that implementation.

### DIFF
--- a/python/lsst/ip/diffim/dcrModel.py
+++ b/python/lsst/ip/diffim/dcrModel.py
@@ -299,7 +299,7 @@ class DcrModel:
 
     def buildMatchedTemplate(self, exposure=None, order=3,
                              visitInfo=None, bbox=None, wcs=None, mask=None,
-                             splitSubfilters=False):
+                             splitSubfilters=True, amplifyModel=1.):
         """Create a DCR-matched template image for an exposure.
 
         Parameters
@@ -320,7 +320,10 @@ class DcrModel:
             reference mask to use for the template image.
         splitSubfilters : `bool`, optional
             Calculate DCR for two evenly-spaced wavelengths in each subfilter,
-            instead of at the midpoint. Default: False
+            instead of at the midpoint. Default: True
+        amplifyModel : `float`, optional
+            Multiplication factor to amplify differences between model planes.
+            Used to speed convergence of iterative forward modeling.
 
         Returns
         -------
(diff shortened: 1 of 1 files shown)


python/lsst/ip/diffim/dcrModel.py
```diff
@@ -319,6 +321,9 @@ def buildMatchedTemplate(self, exposure=None, order=3,
         splitSubfilters : `bool`, optional
             Calculate DCR for two evenly-spaced wavelengths in each subfilter,
             instead of at the midpoint. Default: False
+        amplifyModel : `float`, optional
+            Amplify the differences between model planes.
+            Used to speed convergence of iterative forward modeling.
```
kfindeisen: Is the `float` some kind of multiplier (as I assume from the default)? If so, please say so.

python/lsst/ip/diffim/dcrModel.py
```diff
@@ -319,6 +321,9 @@ def buildMatchedTemplate(self, exposure=None, order=3,
         splitSubfilters : `bool`, optional
             Calculate DCR for two evenly-spaced wavelengths in each subfilter,
             instead of at the midpoint. Default: False
```
kfindeisen: Default is now `True`.

python/lsst/ip/diffim/dcrModel.py
```diff
@@ -369,14 +378,16 @@ def buildMatchedExposure(self, exposure=None,
...
+            bbox = exposure.getBBox()
         templateImage = self.buildMatchedTemplate(exposure=exposure, visitInfo=visitInfo,
                                                   bbox=bbox, wcs=wcs, mask=mask)
         maskedImage = afwImage.MaskedImageF(bbox)
-        maskedImage.image = templateImage
-        maskedImage.mask = self.mask
-        maskedImage.variance = self.variance
+        maskedImage.image = templateImage[bbox]
+        maskedImage.mask = self.mask[bbox]
+        maskedImage.variance = self.variance[bbox]
```
kfindeisen: What are these changes for?

isullivan: This was an unrelated bug I fixed on this branch. I'll move these changes to their own commit.

PR #125, opened 2019-06-06
DM-19978: Fix bug in DcrModel coordinates

isullivan:
Also expands the unit tests to cover a wider range of observing conditions, and adds a test that compares the DCR shift calculation to external coordinate transformations.

Note this branch was initially pushed as DM-19778, but was deleted because that was the wrong ticket number.

### DIFF
--- a/tests/test_dcrModel.py
+++ b/tests/test_dcrModel.py
@@ -285,58 +285,26 @@ class DcrModelTestTask(lsst.utils.tests.TestCase):
         transform the pixel coordinates to altitude and azimuth, add the DCR
         amplitude to the altitude, and transform back to pixel coordinates.
         """
-        dcrNumSubfilters = 3
         afwImageUtils.defineFilter("gTest", self.lambdaEff,
                                    lambdaMin=self.lambdaMin, lambdaMax=self.lambdaMax)
         filterInfo = afwImage.Filter("gTest")
-        lambdaEff = filterInfo.getFilterProperty().getLambdaEff()
         pixelScale = 0.2*arcseconds
+        doFlip = [False, True]
 
         for testIter in range(self.nRandIter):
             rotAngle = 360.*self.rng.rand()*degrees
             azimuth = 360.*self.rng.rand()*degrees
             elevation = (45. + self.rng.rand()*40.)*degrees  # Restrict to 45 < elevation < 85 degrees
             visitInfo = self.makeDummyVisitInfo(azimuth, elevation)
-            loc = EarthLocation(lat=visitInfo.getObservatory().getLatitude().asDegrees()*u.degree,
-                                lon=visitInfo.getObservatory().getLongitude().asDegrees()*u.degree,
-                                height=visitInfo.getObservatory().getElevation()*u.m)
-            wcs = self.makeDummyWcs(rotAngle, pixelScale, crval=visitInfo.getBoresightRaDec())
-            date = visitInfo.getDate()
-            time = Time(date.get(date.MJD, date.TAI), format='mjd', location=loc, scale='tai')
-            altaz = SkyCoord(alt=elevation.asDegrees(), az=azimuth.asDegrees(),
-                             unit='deg', obstime=time, frame='altaz', location=loc)
-            # The DCR calculations are performed at the boresight
-            ra0 = altaz.icrs.ra.degree*degrees
-            dec0 = altaz.icrs.dec.degree*degrees
-            x0, y0 = wcs.skyToPixel(afwGeom.SpherePoint(ra0, dec0))
-            dcrShifts = calculateDcr(visitInfo, wcs, filterInfo, dcrNumSubfilters)
-            refShifts = []
-            # We divide the filter into "subfilters" with the full wavelength range
-            # divided into equal sub-ranges.
-            for wl0, wl1 in wavelengthGenerator(filterInfo, dcrNumSubfilters):
-                # Note that diffRefractAmp can be negative,
-                # since it is relative to the midpoint of the full band
-                diffRefractAmp0 = differentialRefraction(wavelength=wl0, wavelengthRef=lambdaEff,
-                                                         elevation=elevation,
-                                                         observatory=visitInfo.getObservatory(),
-                                                         weather=visitInfo.getWeather())
-                diffRefractAmp1 = differentialRefraction(wavelength=wl1, wavelengthRef=lambdaEff,
-                                                         elevation=elevation,
-                                                         observatory=visitInfo.getObservatory(),
-                                                         weather=visitInfo.getWeather())
-                diffRefractAmp = (diffRefractAmp0 + diffRefractAmp1)/2.
-
-                elevation1 = elevation + diffRefractAmp
-                altaz = SkyCoord(alt=elevation1.asDegrees(), az=azimuth.asDegrees(),
-                                 unit='deg', obstime=time, frame='altaz', location=loc)
-                ra1 = altaz.icrs.ra.degree*degrees
-                dec1 = altaz.icrs.dec.degree*degrees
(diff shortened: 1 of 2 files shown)


tests/test_dcrModel.py
```diff
@@ -341,9 +424,9 @@ def testRotationFlipped(self):
                                         crval=visitInfo.getBoresightRaDec(),
                                         flipX=flip)
                 rotAngle = calculateImageParallacticAngle(visitInfo, wcs)
-                if flip:
+                if ~flip:
```
timj: This confuses me.  Do you mean `not flip`? I think that `flip` is always a Boolean, in which case this `if` statement seems to always be true (`~True == -2` and `~False == -1`), which now makes me doubt the test.

tests/test_dcrModel.py
```diff
@@ -241,13 +267,70 @@ def testDcrCalculation(self):
...
+            wcs = self.makeDummyWcs(rotAngle, pixelScale, crval=visitInfo.getBoresightRaDec())
+            date = visitInfo.getDate()
+            time = Time(date.get(date.MJD, date.TAI), format='mjd', location=loc, scale='tai')
+            # The DCR calculations are performed at the boresight
+            ra0 = visitInfo.getBoresightRaDec().getLongitude()
+            dec0 = visitInfo.getBoresightRaDec().getLatitude()
+            x0, y0 = wcs.skyToPixel(afwGeom.SpherePoint(ra0, dec0))
+            dcrShifts = calculateDcr(visitInfo, wcs, filterInfo, dcrNumSubfilters)
+            refShifts = []
+            for wl0, wl1 in wavelengthGenerator(filterInfo, dcrNumSubfilters):
```
timj: I'd like a comment here explaining what this for loop is doing.

tests/test_dcrModel.py
```diff
@@ -191,16 +193,33 @@ def makeDummyVisitInfo(self, azimuth, elevation):
...
-        dec = lsstLat + np.cos(azimuth.asRadians())*zenithAngle
+        loc = EarthLocation(lat=lsstLat.asDegrees()*u.degree,
+                            lon=lsstLon.asDegrees()*u.degree,
+                            height=lsstAlt*u.m)
+        airmass = 1.0/np.sin(elevation.asDegrees())
+
+        time = Time(2000., format='decimalyear', scale='tai', location=loc)
+        # Pick a random date and time within a 20-year span
+        time += 20*365*self.rng.rand()
+        altaz = SkyCoord(alt=elevation.asDegrees(), az=azimuth.asDegrees(),
```
timj: It might be clearer if this was an `AltAz` constructor directly rather than the more general `SkyCoord`.

tests/test_dcrModel.py
isullivan: Thanks for catching this! That was a dumb mistake: it should just be `if flip:`.

tests/test_dcrModel.py
```diff
@@ -241,13 +267,70 @@ def testDcrCalculation(self):
...
+        filterInfo = afwImage.Filter("gTest")
+        lambdaEff = filterInfo.getFilterProperty().getLambdaEff()
+        pixelScale = 0.2*arcseconds
+
+        for testIter in range(self.nRandIter):
+            rotAngle = 360.*self.rng.rand()*degrees
+            azimuth = 360.*self.rng.rand()*degrees
+            elevation = (45. + self.rng.rand()*40.)*degrees  # Restrict to 45 < elevation < 85 degrees
+            visitInfo = self.makeDummyVisitInfo(azimuth, elevation)
+            loc = EarthLocation(lat=visitInfo.getObservatory().getLatitude().asDegrees()*u.degree,
```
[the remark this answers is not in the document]
isullivan: The location is the same for each iteration, but it is calculated from `visitInfo` which is only available inside the loop. I could make the observatory coordinates globals, or construct a dummy `visitInfo` before the loop, but this felt cleaner. I'll add a comment.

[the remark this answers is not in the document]
timj: Or ask for a "real" VisitInfo outside the loop and calculate the `loc` once...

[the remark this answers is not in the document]
isullivan: I'm sorry, I don't understand what you mean. The visitInfo changes for each iteration since it requires the pointing information, so I don't know what it would mean to have it outside the loop.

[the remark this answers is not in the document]
timj: I mean call it additionally once outside the loop to get a visitInfo you are only going to use to create the location object. Then create new visitInfos within the loop as now.  It's probably not worth doing given that it's not going to appreciably slow down the loop.

tests/test_dcrModel.py
isullivan: That's a good point, though an advantage of the more general `SkyCoord` is that it allows convenient conversion to RA and Dec through `skycoord.icrs`. If I construct it as an `AltAz`, that requires an extra line of `radec = altaz.transform_to(ICRS())`. The two give the same results, so I feel it's a pretty even trade-off and I am fine using either way. What's your preference?

timj: Okay. Leave it then.

isullivan:
Please take another look. I've removed the hour angle calculation, and loosened the tolerances of the tests to account for the bug in `VisitInfo`.

tests/test_dcrModel.py
```diff
@@ -165,64 +175,85 @@ def makeDummyWcs(self, rotAngle, pixelScale, crval, flipX=True):
...
-                                       boresightRotAngle=0.*radians,
-                                       observatory=lsstObservatory,
-                                       weather=lsstWeather
-                                       )
+        loc = EarthLocation(lat=lsstLat,
+                            lon=lsstLon,
+                            height=lsstAlt)
+        airmass = 1.0/np.sin(elevation.asDegrees())
+
+        time = Time(2000., format='decimalyear', scale='tai', location=loc)
```
timj: Can you define this with `format="jyear"` so that you don't need to fiddle with the 0.5 day correction next line?
```python
>>> Time(2000.0, format="jyear", scale="tai").isot
'2000-01-01T12:00:00.000'
```

tests/test_dcrModel.py
```diff
@@ -165,64 +175,85 @@ def makeDummyWcs(self, rotAngle, pixelScale, crval, flipX=True):
...
+                                      boresight_rotation_coord='sky',
+                                      temperature=lsstTemperature,
+                                      pressure=lsstPressure,
+                                      relative_humidity=lsstHumidity,
+                                      tracking_radec=altaz.icrs,
+                                      altaz_begin=altaz,
+                                      observation_type='science',
+                                      )
+        makeVisitInfo = MakeRawVisitInfoViaObsInfo()
+        visitInfo = makeVisitInfo.observationInfo2visitInfo(obsInfo=obsInfo)
```
timj: This is a static method so it should be enough to do `MakeRawVisitInfoViaObsInfo.observationInfo2visitInfo()`.

tests/test_dcrModel.py
```diff
@@ -241,13 +272,74 @@ def testDcrCalculation(self):
...
+            time = Time(date.get(date.MJD, date.TAI), format='mjd', location=loc, scale='tai')
+            altaz = SkyCoord(alt=elevation.asDegrees(), az=azimuth.asDegrees(),
+                             unit='deg', obstime=time, frame='altaz', location=loc)
+            # The DCR calculations are performed at the boresight
+            ra0 = altaz.icrs.ra.degree*degrees
+            dec0 = altaz.icrs.dec.degree*degrees
+            x0, y0 = wcs.skyToPixel(afwGeom.SpherePoint(ra0, dec0))
+            dcrShifts = calculateDcr(visitInfo, wcs, filterInfo, dcrNumSubfilters)
+            refShifts = []
+            # We divide the filter into "subfilters" with with the full wavelength range
```
timj: "with with"

PR #129, opened 2019-08-06
DM-20554 Log or store if ip_diffim running in deconvolution case

kgabor:


python/lsst/ip/diffim/makeKernelBasisList.py
```diff
@@ -232,6 +252,8 @@ def generateAlardLuptonBasisList(config, targetFwhmPix=None, referenceFwhmPix=No
         # http://iopscience.iop.org/0266-5611/26/8/085002  Equation 40
 
         # Use specializations for deconvolution
+        logger.debug("Target psf fwhm is the greater, deconvolution mode")
```
isullivan: I think this should be a higher level of logging output than `debug`. Maybe even `warn`, but at least `info`.

python/lsst/ip/diffim/psfMatch.py
```diff
@@ -376,25 +376,26 @@ def setDefaults(self):
...
-        doc="Default sigma values in pixels of Gaussian kernel base functions. "
+        doc="Default sigma values in pixels of base Gaussians. "
             "List length must be `alardNGauss`.",
         default=(0.7, 1.5, 3.0),
     )
     alardGaussBeta = pexConfig.Field(
         dtype=float,
-        doc="""Multiplier of Gaussian sigmas between adjacent elements of the kernel basis list.""",
+        doc="Scaling multiplier of base Gaussian sigmas for automated sigma determination "
+            "(scaleByFwhm = True case).",
```
isullivan: scaleByFwhm should be in back-ticks, like `alardNGauss` in the entry above. I would also write this as "Used if `scaleByFwhm`=True". Or, something like that if that's inaccurate.

kgabor: Reworded.

python/lsst/ip/diffim/makeKernelBasisList.py
kgabor: Increased to `info`.

PR #131, opened 2019-08-28
DM-21065: Replace Policy with PropertySet

timj:


src/AssessSpatialKernelVisitor.cc
```diff
@@ -59,13 +59,13 @@ namespace detail {
...
-        _ps(ps),
-        _imstats(ImageStatistics<PixelT>(_ps)),
+        _ps(ps.deepCopy()),
+        _imstats(ImageStatistics<PixelT>(ps)),
         _nGood(0),
         _nRejected(0),
         _nProcessed(0),
-        _useCoreStats(_ps.getAsBool("useCoreStats")),
-        _coreRadius(_ps.getAsInt("candidateCoreRadius"))
+        _useCoreStats(ps.getAsBool("useCoreStart")),
```
ktlim: Is this a new typo or fixing an old one?

ktlim:
Sorry to think of this after all this work, but does it make sense to pull the `PropertySet` accesses out of the mainline code into the initialization of some kind of custom config struct/class, hence obviating the need for a copy of the `PropertySet` inside the class?

src/AssessSpatialKernelVisitor.cc
timj: I have no idea how I introduced it in my edits but the tests told me pretty quickly that I had broken it.

timj:
I think the changes are now complete in that this PR completely removes Policy and includes deprecation messages if you try to use Policy.

As for the copying issue, are you suggesting pre-emptively copying out solely the items from the PropertySet that are required by the class? (as is done in a couple of classes for a subset of parameters where they are assigned to instance properties directly).

timj:
I've had a quick go at storing the property set properties I need in the object. This does work but it will mean that every property that might be needed has to be present. Currently some properties are never read if certain booleans have different values, so you can have a call that works at the moment but which would fail if I force all possible properties to be read in advance.

include/lsst/ip/diffim/AssessSpatialKernelVisitor.h
```diff
@@ -31,7 +31,7 @@ namespace detail {
         AssessSpatialKernelVisitor(
             std::shared_ptr<lsst::afw::math::LinearCombinationKernel> spatialKernel,   ///< Spatially varying kernel
             lsst::afw::math::Kernel::SpatialFunctionPtr spatialBackground, ///< Spatially varying background
-            lsst::pex::policy::Policy const& policy                        ///< Policy file
+            lsst::daf::base::PropertySet const& ps                        ///< config
```
ktlim: Maybe call this a PropertySet or PS for consistency with other headers?

include/lsst/ip/diffim/BasisLists.h
```diff
@@ -42,15 +42,15 @@ namespace diffim {
     /**
      * @brief Build a regularization matrix for Delta function kernels
      *
-     * @param policy           Policy file dictating which type of matrix to make
+     * @param ps           PropertySet file dictating which type of matrix to make
```
ktlim: It's not a file.

include/lsst/ip/diffim/KernelCandidate.h
```diff
@@ -164,13 +164,13 @@ namespace diffim {
 
         /*
          * @note This method uses an estimate of the variance which is the
-         * straight difference of the 2 images.  If requested in the Policy
+         * straight difference of the 2 images.  If requested in the config
```
ktlim: PropertySet or "configuration PropertySet"?

src/ImageSubtract.cc
```diff
@@ -42,15 +42,15 @@
 #include "lsst/afw/image.h"
 #include "lsst/afw/math.h"
 #include "lsst/log/Log.h"
-#include "lsst/pex/policy/Policy.h"
+#include "lsst/daf/base/PropertySet.h"
```
ktlim: If it's not actually used, can this just be eliminated?

python/lsst/ip/diffim/deprecated.py
```diff
@@ -0,0 +1,63 @@
...
+        Known positions of likely `~lsst.pex.Policy` arguments for the wrapped
+        function.  Can be out of range since some pybind11 constructors take
+        different numbers of arguments.  If `None` all arguments will be
+        checked.
+    """
+
+    @functools.wraps(func)
+    def internal(*args, **kwargs):
+        newargs = list(args)
+        nonlocal policy_args
```
ktlim: Since you're not assigning to it, I think this isn't required, but I guess it helps make things clear.

timj: I'll take another look. flake8 was complaining unless I added the `nonlocal`.

PR #139, opened 2019-11-26
Tickets/dm 11991 -- Fix relative calibration between images for ZOGY algorithm

wmwv:
Sets the ZOGY Fn, Fr normalizations based on the photometric calibration of each image with new `scaleByCalibration` option in `ZogyTask` config.  Option is set to `True` by default.


parejkoj:
Also, your note in the PR description says "scaleByCalibration is set to True by default", but looking at the code, it defaults to False.

wmwv:
> Also, your note in the PR description says "scaleByCalibration is set to True by default", but looking at the code, it defaults to False.
🤦‍♂ 
Thank you.  Fixed
(and rebased to latest `origin/master`).

PR #147, opened 2020-03-04
one pixel shift

shuliu2017:
1. Fix the one pixel offset bug.
When padding a PSF with (odd, odd) dimension to (even, even) dimension, the center of the PSF should be placed to the center-right of the array. Otherwise the peak value could be shifted to the end of the array after applying np.fft.fftshift(e.g. Correct padding: [0,0,1,2,3,2,1,0] -> [3,2,1,0,0,0,1,2], Incorrect padding: [0,1,2,3,2,1,0,0] - > [2,1,0,0, 0,1,2,3]). This corresponds to a one pixel offset since the np.fft.fft assume the first component of the array as the origin.
2. Add a test file for checking the pixel offset of the ZOGY D image.

### DIFF
--- a/python/lsst/ip/diffim/zogy.py
+++ b/python/lsst/ip/diffim/zogy.py
@@ -346,7 +346,7 @@ class ZogyTask(pipeBase.Task):
             The padded copy of the input `psf`.
         """
         newArr = np.zeros(size)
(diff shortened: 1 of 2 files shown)


jdswinbank:
Thank you for this, @shuliu2017! I'm closing this PR so that we can handle it on #148.

PR #148, opened 2020-03-05
DM-11990: Fix one-pixel offset for ZOGY.

wmwv:
Fix one-pixel offset for ZOGY.

Duplicate of #147 
But this PR is from a branch named after the ticket so it will be automatically linked from JIRA.


wmwv:
@kgabor Will you have a chance to look at this PR this week?

PR #149, opened 2020-03-10
DM-23558: Add PhotoCalib for DcrModels

isullivan:


isullivan:
It's optional because I think you only care about it if you are going to make an exposure. The basic DcrModel is only a masked image, and doesn't carry around metadata like that in general. But, I did want to provide a mechanism for someone to construct a DCR matched exposure this way if they wanted to (say, in a notebook).

PR #152, opened 2020-04-01
DM-22952: Gen 3 version of DCR image differencing

isullivan:


python/lsst/ip/diffim/dcrModel.py
```diff
@@ -150,6 +150,44 @@ def fromDataRef(cls, dataRef, datasetType="dcrCoadd", numSubfilters=None, **kwar
...
+        -------
+        dcrModel : `lsst.pipe.tasks.DcrModel`
+            Best fit model of the true sky after correcting chromatic effects.
+        """
+        filterInfo = None
+        psf = None
+        mask = None
+        variance = None
+        photoCalib = None
+        modelImages = [[] for _s in range(len(availableCoaddRefs))]
```
kgabor, line 173: There should be exactly one image per subfilter in a stored model shouldn't be ?  I'd recommend to initialize `modelImages` with a list of `None`-s as they are simply replaced later. What about `[None, ] * len(availableCoaddRefs)` ?

PR #154, opened 2020-04-13
DM-24371 Implement fixed correction fixed PSF support decorrelation afterburner

kgabor:


python/lsst/ip/diffim/imageDecorrelation.py
```diff
@@ -196,235 +196,239 @@ def run(self, exposure, templateExposure, subtractedExposure, psfMatchingKernel,
...
+        """Compute the (decorrelated) difference image's new PSF.
 
         Parameters
         ----------
-        kernel : `numpy.array`
-            a numpy.array
+        corrft : `numpy.ndarray`
+            The frequency space representation of the correction calculated by
+            `computeCorrection`. Shape must be `self.freqSpaceShape`.
+        psfArr : `numpy.ndarray`
```
isullivan, line 379: `psfArr` made me think of an array of psfs at first. Perhaps `psfVals` or `psfOld`? Also, the current implementation modifies `psfArr` in place (as well as returning it), so you should document that if you leave that behavior.

python/lsst/ip/diffim/imageDecorrelation.py
```diff
@@ -196,235 +196,239 @@ def run(self, exposure, templateExposure, subtractedExposure, psfMatchingKernel,
...
-    @staticmethod
-    def _doConvolve(exposure, kernel):
-        """Convolve an Exposure with a decorrelation convolution kernel.
+        Notes
+        ----
+        There is no algorithmic guarantee that the corrected psf can
+        meaningfully fit to the same size as the original one.
+        """
+        psfShape = psfArr.shape
+        psfArr = self.padCenterOriginArray(psfArr, self.freqSpaceShape)
```
isullivan, line 393: I think you want to use `psfNew = self.padCenterOriginArray(psfArr, self.freqSpaceShape)` here, and then use `psfNew` for the rest of the calculations.

python/lsst/ip/diffim/imageDecorrelation.py
```diff
@@ -513,7 +517,7 @@ def run(self, subExposure, expandedSubExposure, fullBBox,
         self.log.setLevel(logLevel)  # reset the log level
 
         diffim = res.correctedExposure.Factory(res.correctedExposure, subExposure.getBBox())
-        out = pipeBase.Struct(subExposure=diffim, decorrelationKernel=res.correctionKernel)
+        out = pipeBase.Struct(subExposure=diffim, )
```
isullivan, line 520: Again, just return `diffim` instead of wrapping it in a `Struct`

tests/test_imageDecorrelation.py
```diff
@@ -122,17 +122,18 @@ def makeFakeImages(size=(256, 256), svar=0.04, tvar=0.04, psf1=3.3, psf2=2.2, of
     im1 = np.random.normal(scale=np.sqrt(svar), size=x0im.shape)  # variance of science image
```
isullivan, line 122: It's slightly out of scope, but since you're modifying `makeFakeImages` it would be helpful if you could convert the docstring to numpydoc format (commenting here because I can't leave a comment on line 77).

tests/test_imageDecorrelation.py
```diff
@@ -347,6 +399,38 @@ def testDiffimCorrection(self):
...
+        self.assertFloatsAlmostEqual(rho_rawdiff[0], 1., atol=0.1, rtol=None)
+        self.assertFloatsAlmostEqual(rho_corrdiff[0], 1., atol=0.1, rtol=None)
+
+        self.assertFloatsAlmostEqual(rho_sci[1:], 0., atol=0.1, rtol=None)
+
+        self.assertGreater(rho_rawdiff[1], 0.2)
+        self.assertGreater(rho_rawdiff[2], 0.2)
+        self.assertGreater(rho_rawdiff[3], 0.2)
+
+        self.assertFloatsAlmostEqual(rho_corrdiff[1:], 0., atol=0.1, rtol=None)
```
isullivan, line 432: Please define variables with meaningful names to hold the numbers used in the comparisons. Or, add a comment explaining why we should expect, e.g. `rho_rawdiff[1]` to always be greater than 0.2

python/lsst/ip/diffim/imageDecorrelation.py
```diff
@@ -196,235 +196,239 @@ def run(self, exposure, templateExposure, subtractedExposure, psfMatchingKernel,
                     np.all(np.isnan(templateExposure.getMaskedImage().getImage().getArray()))):
                 self.log.warn('Template or science image is entirely NaNs: skipping decorrelation.')
                 outExposure = subtractedExposure.clone()
-                return pipeBase.Struct(correctedExposure=outExposure, correctionKernel=None)
+                return pipeBase.Struct(correctedExposure=outExposure, )
+
+        tOverSVar = tvar/svar
+        if tOverSVar > 1e8:
```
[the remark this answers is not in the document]
kgabor: As the matching kernel high frequency components go to zero, the correction factor converges to sqrt(tvar/svar) i.e. this is a warning when high frequency pixels of the difference image is multiplied by more than 4 orders of magnitude as correction. Yes, in the regular case, the template variance is the smaller one and this is no concern; this can happen if we convolve the science image and thus svar becomes << tvar due to the convolution.

I don't really have a good justification for the value itself at the moment, just wanted an indicator to watch for diverging solutions, we'll see whether this is going to be met in practice.

python/lsst/ip/diffim/imageDecorrelation.py
```diff
@@ -142,34 +143,33 @@ def run(self, exposure, templateExposure, subtractedExposure, psfMatchingKernel,
...
-
+        result : `lsst.pipe.base.Struct`
             - ``correctedExposure`` : the decorrelated diffim
-            - ``correctionKernel`` : the decorrelation correction kernel (which may be ignored)
 
         Notes
         -----
-        The `subtractedExposure` is NOT updated
+        The `subtractedExposure` is NOT updated. The returned `correctedExposure` has an updated but fixed
+        PSF.
```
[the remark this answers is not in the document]
kgabor: Updated. _Fixed_ meant _not spatially varying_ in this case.

python/lsst/ip/diffim/imageDecorrelation.py
```diff
@@ -142,34 +143,33 @@ def run(self, exposure, templateExposure, subtractedExposure, psfMatchingKernel,
...
         tvar : `float`, optional
             Image variance for template image
             If `None` (default) then compute the variance over the entire input template image.
 
         Returns
         -------
-        result : `Struct`
-            a `lsst.pipe.base.Struct` containing:
-
+        result : `lsst.pipe.base.Struct`
```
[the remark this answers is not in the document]
kgabor: This is a result of dropping the _correction kernel_ from the return. I am not 100% sure that we don't want to return any additional information; I'll postpone the removal of the Struct for the spatially varying correction implementation DM-23857. 

Considerations: It'd be logical to return something about the correction, but... 

-  returning the Fourier space correction needs the freqSpaceShape and may not suit for any outside operation.
- Returning the image space correction (full image size) needs one additional, otherwise unnecessary, IFT. Also where the origin is placed would need detailed description or additional, otherwise unnecessary, copying (re-centering) of the array. 
- If a matchingKernel size cropped version is returned then we throw away the main lesson from this whole change i.e. that the correction can be larger than the original matching kernel in general so should not be used.

python/lsst/ip/diffim/imageDecorrelation.py
```diff
@@ -196,235 +196,239 @@ def run(self, exposure, templateExposure, subtractedExposure, psfMatchingKernel,
...
-        fkernel : `numpy.ndarray`
-            a 2-d numpy.array containing the correction kernel
+        None
 
         Notes
         -----
-        As currently implemented, kappa is a static (single, non-spatially-varying) kernel.
+        For each dimension, gets the smallest even number greater than or equal to
+        `N1+N2-1` where `N1` and `N2` are the two largest values.
+        In case of only one shape given, rounds up to even each dimension value.
```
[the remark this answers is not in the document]
kgabor: Done.

python/lsst/ip/diffim/imageDecorrelation.py
```diff
@@ -196,235 +196,239 @@ def run(self, exposure, templateExposure, subtractedExposure, psfMatchingKernel,
...
-            changed = True
-        if (out.shape[1] % 2) == 1:
-            out = np.pad(out, ((0, 0), (1, 0)), mode='constant')
-            changed = True
-        if changed:
-            out *= (np.mean(kernel) / np.mean(out))  # need to re-scale to same mean for FFT
-        return out
+        kappa = self.padCenterOriginArray(kappa, self.freqSpaceShape)
+        kft = np.fft.fft2(kappa)
+        kft2 = np.conj(kft) * kft
```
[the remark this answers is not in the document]
kgabor: My original consideration here was that converting them to real then back to complex when applying the returned correction may take more time. But this is not the case, np.real creates a view and taking the sqrt of reals is much faster. See notebook attached to the ticket.

python/lsst/ip/diffim/imageDecorrelation.py
kgabor: `padCenterOriginArray` always creates a copy.

kgabor: However `computeCorrectedImage` indeed made an unintended in place modification...

PR #156, opened 2020-04-25
DM-23173: Change W504 to W503 and fix flake8 warnings

r-owen:


python/lsst/ip/diffim/imagePsfMatch.py
```diff
@@ -848,9 +848,9 @@ def makeCandidateList(self, templateExposure, scienceExposure, kernelSize, candi
             try:
                 candidateList[0].getSource()
             except Exception as e:
-                raise RuntimeError("Candidate List is of type: %s. " % (type(candidateList[0])) +
-                                   "Can only make candidate list from list of afwTable.SourceRecords, " +
-                                   "measAlg.PsfCandidateF or other type with a getSource() method: %s" % (e))
+                raise RuntimeError(f"Candidate List is of type: {type(candidateList[0])} "
+                                   "Can only make candidate list from list of afwTable.SourceRecords, "
+                                   "measAlg.PsfCandidateF or other type with a getSource() method") from e
```
timj, line 853: People probably still might like the exception string in the new exception string.

r-owen: I thought the point of "raise...from" was to avoid the need for that. But I will put it back.

r-owen: I think the `from e` is not actually useful here (it will be obvious what is going on) so I'll restore the original -- the RuntimeError message will contain the error and no `from e`.

TallJimbo: Please keep the `from e` or add `from None`, if it's not too late - without one of those, exceptions raised inside `except` blocks get a confusing "while processing this exception, another one was raised" (or something like that) message.

(I'm sure the original code was written before that behavior existed, but it's definitely something we should be fixed in re-raises as we encounter them.)

r-owen: It is too late. But in any case I strongly feel that the `from` is not useful here. This is just a way to test if getSource() works and complain that the data is the wrong type if not. That is all the user really needs to know. I especially think it would be wasteful to quote the original exception and also raise from it.

I suppose one could call `hasattr` instead to save the first exception, though that would not catch the case that `hasattr` exists but cannot be called as a function with no arguments.

TallJimbo: I get that it's too late, but for future scenarios, this is exactly what `from None` is for.

PR #160, opened 2020-06-10
DM-25115 Implement zogy subtraction with non-spatial PSF

kgabor:


tests/test_pixelOffset.py
```diff
@@ -40,16 +40,17 @@ def _find_max(data):
...
 
     def testPixelOffset(self):
         """ Test whether the peak position is at [300][300].
         """
         self._setUpImages()
         config = ZogyConfig()
-        task = ZogyTask(templateExposure=self.imrex, scienceExposure=self.imnex, config=config)
-        D_F = task.computeDiffim(inImageSpace=False)
-        max_loc = PixelOffsetTest._find_max(D_F.D.image.array)
+        config.scaleByCalibration = False
```
isullivan, line 51: I believe you can do `config = ZogyConfig(scaleByCalibration=False)`

tests/test_imageDecorrelation.py
```diff
@@ -74,8 +74,9 @@ def singleGaussian2d(x, y, xc, yc, sigma_x=1., sigma_y=1., theta=0., ampl=1.):
 def makeFakeImages(size=(256, 256), svar=0.04, tvar=0.04, psf1=3.3, psf2=2.2, offset=None,
                    psf_yvary_factor=0., varSourceChange=1/50., theta1=0., theta2=0.,
                    n_sources=500, seed=66, verbose=False):
-    """Make two exposures: science and template pair with flux sources and random noise.
-    In all cases below, index (1) is the science image, and (2) is the template.
+    """Deprecated. Use ``test_zogy : makeFakeImages``. DM-25115: This function is kept for
+    numerical compatibility with existing test cases only. It has x,y axes handling
+    and PSF centering bugs.
```
isullivan, line 79: This would be a good time to convert the test cases over to only use the version of makeFakeImages that's in test_zogy. We really don't want two versions of this code.

tests/test_zogy.py
```diff
@@ -25,14 +25,14 @@
...
 import lsst.meas.algorithms as measAlg
+import lsst.daf.base as dafBase
 
-from test_imageDecorrelation import makeFakeImages
+from test_imageDecorrelation import singleGaussian2d
 
-from lsst.ip.diffim.zogy import ZogyTask, ZogyConfig, ZogyMapReduceConfig, \
-    ZogyImagePsfMatchConfig, ZogyImagePsfMatchTask
-from lsst.ip.diffim.imageMapReduce import ImageMapReduceTask
+from lsst.ip.diffim.zogy import ZogyTask, ZogyConfig
```
isullivan, line 35: Minor point, but please alphabetize the lsst imports

tests/test_zogy.py
```diff
@@ -44,6 +44,171 @@ def setup_module(module):
...
+    overfitting -- perhaps to the source itself). This might be fixed by
+    adding more constant sources.
+    """
+    rng = np.random.default_rng(seed)
+
+    psf1 = [3.3, 3.3] if psf1 is None else psf1
+    if not hasattr(psf1, "__len__") and not isinstance(psf1, str):
+        psf1 = [psf1, psf1]
+    psf2 = [2.2, 2.2] if psf2 is None else psf2
+    if not hasattr(psf2, "__len__") and not isinstance(psf2, str):
```
isullivan, line 108: This seems to be a relic from some previous version of the code, because it won't work if psf1 or psf2 are `str`. Maybe just remove the `and not isinstance(psf2, str)`?

tests/test_zogy.py
```diff
@@ -44,6 +44,171 @@ def setup_module(module):
...
+    yim = np.arange(-size[1]//2, size[1]//2, 1)
+    x0im, y0im = np.meshgrid(xim, yim)
+
+    im1 = rng.normal(scale=np.sqrt(svar), size=x0im.shape)  # variance of science image
+    im2 = rng.normal(scale=np.sqrt(tvar), size=x0im.shape)  # variance of template
+
+    if n_sources > 0:
+        fluxes = rng.uniform(50, 30000, n_sources)
+        xposns = rng.uniform(xim.min()+16, xim.max()-5, n_sources)
+        yposns = rng.uniform(yim.min()+16, yim.max()-5, n_sources)
```
isullivan, line 127: Add a space around + or - (here and below)

tests/test_zogy.py
```diff
@@ -44,6 +44,171 @@ def setup_module(module):
...
+
+        # Make the source closest to the center of the image the one that increases in flux
+        ind = np.argmin(xposns**2. + yposns**2.)
+
+        # vary the y-width of psf across x-axis of science image (zero means no variation):
+        psf1_yvary = psf_yvary_factor * (yim.mean() - yposns) / yim.max()
```
isullivan, line 133: Remove spaces around * and / (here and below)

tests/test_zogy.py
```diff
@@ -44,6 +44,171 @@ def setup_module(module):
...
+
+    def makeExposure(imgArray, psfArray, imgVariance):
+        """! Convert an image numpy.array and corresponding PSF numpy.array into an exposure.
+
+        Add the (constant) variance plane equal to `imgVariance`.
+
+        @param imgArray 2-d numpy.array containing the image
+        @param psfArray 2-d numpy.array containing the PSF image
+        @param imgVariance variance of input image
+        @return a new exposure containing the image, PSF and desired variance plane
```
isullivan, line 187: Please convert the docstring to numpydoc style

tests/test_zogy.py
```diff
@@ -44,6 +44,171 @@ def setup_module(module):
...
+        """
+        # All this code to convert the template image array/psf array into an exposure.
+        bbox = geom.Box2I(geom.Point2I(0, 0), geom.Point2I(imgArray.shape[1]-1, imgArray.shape[0]-1))
+        im1ex = afwImage.ExposureD(bbox)
+        im1ex.getMaskedImage().getImage().getArray()[:, :] = imgArray
+        im1ex.getMaskedImage().getVariance().getArray()[:, :] = imgVariance
+        psfBox = geom.Box2I(geom.Point2I(-12, -12), geom.Point2I(12, 12))  # a 25x25 pixel psf
+        psf = afwImage.ImageD(psfBox)
+        psfBox.shift(geom.Extent2I(-(-size[0]//2), -(-size[1]//2)))  # -N//2 != -(N//2) for odd numbers
+        im1_psf_sub = psfArray[psfBox.getMinY():psfBox.getMaxY()+1, psfBox.getMinX():psfBox.getMaxX()+1]
```
isullivan, line 197: Add space around +

tests/test_zogy.py
```diff
@@ -44,6 +44,171 @@ def setup_module(module):
...
+
+        Add the (constant) variance plane equal to `imgVariance`.
+
+        @param imgArray 2-d numpy.array containing the image
+        @param psfArray 2-d numpy.array containing the PSF image
+        @param imgVariance variance of input image
+        @return a new exposure containing the image, PSF and desired variance plane
+        """
+        # All this code to convert the template image array/psf array into an exposure.
+        bbox = geom.Box2I(geom.Point2I(0, 0), geom.Point2I(imgArray.shape[1]-1, imgArray.shape[0]-1))
```
isullivan, line 190: Add space around -

python/lsst/ip/diffim/zogy.py
```diff
@@ -330,748 +127,715 @@ def _computeVarianceMean(self, exposure):
...
-            res = task.computeScorr(**kwargs)
-            D = res.S
+            scoreSubExposure = None
 
-        outExp = D.Factory(D, subExposure.getBBox())
-        out = pipeBase.Struct(subExposure=outExp)
-        return out
+        return pipeBase.Struct(diffSubExp=diffSubExposure, scoreSubExp=scoreSubExposure)
 
+    def run(self, exposure1, exposure2, calculateS=True):
```
isullivan, line 793: Please rename `calculateS` to `calculateScorr` (or `calculateScore` if that's what the variable is called)

python/lsst/ip/diffim/zogy.py
```diff
@@ -330,748 +127,715 @@ def _computeVarianceMean(self, exposure):
...
+            Photometric scaling of the images. See eqs. (5)--(9)
 
-class ZogyMapper(ZogyTask, ImageMapper):
-    """Task to be used as an ImageMapper for performing
-    ZOGY image subtraction on a grid of subimages.
-    """
-    ConfigClass = ZogyConfig
-    _DefaultName = 'ip_diffim_ZogyMapper'
+        calculateS : `bool`
+            If True, calculate and return the detection significance (score) image.
```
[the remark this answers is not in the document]
kgabor: Sorry about the `S` vs. `Scorr` confusion: The ZOGY paper calls the image pixel wise constructed of the detection significance statistic the "score" image, and is denoted by `S`. In later sections of the paper, there are "corrections" defined to this score image, the first of these is the calculation of its variance plane and dividing S by its per pixel sigma . (Ideally in the zogy image model, there is no "variance plane" as all pixel variance equals to one number, and S would have a variance of 1.) 

In the present implementation, only this first "correction", the variance plane of S is calculated; but I decided that the division can be done "outside" of ZogyTask if one explicitly wants S in units of sigma directly. S is returned as an Exposure anyway. So there is no point calling it `Scorr` at this point, but I plan to add the other correction tags later.

python/lsst/ip/diffim/zogy.py
```diff
@@ -330,748 +127,715 @@ def _computeVarianceMean(self, exposure):
...
+        # if sDenom < tiny then it can be == 0. -> `denom` = 0. and 0/0 occur at `c1` , `c2`
+        # if we keep SDenom = tiny, denom ~ O(sqrt(tiny)), Pd ~ O(sqrt(tiny)), S ~ O(sqrt(tiny)*tiny) == 0
+        # Where S = 0 then Pd = 0 and D should still yield the same variance ~ O(1)
+        # For safety, we set S = 0 explicitly, too, though it should be unnecessary.
+        fltZero = sDenom < tiny
+        nZero = np.sum(fltZero)
+        self.log.debug(f"Handling {nZero} both PSFs are zero points.")
+        if nZero > 0:
+            fltZero = np.nonzero(fltZero)  # We expect only a small fraction of such frequencies
+            sDenom[fltZero] = tiny  # Avoid division problem but overwrite result anyway
```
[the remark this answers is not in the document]
kgabor: Yes, it works without converting the _bool_ into an _index array_ (of ints). This supposed to be a speed optimization. If only a small fraction of a large array is reassigned, it is faster to index the array with an array of ints rather than with a bool array.

python/lsst/ip/diffim/zogy.py
```diff
@@ -330,748 +127,715 @@ def _computeVarianceMean(self, exposure):
...
+        # mImgs can be in-place modified
+        if correctBackground:
+            self.subtractImageMean(mImg1.image, mImg1.mask, self.statsControl)
+            self.subtractImageMean(mImg2.image, mImg2.mask, self.statsControl)
+
+        psfBBox1 = exposure1.getPsf().computeBBox()
+        psfBBox2 = exposure2.getPsf().computeBBox()
+        self.psfShape1 = (psfBBox1.getHeight(), psfBBox1.getWidth())
+        self.psfShape2 = (psfBBox2.getHeight(), psfBBox2.getWidth())
+        self.imgShape = (mImg1.getHeight(), mImg1.getWidth())
```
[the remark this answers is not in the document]
kgabor: getDimensions() return Extent of x, y; here I switch to the numpy shape convention which is y, x. I leave it to be explicit about the order of dimensions.

python/lsst/ip/diffim/zogy.py
```diff
@@ -330,748 +127,715 @@ def _computeVarianceMean(self, exposure):
...
-
-    def computeScorr(self, xVarAst=0., yVarAst=0., inImageSpace=None, padSize=0, **kwargs):
-        """Wrapper method to compute ZOGY corrected likelihood image, optimal for
-        source detection
-
-        This method should be used as the public interface for
-        computing the ZOGY S_corr.
+        s = np.sum(A)
+        if s == 0.:
+            A = np.fabs(A)
```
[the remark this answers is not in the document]
kgabor: `A = np.fabs(A)` just replaces a local variable with a new object (new array), it does not change the original array.

python/lsst/ip/diffim/zogy.py
```diff
@@ -330,748 +127,715 @@ def _computeVarianceMean(self, exposure):
...
-        `expandedSubExposure` to allow for invalid edge pixels arising
-        from convolutions, which are then removed.
+        ``varMean1``, ``varMean2`` quantities are part of the noise model and not to be confused
+        with the variance of image frequency components or with ``varPlane1``, ``varPlane2`` that
+        are the Fourier transform of the variance planes.
+        """
+        var1F2Sq = varMean1*F2*F2
+        var2F1Sq = varMean2*F1*F1
+        # We need reals for comparison, also real operations are usually faster
+        psfAbsSq1 = np.real(np.conj(psf1)*psf1)
```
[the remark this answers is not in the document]
kgabor: I think it's faster. `np.abs(psf1)**2` contains a square root operation then squaring again.

python/lsst/ip/diffim/zogy.py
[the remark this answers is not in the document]
isullivan: You should add a comment that you are switching axis conventions from x,y to numpy's y,x here.

python/lsst/ip/diffim/zogy.py
```diff
@@ -330,748 +127,715 @@ def _computeVarianceMean(self, exposure):
...
+            subExposure2 = self.fullExp2
+        else:
+            subExposure2 = self.fullExp2.Factory(self.exposure2, bbox2)
+
+        if subExposure1.getDimensions() != subExposure2.getDimensions():
+            raise ValueError("Subexposure dimensions do not match.")
+
+        if psf1 is None:
+            self.subExpPsf1 = self.computePsfAtCenter(subExposure1)
+        if psf2 is None:
```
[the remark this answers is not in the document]
kgabor: Good catch.

python/lsst/ip/diffim/zogy.py
```diff
@@ -330,748 +127,715 @@ def _computeVarianceMean(self, exposure):
...
-    def computeScorrImageSpace(self, xVarAst=0., yVarAst=0., padSize=None, **kwargs):
-        """Compute corrected likelihood image, optimal for source detection
-
-        Compute ZOGY S_corr image. This image can be thresholded for
-        detection without optimal filtering, and the variance image is
-        corrected to account for astrometric noise (errors in
-        astrometric registration whether systematic or due to effects
-        such as DCR). The calculations here are all performed in
-        Real (image) space.
+        D = np.real(np.fft.ifft2(D))
```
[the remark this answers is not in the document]
kgabor: `irfft` and `ifft` interpret their inputs differently (the negative frequencies are not present for `irfft`). I need to check more thoroughly that the conjugate symmetry is kept everywhere and then all fft transformations may be replaced with the "real" functions to speed up and save memory.

python/lsst/ip/diffim/zogy.py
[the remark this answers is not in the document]
kgabor: `s==0.` should not really occur for PSFs, they're normed to 1.

python/lsst/ip/diffim/zogy.py
```diff
@@ -330,748 +127,715 @@ def _computeVarianceMean(self, exposure):
...
+        diffSubExposure.mask = self.calculateMaskPlane(self.subExposure1.mask, self.subExposure2.mask)
+
+        # PhotoCalib does not support ImageD.
+        # calib = afwImage.PhotoCalib(1./ftDiff.Fd)
+        # calibImg = calib.calibrateImage(
+        #     afwImage.MaskedImage(diffSubExposure.maskedImage, deep=True, dtype=np.float32))
+        # diffSubExposure.maskedImage = calibImg
+        diffSubExposure.maskedImage /= ftDiff.Fd
+
+        # Now the subExposure calibration is 1. everywhere
```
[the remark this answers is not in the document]
kgabor: At the moment, we strictly don't need to apply the calibration but we will need to do it once the subtraction is done on a grid and we stitch the sub images together.

tests/test_imageDecorrelation.py
kgabor: I actually tried to get rid of it but it escalates into the decorrelation afterburner code/tests  - it is a better time to remove (and rewrite/recalculate the test cases) when we have settled the spatially varying versions for both the decorrelation afterburner and zogy differencing.

PR #169, opened 2020-11-02
DM-24331 Replace deprecated getCentroidKey calls.

kgabor:


kgabor:
This change was finally done by DM-22814.

PR #170, opened 2021-01-02
DM-27170: Deprecate and phase out afw::image::Filter

parejkoj:


kfindeisen (approved):
Not sure I understand the point of renaming all the variables, but otherwise looks good.

PR #171, opened 2021-01-14
DM-25050 Fix non-normalized matching kernel case in the decorrelation afterburner

kgabor:


python/lsst/ip/diffim/imageDecorrelation.py
```diff
@@ -367,17 +367,34 @@ def computeCorrection(self, kappa, svar, tvar, preConvArr=None):
...
         kft = np.fft.fft2(kappa)
-        kft2 = np.real(np.conj(kft) * kft)
-        if preConvArr is None:
-            denom = svar + tvar * kft2
-        else:
+        kftAbsSq = np.real(np.conj(kft) * kft)
+        # If there is no pre-convolution kernel, use placeholder scalars
+        preSum = 1.
+        preAbsSq = 1.
+        if preConvArr is not None:
```
morriscb, line 377: I've been scolded for using `is not None` which can be a bit hard to read. Could you change this structure back to the if/else setting the `preSum`, `preAbsSq` in the `if is None` statement.

python/lsst/ip/diffim/imageDecorrelation.py
```diff
@@ -367,17 +367,34 @@ def computeCorrection(self, kappa, svar, tvar, preConvArr=None):
...
-        kft = np.sqrt((svar + tvar) / denom)
+            preK = np.fft.fft2(preConvArr)
+            preAbsSq = np.real(np.conj(preK)*preK)
+
+        denom = svar * preAbsSq + tvar * kftAbsSq
+        # Division by zero protection, though we don't expect to hit it
+        # (rather we'll have numerical noise)
+        tiny = np.finfo(kftAbsSq.dtype).tiny * 1000.
+        flt = denom < tiny
+        if np.sum(flt) > 0:
```
morriscb, line 388: As I asked during standup: would it be possible to test for this bit of code int he unittests?

Another option, since on the ticket you mention this should be a rare (or possibly impossible) occurrence, would it make sense to add a warning here for when the `if` statement is tripped?

morriscb (changes requested):
Looks fine mostly. Some restructuring of an `if` statement and adding a warn when the divide by zero code is needed.

morriscb (approved):

python/lsst/ip/diffim/imageDecorrelation.py
kgabor: Added a warning for this case.

PR #174, opened 2021-02-09
DM-28442 Fix variance plane calculation scaling in non-normalized matching kernel case

kgabor:
* Add templateMatched to ImageDecorrelation and add variance scaling in science convolution mode.
 * Add the new option to pass on in the DecorrelateALSpatialTask mapper layer.


morriscb (approved):
Looks good. Needs some clean up of the comments around where I noted as it doesn't read very clearly.

PR #176, opened 2021-02-10
DM-26372: Units wrong in deepDiff_diaSrc catalog

mrawls:
This PR fixes instFluxErr units, which were incorrectly set as pixel instead of count.


kherner (approved):
CI build successful.

PR #180, opened 2021-04-14
DM-29703: Do not use skyMap to find overlapping patches in Gen3.

yalsayyad:
runQuantum has the candidate patches that spatially matches the detector.

### DIFF
--- a/python/lsst/ip/diffim/getTemplate.py
+++ b/python/lsst/ip/diffim/getTemplate.py
@@ -24,6 +24,7 @@ import numpy as np
 
 import lsst.afw.image as afwImage
 import lsst.geom as geom
+import lsst.sphgeom as sphgeom
 import lsst.pex.config as pexConfig
 import lsst.pipe.base as pipeBase
 from lsst.ip.diffim.dcrModel import DcrModel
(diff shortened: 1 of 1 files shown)


TallJimbo:
Just saw this fly by, commenting in case it's useful: remember that you may well have a better WCS for the science image than the butler had when it generated the regions (and then padded them, knowing that its WCS wasn't). So it's probably there still is more filtering you could do here.

erykoff (approved):
Looks good, a couple of minor comments.

yalsayyad:
@TallJimbo This commit does include a geometric overlap check (based on the calexp.wcs).  Difference is that it doesn't require the patches to fall in the same tract that `skyMap.findTract(ctrSkyPos)` says. Registry has more info (like that we're on 9813 right now, not 9812). 

There's a 'fail gently" version that goes beyond:  https://github.com/lsst/ip_diffim/tree/u/yusra%2Fdiffim_fail_gently which configurably doesn't bother to difference images if < ~2% of calexp has template coverage. But then I didn't know what to do with `forcedPhotDiffim` crying that it couldn't find `goodSeeingDiff_differenceExp`, and nothing workable solutions pair-coding that could be done about it today.

PR #181, opened 2021-04-19
DM-27031: Use primary flag in source count metrics

kfindeisen:
This PR modifies the metrics that count direct-imaging sources to ignore any sources that don't have the `detect_isPrimary` set (if available).

### DIFF
--- a/tests/test_metrics.py
+++ b/tests/test_metrics.py
@@ -22,6 +22,7 @@
 import unittest
 
 import astropy.units as u
+from astropy.tests.helper import assert_quantity_allclose
 
 from lsst.afw.table import SourceCatalog
 import lsst.utils.tests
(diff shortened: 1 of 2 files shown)


tests/test_metrics.py
```diff
@@ -163,6 +178,16 @@ def testSkySources(self):
...
 
+    def testPrimarySources(self):
+        sciCatalog = _makeDummyCatalog(5, skyFlag=True, priFlag=True)
+        diaCatalog = _makeDummyCatalog(3)
+        result = self.task.run(sciCatalog, diaCatalog)
+        lsst.pipe.base.testUtils.assertValidOutput(self.task, result)
+        meas = result.measurement
+
+        self.assertEqual(meas.metric_name, Name(metric="ip_diffim.fracDiaSourcesToSciSources"))
+        self.assertEqual(meas.quantity, len(diaCatalog) * u.dimensionless_unscaled)
```
kgabor, line 189: Also at lines 112, 179: All `Quantities` are dtype float. Is it safe to use `assertEqual` (instead of `assertFloatsAlmostEqual` or similar) in general?

kgabor (approved):

PR #183, opened 2021-04-22
DM-28955: Allow passing `basisSigGauss` parameter to `makeKernelBasisList`

arunkannawadi:
This PR enables an option to pass `basisSigGauss` - the size parameter of the basis function to `makeKernelBasisList` function at run time instead of looking up the frozen values from the config. This allows to control the kernel parametrization during PSF Gaussianization stage for GAaP fluxes/colors. This has to be at runtime and not passed as a config parameter because the size depends on the seeing at the location of each source.


cmsaunders (approved):
Everything looks good!

PR #184, opened 2021-05-05
DM-29490 Decorrelation afterburner fix for science image convolution case

kgabor:


kherner (commented):
I think these changes look good though I have one question about the following syntax:

> subtractedMaskedImage = subtractedExposure.maskedImage
 subtractedMaskedImage[:, :] = results.warpedExposure.maskedImage[:, :]

Does the second line not overwrite the first (then there's no point in the first line), and if not, exactly what parts of subtractedMaskedImage get replaced in the second line?

kgabor:
> Does the second line not overwrite the first (then there's no point in the first line), and if not, exactly what parts of subtractedMaskedImage get replaced in the second line?

`A=B` _by definition_ always makes A, and B refer to the same object (and throwing away the original object in A, if any) in Python. A[expr] = B however is a special syntax, where the object in A can define the behavior for assignment. (As such, not equivalent to C= A[expr], C = B)

afw Images, Exposures behave as container objects similarly to numpy.ndarrays. They can contain their own pixel memory or "view" (slices of) other object's memory. 

> subtractedMaskedImage[:, :] = results.warpedExposure.maskedImage[:, :]

This copies all the pixel values from warpedExposure to subtractedMaskedImage keeping the objects' currently allocated memories.

This was the buggy statement:

> subtractedExposure.setMaskedImage(results.warpedExposure.getMaskedImage())

because it's equivalent to

> subtractedExposure.maskedImage = results.warpedExposure.maskedImage

that _replaces the image object_ in subtractedExposure and thus all further in-place pixel modification appears in warpedExposure as well.

 
> Does the second line not overwrite the first (then there's no point in the first line), and if not, exactly what parts of subtractedMaskedImage get replaced in the second line?

kherner:
> afw Images, Exposures behave as container objects similarly to numpy.ndarrays. They can contain their own pixel memory or "view" (slices of) other object's memory.

Thanks; this is what was missing from my understanding.

kherner (approved):

PR #187, opened 2021-05-07
DM-30079: Corrupted documentation breaks documentation builds

kfindeisen:
This PR fixes a duplicate `Notes` that was causing Sphinx to crash, as well as some less-urgent Sphinx warnings.


kgabor (approved):
Looks all right.

PR #188, opened 2021-05-25
DM-30281 Demonstrate the effect of random padding regeneration and fix unit test by making padding deterministic

kgabor:


python/lsst/ip/diffim/zogy.py
```diff
@@ -247,11 +247,12 @@ def initializeSubImage(self, fullExp, innerBox, outerBox, noiseMeanVar, useNoise
         borderBoxes = self.splitBorder(innerBox, outerBox)
         # Initialize the border region that are not fully within the exposure
         if useNoise:
+            rng = np.random.default_rng(seed=np.array([noiseMeanVar]).view(int))
```
kfindeisen, line 250: I realize this is preexisting code, but I'd recommend using an explicit generator (e.g., `Generator(PCG64(seed))`) rather than `random.default_rng`. Since the default generator is subject to change in future Numpy versions, using a specific algorithm is more reproducible.

kfindeisen (approved):

PR #190, opened 2021-06-09
DM-30683: Remove diaSourceAnalysis.py

parejkoj:


mrawls:
Oh you did make a PR, Jira is just full of lies.

mrawls (approved):
Assuming you combed through the stack and didn't find this used anywhere, let's get rid of it.

PR #191, opened 2021-07-09
DM-31066: Suppress log message spam

isullivan:
Code beyond image differencing is now using `makeKernelBasisList` and the image differencing-related log messages are too numerous. Adjust the logging to be less verbose, except when used for image differencing.

### DIFF
--- a/python/lsst/ip/diffim/imagePsfMatch.py
+++ b/python/lsst/ip/diffim/imagePsfMatch.py
@@ -433,14 +433,14 @@ class ImagePsfMatchTask(PsfMatchTask):
                 self.log.info("scienceFwhmPix: {}".format(scienceFwhmPix))
 
         if convolveTemplate:
-            kernelSize = makeKernelBasisList(self.kConfig, templateFwhmPix, scienceFwhmPix)[0].getWidth()
+            kernelSize = self.makeKernelBasisList(templateFwhmPix, scienceFwhmPix)[0].getWidth()
             candidateList = self.makeCandidateList(
                 templateExposure, scienceExposure, kernelSize, candidateList)
             results = self.matchMaskedImages(
                 templateExposure.getMaskedImage(), scienceExposure.getMaskedImage(), candidateList,
                 templateFwhmPix=templateFwhmPix, scienceFwhmPix=scienceFwhmPix)
         else:
-            kernelSize = makeKernelBasisList(self.kConfig, scienceFwhmPix, templateFwhmPix)[0].getWidth()
+            kernelSize = self.makeKernelBasisList(scienceFwhmPix, templateFwhmPix)[0].getWidth()
             candidateList = self.makeCandidateList(
                 templateExposure, scienceExposure, kernelSize, candidateList)
             results = self.matchMaskedImages(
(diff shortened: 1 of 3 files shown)


python/lsst/ip/diffim/imagePsfMatch.py
```diff
@@ -870,6 +870,48 @@ def makeCandidateList(self, templateExposure, scienceExposure, kernelSize, candi
 
         return candidateList
 
+    def makeKernelBasisList(self, targetFwhmPix=None, referenceFwhmPix=None,
+                            basisDegGauss=None, basisSigmaGauss=None, metadata=None):
+        """Wrapper for `lsst.ip.diffim.makeKernelBasisList` to set log messages consistently.
```
laurenam, line 875: This line is too long (93 > 79 characters).  Also, I'm not sure "consistently" is the best descriptor here (i.e consistent with what?)

python/lsst/ip/diffim/imagePsfMatch.py
```diff
@@ -870,6 +870,48 @@ def makeCandidateList(self, templateExposure, scienceExposure, kernelSize, candi
...
+                                        referenceFwhmPix=referenceFwhmPix,
+                                        basisDegGauss=basisDegGauss,
+                                        basisSigmaGauss=basisSigmaGauss,
+                                        metadata=metadata)
+        if targetFwhmPix == referenceFwhmPix:
+            self.log.info("Target and reference psf fwhms are equal, falling back to config values")
+        elif referenceFwhmPix > targetFwhmPix:
+            self.log.info("Reference psf fwhm is the greater, normal convolution mode")
+        else:
+            self.log.info("Target psf fwhm is the greater, deconvolution mode")
```
laurenam, line 911: So, if one was to run diffim in debug mode, does this mean you'll get double entries (one at `info` and one at `debug`)?

python/lsst/ip/diffim/modelPsfMatch.py
```diff
@@ -451,8 +451,8 @@ def _buildCellSet(self, exposure, referencePsfModel):
             paddingPix = self.config.padPsfBy
 
         if paddingPix > 0:
-            self.log.info("Padding Science PSF from (%s, %s) to (%s, %s) pixels" %
-                          (psfSize, psfSize, paddingPix + psfSize, paddingPix + psfSize))
+            self.log.debug("Padding Science PSF from (%d, %d) to (%d, %d) pixels",
+                           psfSize, psfSize, paddingPix + psfSize, paddingPix + psfSize)
```
laurenam, line 455: You're ok with losing this one to `debug`?

laurenam (approved):

python/lsst/ip/diffim/modelPsfMatch.py
isullivan: Yes, I think that's a good level for diffim, also.

python/lsst/ip/diffim/imagePsfMatch.py
isullivan: It does, but I think we can deal with that since the logs can be long in debug mode.

python/lsst/ip/diffim/makeKernelBasisList.py
```diff
@@ -260,7 +260,7 @@ def generateAlardLuptonBasisList(config, targetFwhmPix=None, referenceFwhmPix=No
         # http://iopscience.iop.org/0266-5611/26/8/085002  Equation 40
 
         # Use specializations for deconvolution
-        logger.info("Target psf fwhm is the greater, deconvolution mode")
+        logger.debug("Target psf fwhm is the greater, deconvolution mode")
```
arunkannawadi, line 263: This particular deconvolution mode will never get triggered in GAaP by construction. And if at all it gets triggered (either due to a bug in GAaP or in diffim), I think we'd want to know immediately. So perhaps, this should/could stay at INFO level?

PR #193, opened 2021-07-13
DM-6056: replace CONST_PTR and PTR macros by std::shared

mwittgen:


src/DipoleAlgorithms.cc
```diff
@@ -358,14 +358,14 @@ std::pair<double,int> PsfDipoleFlux::chi2(
     geom::Point2D negCenter(negCenterX, negCenterY);
     geom::Point2D posCenter(posCenterX, posCenterY);
 
-    CONST_PTR(afw::detection::Footprint) footprint = source.getFootprint();
+    std::shared_ptr<afw::detection::Footprint const > footprint = source.getFootprint();
```
kfindeisen, line 361:
```suggestion
    std::shared_ptr<afw::detection::Footprint const> footprint = source.getFootprint();
```

kfindeisen (approved):

PR #202, opened 2021-11-08
DM-31801: Use lsst prefix for logger name

timj:


isullivan (approved):
Looks good, the changes seem trivial in this package.

PR #203, opened 2021-11-10
DM-31777: Build CoaddPsf of CoaddPsfs in tract WCS

yalsayyad:
Previously it was loading the PSFs pre-warped to the detector WCS.
CoaddPsf warps when evaluating anyway so pre-warping is overkill
and harder to debug.


TallJimbo (approved):
Looks like there are a few more lines you can remove now, because `warped` doesn't survive outside that `for` loop:

652: `xyTransform = afwGeom.makeWcsPairTransform(coaddPatch.getWcs(), finalWcs)`
653: `psfWarped = WarpedPsf(coaddPatch.getPsf(), xyTransform)`
661: `warped.setPsf(psfWarped)`

You may not need to import `WarpedPsf` after that, either.

PR #208, opened 2022-02-28
DM-33822 Fix WCS bug in DCR templates

isullivan:


python/lsst/ip/diffim/dcrModel.py
```diff
@@ -422,24 +389,29 @@ def buildMatchedTemplate(self, exposure=None, order=3,
...
-        refModel = self.getReferenceImage(bbox)
+        refModel = None
         for subfilter, dcr in enumerate(dcrShift):
+            if self[subfilter] is None:
+                # It is possible to load only a single DCR subfilter at a time.
+                self.log.debug("Skipping missing DCR model subfilter %d", subfilter)
+                continue
             if amplifyModel > 1:
+                if refModel is None:
+                    refModel = self.getReferenceImage(bbox)
```
kherner, line 414: I'm not quite following the logic here. Before one would run self.getReferenceImage(bbox) once before the for loop. Now there is code to run it only the first time through the loop. What's the advantage? The only one I see is in case the first time through has a None subfilter, you would skip it. Am I missing something else?

kherner (commented):
Everything looks OK except for the one comment in dcrModel.py

python/lsst/ip/diffim/dcrModel.py
isullivan: The advantage is that `getReferenceImage()` does not have to be run during image differencing. `amplifyModel` is an option that is only used while building the DcrCoadds. I'll add a comment to make it more clear.

PR #209, opened 2022-02-28
DM-33843: Fix the logger name in the DipoleFit plugin

timj:
This logger did not have an "lsst." prefix and now it does. I also switched the logger over to python logging to match the current policy.


isullivan (approved):
Looks fine to me.

PR #212, opened 2022-03-22
DM-32756: Compute average FWHM of the PSF on an exposure as the media…

surhudm:
…n of PSF determined on a grid of points


arunkannawadi:
Changing the base to a corresponding ticket branch within `lsst` so we can test this more thoroughly and merge it to `main`. Thanks @surhudm !

PR #214, opened 2022-05-04
DM-34644: Clean up decorrelation code

isullivan:


python/lsst/ip/diffim/imageDecorrelation.py
```diff
@@ -737,7 +738,7 @@ def run(self, subExposure, expandedSubExposure, fullBBox,
         subExp1 = templateExposure.Factory(templateExposure, expandedSubExposure.getBBox())
 
         # Prevent too much log INFO verbosity from DecorrelateALKernelTask.run
-        logLevel = self.log.getLevel()
+        logLevel = self.log.level
         self.log.setLevel(self.log.WARNING)
```
timj, line 742: I really need to add the logging context manager to the `utils` package so that the log level resets when the task below completes. I'm not overly happy with how we handle this since it really is saying that this task is too verbose by default and it's impossible to ever see the info messages. As written, if someone sets the log level to DEBUG in pipetask they won't get any debug messages out of this task and that seems wrong.

tests/test_imageDecorrelation.py
```diff
@@ -449,7 +449,7 @@ def testNoiseDiffimCorrection(self):
 
         self._setUpSourcelessImages(svar=svar, tvar=tvar)
         diffExp, mKernel, expected_var = self._makeAndTestUncorrectedDiffim()
-        corrected_diffExp = self._runDecorrelationTask(diffExp, mKernel)
+        corrected_diffExp = self._runDecorrelationTask(diffExp.clone(), mKernel)
```
kherner, line 452: Just for my edification, why the change to .clone() here?

kherner: Other places too, actually....

kherner (approved):

tests/test_imageDecorrelation.py
isullivan: That's so I can modify the difference exposure in place in the decorrelation task, and not effect the tests.

python/lsst/ip/diffim/imageDecorrelation.py
isullivan: I completely agree that code that forces the log level like this is problematic. This is code that is not used outside of the unit tests, for now at least, and I just wanted to change this line to fix a warning while building the package. This particular task will need to be refactored if we want to use it in production.

PR #215, opened 2022-05-12
DM-34770: Use positions explicitly in PSF methods

arunkannawadi:


erykoff (approved):

python/lsst/ip/diffim/modelPsfMatch.py
```diff
@@ -416,8 +416,7 @@ def _buildCellSet(self, exposure, referencePsfModel):
 
         sciencePsfModel = exposure.getPsf()
 
-        dimenR = referencePsfModel.getLocalKernel().getDimensions()
-        psfWidth, psfHeight = dimenR
+        dimenR = referencePsfModel.getLocalKernel(scienceBBox.getCenter()).getDimensions()
```
isullivan, line 419: This is probably fine. In other places we have used `sciencePsfModel.getAveragePosition()` instead of `scienceBBox.getCenter()`, but I think it makes sense to use the center of the science image.

isullivan (approved):

python/lsst/ip/diffim/imagePsfMatch.py
```diff
@@ -423,14 +425,16 @@ def matchExposures(self, templateExposure, scienceExposure,
             if not templateExposure.hasPsf():
                 self.log.warning("No estimate of Psf FWHM for template image")
             else:
-                templateFwhmPix = self.getFwhmPix(templateExposure.getPsf())
+                templateFwhmPix = self.getFwhmPix(templateExposure.getPsf(),
+                                                  templateExposure.getBBox().getCenter())
```
erykoff, line 429: I think here and below should explicitly say `position=`

arunkannawadi: Sure, but do you agree that that this is the appropriate position to use?

python/lsst/ip/diffim/modelPsfMatch.py
arunkannawadi: I have changed it to use the center of the bbox explicitly. This makes a significant difference, say, when one passes in a postage stamp of an exposure. The PSF's average position may be far away from the stamp that is provided.

python/lsst/ip/diffim/imagePsfMatch.py
erykoff: Looking at the code, I believe that this is run _after_ the warping, so that the science and template bboxes should match.  In this case we want to evaluate the psf at the same position (center of bbox) rather than at different positions (individual psf average positions).  So I think this is correct.

arunkannawadi:
Looks like replacing PSF average position with center of the bbox isn't robust and we need changes from DM-32756 for the proposed changes here to not fail. So leaving TODO comments for now and will merge once Jenkins passes.

python/lsst/ip/diffim/imagePsfMatch.py
yalsayyad: I set the CoaddPsf's average position to be the *center of mass* of the template. 
This is not the same as the center of the bbox! 
Please keep the existing behavior by evaluating at `psf.getAveragePosition()`

arunkannawadi: I removed the comment referring to DM-32756 if this was intentional.

arunkannawadi: The changes in this snippet breaks ci_hsc, so I removed it. It's an outdated change.

PR #217, opened 2022-05-19
DM-33001: Refactor ImageDifferenceTask

isullivan:
This is a complete re-write of ImageDifferenceTask, which has also been moved from pipe_tasks.
Image differencing is now broken apart into three steps:

1. Constructing a warped template (completed previously on a separate ticket)
2. PSF-matching the template or science image, and subtracting the template from the science image
3. Running diaSource detection and measurement on the difference image


kherner (approved):

python/lsst/ip/diffim/utils.py
```diff
@@ -1078,3 +1078,23 @@ def detectDipoleSources(self, doMerge=True, diffim=None, detectSigma=5.5, grow=3
...
+    ----------
+    psf : `lsst.afw.detection.Psf`
+        Point spread function (PSF) to evaluate.
+
+    Returns
+    -------
+    psfSize : `float`
+        The FWHM of the PSF computed at its average position.
+    """
+    # TODO: there's an identical one in makeKernel: can we share them?
```
parejkoj, line 1096: We can dump this comment, since we took care of it.

tests/test_subtractTask.py
```diff
@@ -0,0 +1,459 @@
...
+    rngNoise = np.random.RandomState(noiseSeed)
+    x0, y0 = bbox.getBegin()
+    xSize, ySize = bbox.getDimensions()
+    xLoc = rng.rand(nSrc)*(xSize - 2*bufferSize) + bufferSize + x0
+    yLoc = rng.rand(nSrc)*(ySize - 2*bufferSize) + bufferSize + y0
+
+    flux = (rng.rand(nSrc)*(fluxRange - 1.) + 1.)*fluxLevel
+    sigmas = [psfSize for src in range(nSrc)]
+    coordList = list(zip(xLoc, yLoc, flux, sigmas))
+    # kernelSize = int(xSize/2)  # Need a careful explanation of this kernel size choice
```
parejkoj, line 105: unused code?

tests/test_subtractTask.py
```diff
@@ -0,0 +1,459 @@
...
+    skyLevel = 0
+    # Don't use the built in poisson noise: it modifies the global state of numpy random
+    modelExposure = plantSources(bbox, kernelSize, skyLevel, coordList, addPoissonNoise=False)
+    modelExposure.setWcs(makeFakeWcs())
+    noise = rngNoise.randn(ySize, xSize)*noiseLevel
+    # + rngNoise.randn(ySize, xSize)*np.sqrt(np.abs(modelExposure.image.array)))
```
parejkoj, line 111: Unused code?

tests/test_subtractTask.py
```diff
@@ -0,0 +1,459 @@
...
+    def test_few_sources(self):
+        """Test with only 1 source, to check that we get a useful error.
+        """
+        xSize = 256
+        ySize = 256
+        science, sources = makeTestImage(psfSize=2.4, nSrc=1, xSize=xSize, ySize=ySize)
+        template, _ = makeTestImage(psfSize=2.0, nSrc=1, xSize=xSize, ySize=ySize)
+        config = subtractImages.AlardLuptonSubtractTask.ConfigClass()
+        task = subtractImages.AlardLuptonSubtractTask(config=config)
+        with self.assertRaises(lsst.pex.exceptions.Exception):
```
parejkoj, line 239: We don't want to check for a generic exception here, we want to check for a specific one, and that the error message is useful (using `self.assertRaisesRegex`). I don't know whether the existing behavior is actually what we want, or whether the subtractTask should do a `raise SomeException("not enough sources") from e` or something like that: I recall that the exception we get in this case is not very helpful.

tests/test_subtractTask.py
```diff
@@ -0,0 +1,459 @@
...
+        templateVarianceOrig = template.variance.array[...]
+        for scaleFactor in [1., scaleFactor]:
+            science.variance.array[...] = scienceVarianceOrig/scaleFactor
+            template.variance.array[...] = templateVarianceOrig/scaleFactor
+            scienceVarMean = np.mean(science.variance.array)
+            templateVarMean = np.mean(template.variance.array)
+            config.doSubtractBackground = False
+            for doDecorrelation in [True, False]:
+                for doScaleVariance in [True, False]:
+                    config.doDecorrelation = doDecorrelation
```
parejkoj, line 327: This inner block should be a separate `_check_???` method called with all of the loop variables as args: otherwise it can be difficult to determine which branch of the loop actually failed.

tests/test_subtractTask.py
```diff
@@ -0,0 +1,459 @@
...
+        # Generate a random aperture correction map
+        apCorrMap = lsst.afw.image.ApCorrMap()
+        for name in ("a", "b", "c"):
+            apCorrMap.set(name, lsst.afw.math.ChebyshevBoundedField(science.getBBox(), rng.randn(3, 3)))
+        science.info.setApCorrMap(apCorrMap)
+
+        config = subtractImages.AlardLuptonSubtractTask.ConfigClass()
+        config.mode = "convolveTemplate"
+        for doDecorrelation in [True, False]:
+            config.doDecorrelation = doDecorrelation
```
parejkoj, line 371: Can the internal of this loop, and the science one below, become one `_check_something` method? That would also make the tests easier to interpret, because you could then drop the `for` loop and just call it directly with `True` and `False` on separate lines.

tests/test_subtractTask.py
```diff
@@ -0,0 +1,459 @@
...
+                psfOutImg = psfOut.computeKernelImage(psfAvgPos)
+                self.assertImagesAlmostEqual(psfImg, psfOutImg)
+
+            # check PSF, WCS, bbox, filterLabel, photoCalib, aperture correction
+            self.compare(apCorrMap, output.difference.info.getApCorrMap())
+            self.assertWcsAlmostEqualOverBBox(science.getWcs(), output.difference.getWcs(), science.getBBox())
+            self.assertEqual(science.getFilterLabel(), output.difference.getFilterLabel())
+            self.assertEqual(science.getPhotoCalib(), output.difference.getPhotoCalib())
+
+    def compare(self, a, b):
```
parejkoj, line 434: `_compare_apCorrMaps`? And docstrings for the parameters?

python/lsst/ip/diffim/makeKernel.py
```diff
@@ -0,0 +1,336 @@
...
+            yPos = cand.getCentroid()[1]
+            cand = diffimLib.makeKernelCandidate(xPos, yPos, tmi, smi, ps)
+
+            self.log.debug("Candidate %d at %f, %f", cand.getId(), cand.getXCenter(), cand.getYCenter())
+            kernelCellSet.insertCandidate(cand)
+
+        return kernelCellSet
+
+    def _adaptCellSize(self, candidateList):
+        """NOT IMPLEMENTED YET.
```
parejkoj, line 334: Is this true? There's code below.

python/lsst/ip/diffim/makeKernel.py
```diff
@@ -0,0 +1,336 @@
...
+        -------
+        candidateList : `list` of `dict`
+            A list of dicts having a "source" and "footprint"
+            field for the Sources deemed to be appropriate for Psf
+            matching
+
+        Raises
+        ------
+        RuntimeError
+            Description
```
parejkoj, line 209: There are several of these `Descriptions` that were missed.

python/lsst/ip/diffim/makeKernel.py
```diff
@@ -0,0 +1,336 @@
...
+            meas.algorithms.PsfCandidateF.
+        preconvolved : bool, optional
+            Description
+
+        Returns
+        -------
+        candidateList : `list` of `dict`
+            A list of dicts having a "source" and "footprint"
+            field for the Sources deemed to be appropriate for Psf
+            matching
```
parejkoj, line 204: Period at end of sentence.

python/lsst/ip/diffim/makeKernel.py
```diff
@@ -0,0 +1,336 @@
...
+        # Minimal set of measurments for star selection
+        self.selectMeasurement.algorithms.names.clear()
+        self.selectMeasurement.algorithms.names = ('base_SdssCentroid', 'base_PsfFlux', 'base_PixelFlags',
+                                                   'base_SdssShape', 'base_GaussianFlux', 'base_SkyCoord')
+        self.selectMeasurement.slots.modelFlux = None
+        self.selectMeasurement.slots.apFlux = None
+        self.selectMeasurement.slots.calibFlux = None
+
+
+class MakeKernelTask(PsfMatchTask):
```
parejkoj, line 76: docstring

python/lsst/ip/diffim/makeKernel.py
```diff
@@ -0,0 +1,336 @@
...
+    )
+    selectMeasurement = lsst.pex.config.ConfigurableField(
+        target=SingleFrameMeasurementTask,
+        doc="Initial measurements used to feed stars to kernel fitting",
+    )
+
+    def setDefaults(self):
+        # High sigma detections only
+        self.selectDetection.reEstimateBackground = False
+        self.selectDetection.thresholdValue = 10.0
```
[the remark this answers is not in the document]
isullivan: This is only for selecting stars to use for calculating the PSF matching kernel, not source detection and measurement. So we want to use only high sigma detections as the comment says.

python/lsst/ip/diffim/makeKernel.py
```diff
@@ -0,0 +1,336 @@
...
+            bkgd = np.ma.median(miArr)
+
+        # Take off background for detection
+        mi -= bkgd
+        try:
+            table.setMetadata(self.selectAlgMetadata)
+            detRet = self.selectDetection.run(
+                table=table,
+                exposure=exposure,
+                sigma=sigma,
```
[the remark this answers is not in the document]
isullivan: `None` is actually allowed; I'll add a comment.

tests/test_subtractTask.py
```diff
@@ -0,0 +1,459 @@
...
+                                         background=background_model,
+                                         xSize=xSize, ySize=ySize, x0=x0, y0=y0)
+        config = subtractImages.AlardLuptonSubtractTask.ConfigClass()
+        config.doSubtractBackground = True
+
+        config.makeKernel.kernel.name = "AL"
+        config.makeKernel.kernel.active.fitForBackground = True
+        config.makeKernel.kernel.active.spatialKernelOrder = 1
+        config.makeKernel.kernel.active.spatialBgOrder = 2
+        for mode in ["convolveTemplate", "convolveScience"]:
```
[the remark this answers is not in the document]
isullivan: Not for this test of background subtraction, but I can add a new test that checks the behavior of auto mode for different conditions.

python/lsst/ip/diffim/subtractImages.py
```diff
@@ -0,0 +1,529 @@
...
+        doc="Sources measured on the science exposure; "
+            "used to select sources for making the matching kernel.",
+        dimensions=("instrument", "visit", "detector"),
+        storageClass="SourceCatalog",
+        name="{fakesType}src"
+    )
+
+
+class SubtractImageOutputConnections(lsst.pipe.base.PipelineTaskConnections,
+                                     dimensions=_dimensions,
```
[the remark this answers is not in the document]
isullivan: I believe both are used by the `difference` and `matchedTemplate` output connections.

python/lsst/ip/diffim/subtractImages.py
```diff
@@ -0,0 +1,529 @@
...
+
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+        self.makeSubtask("decorrelate")
+        self.makeSubtask("makeKernel")
+        if self.config.doScaleVariance:
+            self.makeSubtask("scaleVariance")
+
+        self.convolutionControl = lsst.afw.math.ConvolutionControl()
+        self.convolutionControl.setDoNormalize(False)
```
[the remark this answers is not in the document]
isullivan: I thought this was necessary, but the tests all pass without it. I will see if I can write a test that explicitly covers normalization.

python/lsst/ip/diffim/subtractImages.py
```diff
@@ -0,0 +1,529 @@
...
+        deprecated="This option is only for backwards compatibility purposes"
+        " and will be removed after v24.",
+    )
+
+    def setDefaults(self):
+        self.makeKernel.kernel.name = "AL"
+        self.makeKernel.kernel.active.fitForBackground = self.doSubtractBackground
+        self.makeKernel.kernel.active.spatialKernelOrder = 1
+        self.makeKernel.kernel.active.spatialBgOrder = 2
+        if self.forceCompatibility:
```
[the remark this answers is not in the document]
isullivan: `False`

python/lsst/ip/diffim/subtractImages.py
```diff
@@ -0,0 +1,529 @@
...
+                                          templateMatched=False)
+
+        return lsst.pipe.base.Struct(difference=correctedExposure,
+                                     matchedTemplate=template,
+                                     backgroundModel=kernelResult.backgroundModel,
+                                     psfMatchingKernel=kernelResult.psfMatchingKernel,)
+
+    def finalize(self, template, science, difference, kernel,
+                 templateMatched=True,
+                 preConvMode=False,
```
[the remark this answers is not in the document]
isullivan: Those were added for the preconvolution option, which I moved to a new ticket branch. I can move these lines to that branch as well.

python/lsst/ip/diffim/detectAndMeasure.py
```diff
@@ -0,0 +1,441 @@
...
+        self.forcedMeasurement.slots.shape = None
+
+
+class DetectAndMeasureTask(lsst.pipe.base.PipelineTask):
+    """Detect and measure sources on a difference image.
+    """
+    ConfigClass = DetectAndMeasureConfig
+    _DefaultName = "detectAndMeasure"
+
+    def __init__(self, butler=None, **kwargs):
```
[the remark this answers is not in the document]
isullivan: We don't actually need it anymore! I removed the code that required it.

python/lsst/ip/diffim/makeKernel.py
isullivan: Yes, the code below returns the input cell sizes without updating them, which is what this method would do.

python/lsst/ip/diffim/subtractImages.py
```diff
@@ -0,0 +1,529 @@
...
+from lsst.ip.diffim.utils import getPsfFwhm
+from lsst.meas.algorithms import ScaleVarianceTask
+import lsst.pex.config
+import lsst.pipe.base
+from lsst.pipe.base import connectionTypes
+
+__all__ = ["AlardLuptonSubtractConfig", "AlardLuptonSubtractTask"]
+
+_dimensions = ("instrument", "visit", "detector")
+_defaultTemplates = {"coaddName": "deep", "fakesType": ""}
```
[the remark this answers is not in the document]
isullivan: Maybe, but if so let's do that change on a different ticket.

python/lsst/ip/diffim/subtractImages.py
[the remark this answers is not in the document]
isullivan: I found a case where it is needed, and added a unit test that covers that case. It is needed if the images are not perfectly calibrated.

tests/test_subtractTask.py
isullivan: I made the test more precise as you suggested, but did not fix up the underlying error message. If we have a chance, it would be a good idea to try to clean up the code that solves for the matching kernel, including better error messages.

python/lsst/ip/diffim/subtractImages.py
```diff
@@ -0,0 +1,535 @@
...
+        doc="Warped and PSF-matched template used to create `subtractedExposure`.",
+        dimensions=("instrument", "visit", "detector"),
+        storageClass="ExposureF",
+        name="{fakesType}{coaddName}Diff_matchedExp",
+    )
+
+
+class AlardLuptonSubtractConnections(SubtractInputConnections, SubtractImageOutputConnections):
+    def __init__(self, *, config=None):
+        super().__init__(config=config)
```
parejkoj, line 82: I think you don't need this init at all. Just put `pass` under the class definition.

python/lsst/ip/diffim/subtractImages.py
```diff
@@ -0,0 +1,535 @@
...
+    )
+    decorrelate = lsst.pex.config.ConfigurableField(
+        target=lsst.ip.diffim.DecorrelateALKernelTask,
+        doc="Task to decorrelate the image difference.",
+    )
+    requiredTemplateFraction = lsst.pex.config.Field(
+        dtype=float,
+        default=0.1,
+        doc="Abort task if template covers less than this fraction of pixels."
+        "Setting to 0 will always attempt image subtraction."
```
parejkoj, line 113: Need a space after line continued sentence.

python/lsst/ip/diffim/subtractImages.py
```diff
@@ -0,0 +1,535 @@
...
+        self.makeKernel.kernel.name = "AL"
+        self.makeKernel.kernel.active.fitForBackground = self.doSubtractBackground
+        self.makeKernel.kernel.active.spatialKernelOrder = 1
+        self.makeKernel.kernel.active.spatialBgOrder = 2
+        if self.forceCompatibility:
+            self.mode = "convolveTemplate"
+
+
+class AlardLuptonSubtractTask(lsst.pipe.base.PipelineTask):
+    """Base class for image subtraction Tasks using the Alard & Lupton (1998)
```
parejkoj, line 154: This is no longer a "base class". Maybe "Difference [or Subtract? or Compute the difference between?] a science and template image using the..."

PR #219, opened 2022-05-27
DM-35006: Fix ordering of PSF dimensions

isullivan:


kherner (approved):
Looks fine, assuming psfShape[1], psfShape[0] is the right order (I don't know the details of that).

PR #220, opened 2022-06-02
DM-35068: Increase test image size to fix failures on ARM machines

isullivan:


timj (approved):
This fixes it for me.

PR #223, opened 2022-06-24
DM-34786: Fixes to the new subtract and detection tasks after running ap_verify

parejkoj:


isullivan (approved):
Thanks for catching and fixing these bugs!

PR #226, opened 2022-07-20
DM-35654: Add finalized PSF support to the new image differencing task

isullivan:
This PR copies the similar code from `pipe_tasks.ImageDifferenceTask`.

### DIFF
--- a/python/lsst/ip/diffim/subtractImages.py
+++ b/python/lsst/ip/diffim/subtractImages.py
@@ -59,6 +59,14 @@ class SubtractInputConnections(lsst.pipe.base.PipelineTaskConnections,
         storageClass="SourceCatalog",
         name="{fakesType}src"
     )
+    finalizedPsfApCorrCatalog = connectionTypes.Input(
+        doc=("Per-visit finalized psf models and aperture correction maps. "
+             "These catalogs use the detector id for the catalog id, "
+             "sorted on id for fast lookup."),
+        dimensions=("instrument", "visit"),
+        storageClass="ExposureCatalog",
+        name="finalized_psf_ap_corr_catalog",
+    )
 
 
 class SubtractImageOutputConnections(lsst.pipe.base.PipelineTaskConnections,
@@ -79,7 +87,11 @@ class SubtractImageOutputConnections(lsst.pipe.base.PipelineTaskConnections,
 
 
 class AlardLuptonSubtractConnections(SubtractInputConnections, SubtractImageOutputConnections):
-    pass
+
+    def __init__(self, *, config=None):
+        super().__init__(config=config)
+        if not config.doApplyFinalizedPsf:
+            self.inputs.remove("finalizedPsfApCorrCatalog")
 
 
 class AlardLuptonSubtractConfig(lsst.pipe.base.PipelineTaskConfig,
@@ -126,6 +138,12 @@ class AlardLuptonSubtractConfig(lsst.pipe.base.PipelineTaskConfig,
         dtype=bool,
         default=True,
     )
+    doApplyFinalizedPsf = lsst.pex.config.Field(
+        doc="Replace science Exposure's psf and aperture correction map"
+        " with those in finalizedPsfApCorrCatalog.",
+        dtype=bool,
+        default=False,
+    )
 
     forceCompatibility = lsst.pex.config.Field(
         dtype=bool,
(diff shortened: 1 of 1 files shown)


python/lsst/ip/diffim/subtractImages.py
```diff
@@ -180,6 +237,8 @@ def run(self, template, science, sources):
             Identified sources on the science exposure. This catalog is used to
             select sources in order to perform the AL PSF matching on stamp
             images around them.
+        **kwargs
+            Any additional keyword parameters.
```
parejkoj, line 241: What is `**kwargs` for here? We don't use any other args in this `run()` method and we don't call `super` (and won't, be cause the super for pipeTask is not implemented), so we shouldn't accept any.

python/lsst/ip/diffim/subtractImages.py
```diff
@@ -167,7 +172,59 @@ def __init__(self, **kwargs):
...
+        if self.config.doApplyFinalizedPsf:
+            finalizedPsfApCorrCatalog = inputs.get("finalizedPsfApCorrCatalog", None)
+            inputs["science"] = self.prepareCalibratedExposure(
+                inputs["science"],
+                finalizedPsfApCorrCatalog=finalizedPsfApCorrCatalog
+            )
+        subtractResults = self.run(**inputs)
+        return subtractResults
+
+    def prepareCalibratedExposure(self, exposure, finalizedPsfApCorrCatalog=None):
```
parejkoj, line 191: It's not a calibrated exposure; maybe `_applyExternalCalibrations` (since this seems like a private-ish method)?

python/lsst/ip/diffim/subtractImages.py
```diff
@@ -167,7 +172,59 @@ def __init__(self, **kwargs):
...
+        """Prepare a calibrated exposure and apply finalized psf if so configured.
```
parejkoj, line 192: Maybe "Apply replace calibrations (psf, and ApCorrMap) on this exposure with external ones."?

python/lsst/ip/diffim/subtractImages.py
```diff
@@ -167,7 +172,59 @@ def __init__(self, **kwargs):
...
+        -------
+        exposure : `lsst.afw.image.exposure.Exposure`
+            Exposure with adjusted calibrations.
+        """
+        detectorId = exposure.getInfo().getDetector().getId()
+
+        if finalizedPsfApCorrCatalog is not None:
+            row = finalizedPsfApCorrCatalog.find(detectorId)
+            if row is None:
+                self.log.warning("Detector id %s not found in finalizedPsfApCorrCatalog; "
```
parejkoj, line 213: Should this be a `warning`, `error`, or `raise`?

python/lsst/ip/diffim/subtractImages.py
```diff
@@ -167,7 +172,59 @@ def __init__(self, **kwargs):
...
+                                 "Using original psf.", detectorId)
+            else:
+                psf = row.getPsf()
+                apCorrMap = row.getApCorrMap()
+                if psf is None or apCorrMap is None:
+                    self.log.warning("Detector id %s has None for psf/apCorrMap in "
```
parejkoj, line 219: Should this be a `warning`, `error`, or `raise`?

python/lsst/ip/diffim/subtractImages.py
```diff
@@ -167,7 +172,59 @@ def __init__(self, **kwargs):
...
+            the detector id for the catalog id, sorted on id for fast lookup.
+
+        Returns
+        -------
+        exposure : `lsst.afw.image.exposure.Exposure`
+            Exposure with adjusted calibrations.
+        """
+        detectorId = exposure.getInfo().getDetector().getId()
+
+        if finalizedPsfApCorrCatalog is not None:
```
parejkoj, line 210: There's no `else` for this, so if that kwarg is `None`, we'll never know.

python/lsst/ip/diffim/subtractImages.py
```diff
@@ -167,7 +172,59 @@ def __init__(self, **kwargs):
...
+            Exposure catalog with finalized psf models and aperture correction
+            maps to be applied if config.doApplyFinalizedPsf=True.  Catalog uses
+            the detector id for the catalog id, sorted on id for fast lookup.
+
+        Returns
+        -------
+        exposure : `lsst.afw.image.exposure.Exposure`
+            Exposure with adjusted calibrations.
+        """
+        detectorId = exposure.getInfo().getDetector().getId()
```
parejkoj, line 208: `getInfo()` -> `info` (no python properties for the others yet...)

parejkoj (changes requested):
As I said when I initially requested this be removed: it needs some tests that show it works as advertised. As written, it won't because the dataset is not in the Connections. If we don't have a sample ExposureCatalog with a psf and apCorrMap handy, I can probably cook one up.

python/lsst/ip/diffim/subtractImages.py
```diff
@@ -167,7 +172,59 @@ def __init__(self, **kwargs):
...
-    def run(self, template, science, sources):
+    @lsst.utils.inheritDoc(lsst.pipe.base.PipelineTask)
+    def runQuantum(self, butlerQC: lsst.pipe.base.ButlerQuantumContext,
+                   inputRefs: lsst.pipe.base.InputQuantizedConnection,
+                   outputRefs: lsst.pipe.base.OutputQuantizedConnection):
+        inputs = butlerQC.get(inputRefs)
+        self.log.info("Processing %s", butlerQC.quantum.dataId)
+
+        if self.config.doApplyFinalizedPsf:
+            finalizedPsfApCorrCatalog = inputs.get("finalizedPsfApCorrCatalog", None)
```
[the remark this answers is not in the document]
isullivan: That was an oversight that it was left out of the connections. Thanks for catching that!

python/lsst/ip/diffim/subtractImages.py
isullivan: Removed -  the quantum graph should not even build if the config is set that would call this method and the finalizedPsfApCorrCatalog is not present.

python/lsst/ip/diffim/subtractImages.py
isullivan: I'm going to leave these as `warning`s for now.

python/lsst/ip/diffim/subtractImages.py
```diff
@@ -126,6 +126,11 @@ class AlardLuptonSubtractConfig(lsst.pipe.base.PipelineTaskConfig,
         dtype=bool,
         default=True,
     )
+    doApplyFinalizedPsf = lsst.pex.config.Field(
+        doc="Whether to apply finalized psf models and aperture correction map.",
```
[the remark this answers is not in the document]
isullivan: I'm open to changing the name, though I think at least the external wcs will need special treatment to ensure that it is applied the same in this Task and `GetTemplateTask`. That inclines me to include each set of external calibrations that comes from a different input connection as its own config option.

python/lsst/ip/diffim/subtractImages.py
```diff
@@ -167,7 +185,42 @@ def __init__(self, **kwargs):
...
+
+        row = finalizedPsfApCorrCatalog.find(detectorId)
+        if row is None:
+            self.log.warning("Detector id %s not found in finalizedPsfApCorrCatalog; "
+                             "Using original psf.", detectorId)
+        else:
+            psf = row.getPsf()
+            apCorrMap = row.getApCorrMap()
+            if psf is None or apCorrMap is None:
+                self.log.warning("Detector id %s has None for psf/apCorrMap in "
```
parejkoj, line 215: I'd rather this log message be more explicit about which of those are None, to help with debugging. Either two separate `if` statements, or some logic in the log message, something like '"psf=None" if psf is None else ""`.

python/lsst/ip/diffim/subtractImages.py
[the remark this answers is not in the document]
parejkoj: Ok, that's reasonable. I mostly just wanted us to make sure we thought it through. makeWarp and forcedPhot both use a separate config option for each, so this would look the same (though I question the `FinalizedPsf` vs. `ExternalWcs` naming choices). I do want to ask DRP whether that's what they want the final API to look: I'd be surprised if there's a case where you want to use one but not the other.

parejkoj (approved):
Thanks for the cleanups: I think this is clearer now. A conversation with DRP about some of these naming questions is still in order though.

I do still want this to have tests, but I understand the timing here; can you file a ticket to add tests of this logic and note the ticket number on this PR?

isullivan:
DM-35694 will add unit tests of the "finalized psf/apCorr" option.

PR #228, opened 2022-07-29
DM-35674: Remove Gen2 code

timj:


isullivan (approved):
These changes all look good. There are a few more places in the run method of `GetCoaddAsTemplateTask` where comments about Gen2-only behavior could be removed, but since that class will be removed entirely in the (near) future I don't think it's essential to remove them now (e.g. lines 199 and 288).

timj:
I fixed line 199. I wasn't really sure what line 288 was saying to me. Did it mean that those few lines should be deleted because gen3 will never have a problem, or should the comment itself be removed. Since the class is going away I left it in.

PR #229, opened 2022-08-05
DM-35817: Turn off compatibility mode as the default.

isullivan:
Also set the default convolution mode to `convolveTemplate`

### DIFF
--- a/python/lsst/ip/diffim/subtractImages.py
+++ b/python/lsst/ip/diffim/subtractImages.py
@@ -99,7 +99,7 @@ class AlardLuptonSubtractConfig(lsst.pipe.base.PipelineTaskConfig,
                                 pipelineConnections=AlardLuptonSubtractConnections):
     mode = lsst.pex.config.ChoiceField(
         dtype=str,
-        default="auto",
+        default="convolveTemplate",
         allowed={"auto": "Choose which image to convolve at runtime.",
                  "convolveScience": "Only convolve the science image.",
                  "convolveTemplate": "Only convolve the template image."},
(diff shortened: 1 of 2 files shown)


tests/test_subtractTask.py
```diff
@@ -155,6 +155,9 @@ def test_config_validate_forceCompatibility(self):
         config = subtractImages.AlardLuptonSubtractTask.ConfigClass()
         config.mode = "auto"
         config.forceCompatibility = True
+        with self.assertRaises(ValueError):
+            config.validate()
+        config.mode = 'convolveTemplate'
```
parejkoj, line 160: Here's what I'd done (after the `forceCompatibility=True` line), and removing the two lines below:

```
        # Ensure we're not trying to change config values inside validate().
        config.freeze()
        with self.assertRaisesRegex(FieldValidationError,
                                    "forceCompatibility=True requires mode='convolveTemplate'"):
            config.validate()

        config = subtractImages.AlardLuptonSubtractTask.ConfigClass()
        config.mode = "convolveTemplate"
        config.forceCompatibility = True
        # Ensure we're not trying to change config values inside validate().
        config.freeze()
        # Should not raise:
        config.validate()
```

parejkoj (changes requested):
A few suggestions, since I'd just been working on this for DM-971.

Please reword the commit message to say why you defaulted `mode="convolveTemplate"`: it's not obvious at first glance why that's related to here (I'm assuming because that's our transition plan).

python/lsst/ip/diffim/subtractImages.py
```diff
@@ -162,8 +162,10 @@ def setDefaults(self):
         self.makeKernel.kernel.active.spatialBgOrder = 2
 
     def validate(self):
-        if self.forceCompatibility:
-            self.mode = "convolveTemplate"
+        if self.forceCompatibility and not (self.mode == "convolveTemplate"):
+            msg = "forceCompatibility=True requires mode='convolveTemplate'"
```
parejkoj, line 166: Oh, I forgot to include one thing in my suggestion: ` but mode was '{self.mode}'"` at the end of this message (and make it an f-string).

parejkoj (approved):
One tweak to the exception text, otherwise this is good. Thanks for fixing this on this ticket, so I don't have to on DM-971.

PR #233, opened 2022-09-16
DM-30869: Modernize MetricTask for better Gen 3 workflow

kfindeisen:
This PR removes all code for handling `None` inputs from concrete `MetricTasks`.

### DIFF
--- a/python/lsst/ip/diffim/metrics.py
+++ b/python/lsst/ip/diffim/metrics.py
@@ -68,12 +68,12 @@ class NumberSciSourcesMetricTask(MetricTask):
     ConfigClass = NumberSciSourcesMetricConfig
 
     def run(self, sources):
-        """Count the number of non-primary science sources.
+        """Count the number of primary science sources.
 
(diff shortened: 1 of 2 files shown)


python/lsst/ip/diffim/metrics.py
```diff
@@ -68,12 +68,12 @@ class NumberSciSourcesMetricTask(MetricTask):
...
 
     def run(self, sources):
-        """Count the number of non-primary science sources.
+        """Count the number of primary science sources.
 
         Parameters
         ----------
-        sources : `lsst.afw.table.SourceCatalog` or `None`
-            A science source catalog, which may be empty or `None`.
+        sources : `lsst.afw.table.SourceCatalog`.
```
TallJimbo, line 75:
```suggestion
        sources : `lsst.afw.table.SourceCatalog`
```
I don't think we usually put punctuation here.

TallJimbo (approved):

PR #235, opened 2022-10-26
DM-36717: Diffim bug fixes

isullivan:
* Uses a more accurate PSF FWHM for constructing the kernel basis list
* Applies the template PhotoCalib before constructing the PSF matching kernel
* Rejects flagged sources and sky sources from source list before selecting kernel candidates.

### DIFF
--- a/python/lsst/ip/diffim/subtractImages.py
+++ b/python/lsst/ip/diffim/subtractImages.py
@@ -276,6 +276,8 @@ class AlardLuptonSubtractTask(lsst.pipe.base.PipelineTask):
         ------
         RuntimeError
             If an unsupported convolution mode is supplied.
+        RuntimeError
+            If there are too few sources to calculate the PSF matching kernel.
         lsst.pipe.base.NoWorkFound
             Raised if fraction of good pixels, defined as not having NO_DATA
             set, is less then the configured requiredTemplateFraction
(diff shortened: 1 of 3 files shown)


python/lsst/ip/diffim/utils.py
```diff
@@ -1080,20 +1080,31 @@ def detectDipoleSources(self, doMerge=True, diffim=None, detectSigma=5.5, grow=3
...
     Parameters
     ----------
     psf : `lsst.afw.detection.Psf`
         Point spread function (PSF) to evaluate.
+    average : `bool`, optional
+        Set to return the average width.
 
     Returns
     -------
     psfSize : `float`
```
kherner, line 1096: Does the returns description need to be changed since you could also be returning what appears to be a tuple when average = False?

python/lsst/ip/diffim/subtractImages.py
```diff
@@ -297,6 +313,10 @@ def run(self, template, science, sources, finalizedPsfApCorrCatalog=None):
             convolveTemplate = False
         else:
             raise RuntimeError("Cannot handle AlardLuptonSubtract mode: %s", self.config.mode)
+        # put the template on the same photometric scale as the science image
+        photoCalib = template.getPhotoCalib()
+        self.log.info("Applying photometric calibration to template: %f", photoCalib.getCalibrationMean())
+        template.maskedImage = photoCalib.calibrateImage(template.maskedImage)
```
kherner, line 319: Where is the science image involved here?

python/lsst/ip/diffim/subtractImages.py
```diff
@@ -309,13 +329,12 @@ def run(self, template, science, sources, finalizedPsfApCorrCatalog=None):
             self.log.info("Science variance scaling factor: %.2f", sciVarFactor)
             self.metadata.add("scaleScienceVarianceFactor", sciVarFactor)
 
-        kernelSources = self.makeKernel.selectKernelSources(template, science,
-                                                            candidateList=sources,
-                                                            preconvolved=False)
+        selectSources = self._sourceSelector(sources)
+        self.log.info("%i sources used out of %i from the input catalog", len(selectSources), len(sources))
```
kherner, line 333: Would it make sense to do some alerting/error raising if zero sources are selected? Might save time later...

kherner (approved):
See comments, especially the one about setting the photometric calib scale to the science image.

python/lsst/ip/diffim/subtractImages.py
isullivan: It's not, the template is read in with units of counts, so this applies the template's calibration to put it in nJy. That should make it comparable to the science image, which should make work slightly easier for calculating the matching kernel.

PR #236, opened 2022-11-01
DM-36265: Fix subtractImages failures when compatibility mode is off

isullivan:
Use only the smaller of the template and science bbox for calculating the matching kernel.

### DIFF
--- a/python/lsst/ip/diffim/makeKernel.py
+++ b/python/lsst/ip/diffim/makeKernel.py
@@ -348,15 +348,21 @@ class MakeKernelTask(PsfMatchTask):
         -------
         kernelCellSet : `lsst.afw.math.SpatialCellSet`
             a SpatialCellSet for use with self._solve
+
+        Raises
+        ------
+        RuntimeError
+            If no `candidateList` is supplied.
         """
         if not candidateList:
             raise RuntimeError("Candidate list must be populated by makeCandidateList")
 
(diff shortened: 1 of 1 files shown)


python/lsst/ip/diffim/makeKernel.py
```diff
@@ -330,7 +332,7 @@ def makeKernelBasisList(self, targetFwhmPix=None, referenceFwhmPix=None,
 
         return basisList
 
-    def _buildCellSet(self, templateMaskedImage, scienceMaskedImage, candidateList):
+    def _buildCellSet(self, templateMaskedImage, scienceMaskedImage, candidateList, imageBBox):
```
yalsayyad, line 335: This seems like an unnecessary interface change to me:
1)  You don't use your `bbox` variable again in code that calls it after line 123. 
2) `imageBBox` can be exactly determined from the other two inputs already there:  `templateMaskedImage` and `scienceMaskedImage`.

Recommendation would be, right before constructing the SpatialCellSet in 
`kernelCellSet = lsst.afw.math.SpatialCellSet(imageBBox, sizeCellX, sizeCellY)` below

you could:
```
        imageBBox = templateMaskedImage.getBBox()
        imageBBox.clip(scienceMaskedImage.getBBox()
``` 
and thats it.

yalsayyad (approved):

PR #239, opened 2022-12-08
DM-37283: Remove compatibility mode from subtractImages

isullivan:
This option was scheduled to be removed after release 25, which will be based on weekly 48.


bsmartradio (approved):
Everything looks good. I checked that there wasn't any missed pieces of code, and tried sconsing it. There were no complaints, and all the warnings of forcedCompatibility were removed.

PR #242, opened 2023-01-03
DM-35962: Remove examples/ from all packages

parejkoj:


isullivan (approved):
I am concerned that we appear to be losing a useful example with the removal of `maskedKernel.cc`. The removed notes from `KernelSolution.cc` suggest that this is for getting the indexing right, and since this code has not changed in a long time I suspect it is still relevant. That code would be more useful if it could be translated into a unit test, though. Could you file a ticket to create a replacement unit test?

parejkoj:
Created [DM-37514](https://jira.lsstcorp.org/browse/DM-37514) to make a unittest out of the removed file.

PR #246, opened 2023-01-20
DM-32756: Fix template CoaddPsf errors

arunkannawadi:


python/lsst/ip/diffim/utils.py
```diff
@@ -1080,36 +1084,115 @@ def detectDipoleSources(self, doMerge=True, diffim=None, detectSigma=5.5, grow=3
...
+                                    ):
+        pos = geom.Point2D(x, y)
+        try:
+            fwhm = getPsfFwhm(psf, average=average, position=pos)
+            width.append(fwhm)
+        except Exception:
+            pass
+
+    if not width:
+        return np.nan if average else (np.nan, np.nan)
```
arunkannawadi, line 1193: I'm open to suggestions on whether we should return `NaN`s  here or raise an exception, as it would have been the case before. I don't know if the downstream processes are equipped to handle `NaN`s.

python/lsst/ip/diffim/psfMatch.py
```diff
@@ -366,6 +366,17 @@ class PsfMatchConfig(pexConfig.Config):
...
+        doc="Grid size to compute the average FWHM in an exposure",
+        dtype=int,
+        optional=True,
+    )
+    fwhmExposureBuffer = pexConfig.Field(
+        doc="Fractional buffer margin to be left out of all sides of the image during construction"
+            "of grid to compute average FWHM in an exposure",
+        dtype=float,
+        optional=True
+    )
```
arunkannawadi, line 379: I added these config options in the base class here. But I now realize these should probably be just in the task that subtract images.

python/lsst/ip/diffim/utils.py
erykoff: If this raises an exception, it needs to be caught by the caller.  This should be much more robust to failures than before, but I really don't want to be building in more crashing opportunities.

python/lsst/ip/diffim/utils.py
```diff
@@ -1080,36 +1084,115 @@ def detectDipoleSources(self, doMerge=True, diffim=None, detectSigma=5.5, grow=3
...
+    psf = exposure.psf
+
+    if fwhmExposureBuffer is None or fwhmExposureGrid is None:
+        try:
+            fwhm = getPsfFwhm(psf, average=average)
+        except InvalidParameterError as err:
+            _LOG.error("Unable to compute PSF FWHM at average position. Set fwhmExposureBuffer and "
+                       "fwhmExposureGrid parameters to avoid this error.")
+            raise err
+        return fwhm
```
arunkannawadi, line 1171: Another behavior related question: Should this by default try to do `getPsfFwhm` and try to fall back to the grid evaluation only if `getPsfFwhm` fails at the average position>

python/lsst/ip/diffim/subtractImages.py
```diff
@@ -273,12 +273,25 @@ def run(self, template, science, sources, finalizedPsfApCorrCatalog=None):
                                             finalizedPsfApCorrCatalog=finalizedPsfApCorrCatalog)
         checkTemplateIsSufficient(template, self.log,
                                   requiredTemplateFraction=self.config.requiredTemplateFraction)
-        sciencePsfSize = getPsfFwhm(science.psf)
-        templatePsfSize = getPsfFwhm(template.psf)
+
+        fwhmExposureBuffer = self.config.makeKernel.kernel.active.fwhmExposureBuffer
+        fwhmExposureGrid = self.config.makeKernel.kernel.active.fwhmExposureGrid
```
isullivan, line 278: Just `self.config.makeKernel.fwhmExposureBuffer` if you move the parameter definition.

python/lsst/ip/diffim/subtractImages.py
```diff
@@ -642,7 +655,7 @@ def _subtractImages(science, template, backgroundModel=None):
     return difference
 
 
-def _shapeTest(psf1, psf2):
+def _shapeTest(exp1, exp2, **kwargs):
```
isullivan, line 658: Use `evaluateMeanPsfFwhm` and `fwhmExposureGrid` as keywords here, not hidden in `**kwargs`!

python/lsst/ip/diffim/subtractImages.py
```diff
@@ -657,8 +670,8 @@ def _shapeTest(psf1, psf2):
     `bool`
         Returns True if psf1 is narrower than psf2 in either dimension.
     """
-    shape1 = getPsfFwhm(psf1, average=False)
-    shape2 = getPsfFwhm(psf2, average=False)
+    shape1 = evaluateMeanPsfFwhm(exp1, average=False, **kwargs)
+    shape2 = evaluateMeanPsfFwhm(exp2, average=False, **kwargs)
```
isullivan, line 674: Wrap the call to `getPsfFwhm` in a `try except` block, and use `evaluateMeanPsfFwhm` if there is an error.

python/lsst/ip/diffim/utils.py
```diff
@@ -1080,36 +1084,115 @@ def detectDipoleSources(self, doMerge=True, diffim=None, detectSigma=5.5, grow=3
...
+    if position is None:
+        position = psf.getAveragePosition()
+    image = psf.computeKernelImage(position).array
+    peak = psf.computePeak(position)
     peakLocs = np.unravel_index(np.argmax(image), image.shape)
+    width = _sliceWidth(image, peak/2., peakLocs, axis=0), _sliceWidth(image, peak/2., peakLocs, axis=1)
+    return np.nanmean(width) if average else width
+
+
+def evaluateMeanPsfFwhm(exposure, average=True, *, fwhmExposureBuffer=None, fwhmExposureGrid=None):
```
isullivan, line 1128: Remove the `average` keyword. Just have this always return the average.

tests/test_subtractTask.py
```diff
@@ -82,6 +93,59 @@ def test_equal_images(self):
...
+        customPsf = CustomCoaddPsf(exposureCatalog, template.wcs)
+
+        config = subtractImages.AlardLuptonSubtractTask.ConfigClass()
+        config.doSubtractBackground = False
+        task1 = subtractImages.AlardLuptonSubtractTask(config=config)
+
+        config = subtractImages.AlardLuptonSubtractTask.ConfigClass()
+        config.doSubtractBackground = False
+        config.makeKernel.kernel.active.fwhmExposureBuffer = None
+        config.makeKernel.kernel.active.fwhmExposureGrid = None
```
isullivan, line 126: Will need to be `config.makeKernel.fwhmExposureBuffer` if the parameters get moved.

python/lsst/ip/diffim/utils.py
isullivan: I would remove the option for `fwhmExposureBuffer` and `fwhmExposureGrid` to be `None`, and the wrapper around `getPsfFwhm`. The way the defaults are set, these parameters can't be `None`, anyway, and I think it is cleaner to explicitly call `getPsfFwhm` or `evaluateMeanPsfFwhm`, since the values will be different.

python/lsst/ip/diffim/utils.py
isullivan: Remove the support for the `average` keyword. This should raise an exception rather than returning `NaN`, since downstream code can't handle `NaN`s.

isullivan (changes requested):
Two major changes requested:
1) Only use the new `evaluateMeanPsfFwhm` in `makeKernel.py` and `subtractImages.py` (and the unit test), and move the new config parameters to `MakeKernelConfig`. It should never be needed in the `*PsfMatch` tasks since those work on science images, and I am trying to work towards deprecating them.
2) Remove the fallback to `getPsfFwhm` from `evaluateMeanPsfFwhm`, and eliminate the `average` keyword and the support for `fwhmExposureBuffer` and `fwhmExposureGrid` to be `None`. The way you have everything set up, those values should always be set.

python/lsst/ip/diffim/subtractImages.py
isullivan: Then it should just test whether psfFwhm1 < psfFwhm2

python/lsst/ip/diffim/utils.py
isullivan: I think a `ValueError` would be appropriate if the final result would be `NaN`

python/lsst/ip/diffim/modelPsfMatch.py
```diff
@@ -327,11 +327,11 @@ def run(self, exposure, referencePsfModel, kernelSum=1.0):
...
-        # TODO: This should be evaluated at (or close to) the center of the
-        # exposure's bounding box in DM-32756.
-        sciAvgPos = exposure.getPsf().getAveragePosition()
+        fwhmScience = diutils.evaluateMeanPsfFwhm(exposure,
+                                                  fwhmExposureBuffer=self.kConfig.fwhmExposureBuffer,
+                                                  fwhmExposureGrid=self.kConfig.fwhmExposureGrid
+                                                  )
         modelAvgPos = referencePsfModel.getAveragePosition()
-        fwhmScience = exposure.getPsf().computeShape(sciAvgPos).getDeterminantRadius()*sigma2fwhm
         fwhmModel = referencePsfModel.computeShape(modelAvgPos).getDeterminantRadius()*sigma2fwhm
```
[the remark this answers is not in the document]
erykoff: Isn't the problem that the average position may be a position that doesn't evaluate to a finite psf?

python/lsst/ip/diffim/subtractImages.py
erykoff: So you're saying that the evaluateMean should be the fallback and not a replacement, maybe because of performance?

python/lsst/ip/diffim/modelPsfMatch.py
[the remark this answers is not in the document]
isullivan: I believe the problem is for coadds, which might have gaps, rather than for science images. This code is also not used in production.

python/lsst/ip/diffim/subtractImages.py
isullivan: For this one check of the fwhm, yes.

python/lsst/ip/diffim/modelPsfMatch.py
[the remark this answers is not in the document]
erykoff: This code isn't used in production?  I thought that this PR originally was inspired by crashes when running by HSC people, but maybe that's not AP production, but people are using it.  
As for the other issue, science images may also have gaps (one of HSC detectors has a bad amp in the middle for example).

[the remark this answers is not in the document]
isullivan: Sorry, I meant that `modelPsfMatch` (and `imagePsfMatch`) is not used in production. The rest of the code being changed on this PR is, and this PR is correctly accounting for potential gaps in science images in that code.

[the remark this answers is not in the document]
arunkannawadi: Given that `evaluateMeanPsfFwhm` basically tries `getPsfFwhm` first anyway, isn't it safer to use that instead of using `getPsfFwhm` explicitly?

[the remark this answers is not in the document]
isullivan: I was requesting elsewhere in this PR to remove `getPsfFwhm` from `evaluateMeanPsfFwhm`, since the calculations are very different and should not be considered interchangeable in general.

python/lsst/ip/diffim/imagePsfMatch.py
```diff
@@ -416,14 +408,22 @@ def matchExposures(self, templateExposure, scienceExposure,
...
         if scienceFwhmPix is None:
             if not scienceExposure.hasPsf():
                 self.log.warning("No estimate of Psf FWHM for science image")
             else:
-                scienceFwhmPix = self.getFwhmPix(scienceExposure.getPsf())
+                scienceFwhmPix = diffimUtils.evaluateMeanPsfFwhm(
+                    scienceExposure,
+                    fwhmExposureBuffer=self.kConfig.fwhmExposureBuffer,
+                    fwhmExposureGrid=self.kConfig.fwhmExposureGrid
+                )
```
arunkannawadi, line 426: Should these calls to `evaluateMeanPsfFwhm` also be changed then? And if so, should I keep the `getFwhmPix` that I removed in the PR or switch to using `getPsfFwhm`? It looks like there are two ways of doing very similar things and I don't know if that's intentional.

arunkannawadi (commented):
I'm implementing Ian's suggestions now, but I feel like this could be a significant performance hit if we remove the wrapper around `getPsfFwhm`. With the current default values, this would mean evaluating PSF images 100 times on the template, as opposed to 1 to handle edge cases that are (hopefully) rare. AP team should perhaps monitor if there's a significant increase in time if we resort to always evaluating PSF size on a 10x10 grid.

tests/test_subtractTask.py
arunkannawadi: The way we have defined the `CustomCoaddPsf`, we cannot test that the task fails raising an `InvalidParameterError` since we are removing the support for these parameters to be `None`. The unit test will instead test explicitly that `evaluateMeanPsfFwhm` runs while `getPsfFwhm` throws an error.

python/lsst/ip/diffim/subtractImages.py
```diff
@@ -273,12 +274,25 @@ def run(self, template, science, sources, finalizedPsfApCorrCatalog=None):
...
+        fwhmExposureBuffer = self.config.makeKernel.fwhmExposureBuffer
+        fwhmExposureGrid = self.config.makeKernel.fwhmExposureGrid
+        sciencePsfSize = evaluateMeanPsfFwhm(science,
+                                             fwhmExposureBuffer=fwhmExposureBuffer,
+                                             fwhmExposureGrid=fwhmExposureGrid
+                                             )
+        templatePsfSize = evaluateMeanPsfFwhm(template,
+                                              fwhmExposureBuffer=fwhmExposureBuffer,
+                                              fwhmExposureGrid=fwhmExposureGrid
+                                              )
```
arunkannawadi, line 287: We should perhaps have this in a try/except block as well. We try `getPsfFwhm` on the template first, and if that fails, call `evaluateMeanPsfFwhm`. For consistency, both the sizes should be computed from the same function.

python/lsst/ip/diffim/makeKernel.py
```diff
@@ -58,6 +58,17 @@ class MakeKernelConfig(PsfMatchConfig):
...
+        doc="Grid size to compute the average FWHM in an exposure",
+        dtype=int,
+        default=10,
+    )
+    fwhmExposureBuffer = lsst.pex.config.Field(
+        doc="Fractional buffer margin to be left out of all sides of the image during construction"
+            "of grid to compute average FWHM in an exposure",
+        dtype=float,
+        default=0.05,
+    )
```
arunkannawadi, line 71: Are these fieldnames helpful? They are not immediately apparent. We should at least have 'PSF' in their names.

arunkannawadi (commented):
@isullivan I think I have addressed all your comments now. I just have minor suggestions that I can incorporate if you think they are worth it.

python/lsst/ip/diffim/imagePsfMatch.py
```diff
@@ -322,14 +322,6 @@ def __init__(self, *args, **kwargs):
         self.makeSubtask("selectDetection", schema=self.selectSchema)
         self.makeSubtask("selectMeasurement", schema=self.selectSchema, algMetadata=self.selectAlgMetadata)
 
-    def getFwhmPix(self, psf, position=None):
```
isullivan, line 325: It looks like one unit test fails if you remove this: `tests/test_imagePsfMatch.py::PsfMatchTestCases::testSubtractExposures`
(or could be fixed by replacing instances of `self.getFwhmPix` in `imagePsfMatch.py` with `pDiffim.utils.getPsfFwhm`)

python/lsst/ip/diffim/makeKernel.py
isullivan: I think it is sufficiently clear if you expand the docstring to be `...average PSF FWHM...`

python/lsst/ip/diffim/subtractImages.py
isullivan: That sounds good to me. In that case, please add a log message in the except block stating that the initial PSF FWHM calculation failed.

python/lsst/ip/diffim/subtractImages.py
```diff
@@ -642,23 +656,42 @@ def _subtractImages(science, template, backgroundModel=None):
...
+    except InvalidParameterError:
+        shape1 = evaluateMeanPsfFwhm(exp1,
+                                     fwhmExposureBuffer=fwhmExposureBuffer,
+                                     fwhmExposureGrid=fwhmExposureGrid
+                                     )
+        shape2 = evaluateMeanPsfFwhm(exp2,
+                                     fwhmExposureBuffer=fwhmExposureBuffer,
+                                     fwhmExposureGrid=fwhmExposureGrid
+                                     )
+        return shape1 < shape2
```
isullivan, line 692: Minor point, but this should probably be `<=`. `xTest` and `yTest` below should probably use `<=` as well, which was a mistake in the original implementation.

isullivan (approved):
Looks good! I do get one failing test due to the removal of `getFwhmPsf` from `imagePsfMatch`, but I think that can be fixed easily.

python/lsst/ip/diffim/subtractImages.py
arunkannawadi: Done. And I've updated the docstring also accordingly.

tests/test_subtractTask.py
arunkannawadi: Now that we have a log message before `evaluateMeanPsfFwhm` is called after `getPsfFwhm` fails, the unit tests look for the log message. This essentially replaces the earlier version of the test by checking that the code path follows the lines in the `except` block.

python/lsst/ip/diffim/imagePsfMatch.py
arunkannawadi: This didn't show up on Github search. I've fixed this now.

python/lsst/ip/diffim/subtractImages.py
```diff
@@ -273,12 +274,34 @@ def run(self, template, science, sources, finalizedPsfApCorrCatalog=None):
...
+                          "Evaluting PSF on a grid of points."
+                          )
+            templatePsfSize = evaluateMeanPsfFwhm(template,
+                                                  fwhmExposureBuffer=fwhmExposureBuffer,
+                                                  fwhmExposureGrid=fwhmExposureGrid
+                                                  )
+            sciencePsfSize = evaluateMeanPsfFwhm(science,
+                                                 fwhmExposureBuffer=fwhmExposureBuffer,
+                                                 fwhmExposureGrid=fwhmExposureGrid
+                                                 )
```
erykoff, line 296: I think this should be split into two try/except blocks, because it's more likely that only one should fail.  Unless there's a reason that if one fails they both should be recomputed that I'm missing?

python/lsst/ip/diffim/subtractImages.py
```diff
@@ -642,23 +665,43 @@ def _subtractImages(science, template, backgroundModel=None):
...
+        shape2 = getPsfFwhm(exp2.psf, average=False)
+    except InvalidParameterError:
+        shape1 = evaluateMeanPsfFwhm(exp1,
+                                     fwhmExposureBuffer=fwhmExposureBuffer,
+                                     fwhmExposureGrid=fwhmExposureGrid
+                                     )
+        shape2 = evaluateMeanPsfFwhm(exp2,
+                                     fwhmExposureBuffer=fwhmExposureBuffer,
+                                     fwhmExposureGrid=fwhmExposureGrid
+                                     )
```
erykoff, line 701: Same here, I'm not sure why these are not in their own try/except blocks.

python/lsst/ip/diffim/subtractImages.py
erykoff: Also, what happens if the `evaluateMeanPsfFwhm` fails? We get a `ValueError` and is that caught anywhere?

erykoff (commented):
I have a few further comments now that things have settled down.

python/lsst/ip/diffim/subtractImages.py
arunkannawadi: The expectation, based on what Ian had said, is that only template could have holes and not the science images. The error specifically comes from `CoaddPsf` and I believe the PSF of science images are _not_ `CoaddPsf`s. So the failures come from template, which is what we try to do first.

They are not in two try/except blocks because we'd want to have the same definition of PSF size for both the images. Having two different blocks would mean their sizes can get defined differently.

arunkannawadi: My understanding is if we get a `ValueError`, it should indeed crash the pipeline (?). This would only happen if our grid happens to exactly line up with chip gaps or dithering strategy.

erykoff: Can you make this comment in the code for why these are in the same try/except block?

erykoff (approved):
Please add the requested comment so this is clear in the future, and you're good to go from my side.

python/lsst/ip/diffim/subtractImages.py
arunkannawadi: And it is done!

python/lsst/ip/diffim/makeKernel.py
```diff
@@ -119,8 +130,14 @@ def run(self, template, science, kernelSources, preconvolved=False):
...
-        templateFwhmPix = getPsfFwhm(template.psf)
-        scienceFwhmPix = getPsfFwhm(science.psf)
+        templateFwhmPix = evaluateMeanPsfFwhm(template,
+                                              fwhmExposureBuffer=self.config.fwhmExposureBuffer,
+                                              fwhmExposureGrid=self.config.fwhmExposureGrid
+                                              )
+        scienceFwhmPix = evaluateMeanPsfFwhm(science,
+                                             fwhmExposureBuffer=self.config.fwhmExposureBuffer,
+                                             fwhmExposureGrid=self.config.fwhmExposureGrid
+                                             )
```
arunkannawadi, line 140: I don't know how I missed this. @yalsayyad reports an increase of 4 seconds in `diffim` and I believe this is all due to these changes that always evaluate the mean FWHM over a 10x10 grid.

PR #248, opened 2023-02-04
DM-37770: use pybind wrappers, merge into one pybind shared lib

mwittgen:


python/lsst/ip/diffim/diffimLib.py
```diff
@@ -24,11 +24,4 @@
...
 
-from .basisLists import *
-from ._dipoleAlgorithms import *
-from .findSetBits import *
-from .imageStatistics import *
-from .imageSubtract import *
-from .kernelCandidate import *
-from .kernelCandidateDetection import *
-from .kernelSolution import *
+from ._ipdiffimLib import *
```
TallJimbo, line 27:
```suggestion
from ._ipdiffimLib import *

```

python/lsst/ip/diffim/detail/__init__.py
```diff
@@ -19,4 +19,4 @@
 # the GNU General Public License along with this program.  If not,
 # see <https://www.lsstcorp.org/LegalNotices/>.
 #
-from .diffimDetailLib import *
+from .._ipdiffimLib import *
```
TallJimbo, line 22: I hate the fact that we're exposing things in the `detail` namespace to Python at all, but it's preexisting (and probably so tests could be written in Python), so I suppose this isn't the right ticket to do anything about that.

TallJimbo (approved):

PR #249, opened 2023-02-07
DM-37884: Include all evaluateMeanPsfFwhm in try/except blocks

arunkannawadi:
This PR introduces the fix that should have been in place in #246 . All evaluateMeanPsfFwhm calls must be within an `except` block and should be called only after `getPsfFwhm` has been tried first.

### DIFF
--- a/python/lsst/ip/diffim/makeKernel.py
+++ b/python/lsst/ip/diffim/makeKernel.py
@@ -30,6 +30,7 @@ import lsst.afw.table
 import lsst.daf.base
 from lsst.meas.algorithms import SourceDetectionTask, SubtractBackgroundTask
 from lsst.meas.base import SingleFrameMeasurementTask
+from lsst.pex.exceptions import InvalidParameterError
 import lsst.pex.config
 import lsst.pipe.base
 
@@ -38,7 +39,7 @@ from .psfMatch import PsfMatchConfig, PsfMatchTask, PsfMatchConfigAL, PsfMatchCo
 
 from . import diffimLib
 from . import diffimTools
-from .utils import evaluateMeanPsfFwhm
+from .utils import evaluateMeanPsfFwhm, getPsfFwhm
 
 
 class MakeKernelConfig(PsfMatchConfig):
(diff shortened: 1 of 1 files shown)


python/lsst/ip/diffim/makeKernel.py
```diff
@@ -130,14 +131,28 @@ def run(self, template, science, kernelSources, preconvolved=False):
...
+        # evaluate the PSF on a grid specified by fwhmExposure* fields.
+        # To keep consistent definitions for PSF size on the template and
+        # science images, we use the same method for both.
+        try:
+            templateFwhmPix = getPsfFwhm(template.psf)
+            scienceFwhmPix = getPsfFwhm(science.psf)
+        except InvalidParameterError:
+            self.log.info("Unable to evaluate PSF at the average position. "
+                          "Evaluting PSF on a grid of points."
+                          )
```
arunkannawadi, line 146: I don't know if logging this information in many places is helpful. For a given template with a `CoaddPsf` error, this line might end up appearing several times.

isullivan: I think you're right. This message would appear once already from `subtractImages.py`, so I would suggest downgrading the log messages here (and below) to `debug`

python/lsst/ip/diffim/makeKernel.py
```diff
@@ -171,14 +186,27 @@ def selectKernelSources(self, template, science, candidateList=None, preconvolve
...
+        # the template has no input exposures at the average position of the
+        # stars. So we try getPsfFwhm first on template, and if that fails we
+        # evaluate the PSF on a grid specified by fwhmExposure* fields.
+        # To keep consistent definitions for PSF size on the template and
+        # science images, we use the same method for both.
+        try:
+            templateFwhmPix = getPsfFwhm(template.psf)
+            scienceFwhmPix = getPsfFwhm(science.psf)
+        except InvalidParameterError:
+            self.log.info("Unable to evaluate PSF at the average position. "
```
isullivan, line 199: As above, downgrade to `debug`

isullivan (approved):
Looks good. I would suggest these logs be downgraded to `debug` to reduce the clutter.

arunkannawadi:
Thanks for the comment, Ian. I downgraded the logging level for these to be `DEBUG`.

PR #253, opened 2023-03-16
DM-38209: Fix module string in compiled module

yalsayyad:
Correct the compiled module to report the correct qualified import path.


natelust (approved):

natelust:
bah, I should have only approved your changes to the other package, but hey, I do approve of what I wrote :)

PR #256, opened 2023-04-18
DM-31924: Switch to configurable data ID packing.

TallJimbo:


python/lsst/ip/diffim/detectAndMeasure.py
```diff
@@ -242,10 +253,8 @@ def runQuantum(self, butlerQC: pipeBase.ButlerQuantumContext,
                    inputRefs: pipeBase.InputQuantizedConnection,
                    outputRefs: pipeBase.OutputQuantizedConnection):
         inputs = butlerQC.get(inputRefs)
-        expId, expBits = butlerQC.quantum.dataId.pack("visit_detector",
-                                                      returnMaxBits=True)
-        idFactory = self.makeIdFactory(expId=expId, expBits=expBits)
-
+        idGenerator = self.config.idGenerator.apply(butlerQC.quantum.dataId)
+        idFactory = idGenerator.make_table_id_factory()
```
parejkoj, line 257: Should these be moved into `run`? I've wondered about why we do it in `runQuantum` some of the time.

parejkoj: Answering my own question after looking at the pipe_tasks code: because we need the dataId, which isn't available in `run`. See my other comments in pipe_tasks, though.

PR #257, opened 2023-04-28
DM-38901: Clear the template mask plane in image differencing

isullivan:
A few minor bugs uncovered while working on this ticket are also fixed.

### DIFF
--- a/python/lsst/ip/diffim/subtractImages.py
+++ b/python/lsst/ip/diffim/subtractImages.py
@@ -153,7 +153,7 @@ class AlardLuptonSubtractBaseConfig(lsst.pex.config.Config):
     detectionThreshold = lsst.pex.config.Field(
         dtype=float,
         default=10,
-        doc="Minimum signal to noise ration of detected sources "
+        doc="Minimum signal to noise ratio of detected sources "
         "to use for calculating the PSF matching kernel."
     )
     badSourceFlags = lsst.pex.config.ListField(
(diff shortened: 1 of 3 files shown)


python/lsst/ip/diffim/subtractImages.py
```diff
@@ -507,8 +512,17 @@ def finalize(self, template, science, difference, kernel,
...
+        # the image difference so that we can calculate correct statistics
+        # during decorrelation. Do this regardless of whether decorrelation is
+        # used for consistency.
+        template[science.getBBox()].mask.array[...] = difference.mask.array[...]
         if self.config.doDecorrelation:
             self.log.info("Decorrelating image difference.")
+            # We have cleared the template mask plane, so copy the mask plane of
+            # the image difference so that we can calculate correct statistics
+            # during decorrelation
+            template[science.getBBox()].mask.array[...] = difference.mask.array[...]
```
BrunoSanchez, line 525: I don't understand why this line is outside the if and also inside the if statement. I wouldn't place it in the if. Looking at the inline comments, I would think that the correct thing is to do just with line 519.

BrunoSanchez (approved):
I think that the few comments that I have might be more for tiny changes to avoid repetition of lines that were already executed. Otherwise this looks great to me. I approve, and the changes might be needed or not, they should not introduce significant alterations to code results or behavior.

python/lsst/ip/diffim/subtractImages.py
isullivan: Good catch! This line is indeed a duplicate of 519. I'll remove it.

PR #261, opened 2023-05-12
DM-38246: SourceDetectionTask will not detect on pixels flagged as NO_DATA or EDGE

isullivan:


tests/test_detectAndMeasure.py
```diff
@@ -233,9 +234,19 @@ def test_detect_dipoles(self):
...
+        # easily check that the correct number of sources are detected.
+        templateBorderSize = kernelSize//2
         dipoleFlag = "ip_diffim_DipoleFit_flag_classification"
         kwargs = {"seed": staticSeed, "psfSize": 2.4, "fluxLevel": fluxLevel, "fluxRange": fluxRange,
-                  "nSrc": nSources}
+                  "nSrc": nSources, "templateBorderSize": templateBorderSize, "kernelSize": kernelSize,
+                  "xSize": xSize, "ySize": ySize}
+        dipoleFlag = "ip_diffim_DipoleFit_flag_classification"
+        # kwargs = {"seed": staticSeed, "psfSize": 2.4, "fluxLevel": fluxLevel, "fluxRange": fluxRange,
+        #           "nSrc": nSources}
```
TallJimbo, line 249: Commented lines here look like relics.

tests/test_detectAndMeasure.py
```diff
@@ -312,28 +323,57 @@ def test_sky_sources(self):
...
+                    difference[srcBbox].mask.array |= lsst.afw.image.Mask.getPlaneBitMask(badMask)
+            output = detectionTask.run(science, matchedTemplate, difference)
+            refIds = []
+            goodSrcFlags = _checkMask(difference.mask, transientSources, badMaskPlanes)
+            if setFlags:
+                self.assertEqual(np.sum(~goodSrcFlags), nBad)
+            else:
+                self.assertEqual(np.sum(~goodSrcFlags), 0)
+            for diaSource, goodSrcFlag in zip(output.diaSources, goodSrcFlags):
+                if ~goodSrcFlag:
```
TallJimbo, line 370: What type is `goodSrcFlag` here, and why `~` instead of `not`?  I'm used to seeing that elementwise inversion of numpy arrays, but you also can't test for the truthiness of an entire array like this, so I wonder if you really mean `if not goodSrcFlag.any()` here.

tests/test_detectAndMeasure.py
```diff
@@ -551,6 +606,60 @@ def test_sky_sources(self):
...
+            score = subtractTask._convolveExposure(difference, scienceKernel, subtractTask.convolutionControl)
+            output = detectionTask.run(science, matchedTemplate, difference, score)
+            refIds = []
+            goodSrcFlags = subtractTask._checkMask(difference.mask, transientSources, badMaskPlanes)
+            if setFlags:
+                self.assertEqual(np.sum(~goodSrcFlags), nBad)
+            else:
+                self.assertEqual(np.sum(~goodSrcFlags), 0)
+            for diaSource, goodSrcFlag in zip(output.diaSources, goodSrcFlags):
+                if ~goodSrcFlag:
```
TallJimbo, line 655: Same comment as in the other test method.

TallJimbo (approved):

tests/test_detectAndMeasure.py
isullivan: This test is a little simpler than that: it's iterating through all of the sources, so `goodSrcFlag` is a single boolean. I could certainly change that to `not goodSrcFalg` for readability.

PR #266, opened 2023-05-22
DM-39306: Update SourceDetectionTasks to use excludeMaskPlanes EDGE by default.

erykoff:
This PR moves the excludeMaskPlanes default from the SourceDetectionTask default to the tasks within ip_diffim which need this override.


isullivan (approved):
This should be all the places the config is needed for now, thank you for catching them all.

PR #268, opened 2023-06-20
DM-39733: Specify nan_policy="omit" in model fit

timj:
This was deprecated in lmfit 0.9.8 and removed in 2019.

https://github.com/lmfit/lmfit-py/pull/565

### DIFF
--- a/python/lsst/ip/diffim/dipoleFitTask.py
+++ b/python/lsst/ip/diffim/dipoleFitTask.py
@@ -580,6 +580,8 @@ class DipoleFitAlgorithm:
         subim = afwImage.MaskedImageF(self.diffim.getMaskedImage(), bbox=bbox, origin=afwImage.PARENT)
 
         z = diArr = subim.image.array
+        # Make sure we don't overwrite buffers.
+        z = z.copy()
         weights = 1. / subim.variance.array  # get the weights (=1/variance)
 
         if rel_weight > 0. and ((self.posImage is not None) or (self.negImage is not None)):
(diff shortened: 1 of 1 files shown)


timj:
The tests currently assume that `nan_policy="raise"`.

erykoff:
The problem is that `nan_policy="omit"` doesn't work with the fancy way we are using lmfit with multiple arrays and indices.  I believe the right solution is to do what I did here, and manually filter out any infinite values first and set their weights to 0.  (This is basically what lmfit does prior to any fitting, but using masks in a way that appears to be better tested with pandas than numpy).  But I would like @isullivan to confirm that I haven't done anything crazy here.

python/lsst/ip/diffim/dipoleFitTask.py
```diff
@@ -748,6 +750,11 @@ def dipoleModelFunctor(x, flux, xcenPos, ycenPos, xcenNeg, ycenNeg, fluxNeg=None
         if np.any(~mask):
             weights[~mask] = 0.
 
+        # Filter out any nans, and make the weights 0.
+        nans = (~np.isfinite(z) | ~np.isfinite(weights))
```
isullivan, line 754: If you are only worried about `NaN`s, I think `np.isnan(z) | np.isnan(weights)` would be more clear.

python/lsst/ip/diffim/dipoleFitTask.py
```diff
@@ -748,6 +750,11 @@ def dipoleModelFunctor(x, flux, xcenPos, ycenPos, xcenNeg, ycenNeg, fluxNeg=None
...
+        z[nans] = 0
```
isullivan, line 755: I try to avoid setting image values to zero, in case there is a non-zero offset. Perhaps `z[nans] = np.nanmedian(z)`

isullivan (approved):
Looks OK, with a couple of minor suggestions.

python/lsst/ip/diffim/dipoleFitTask.py
erykoff: I guess infinities could be legal?  I've changed it.

PR #270, opened 2023-07-06
DM-39227: rename (with deprecation) SubtractImages external-calibration connections.

TallJimbo:


python/lsst/ip/diffim/subtractImages.py
```diff
@@ -69,12 +72,26 @@ class SubtractInputConnections(lsst.pipe.base.PipelineTaskConnections,
         dimensions=("instrument", "visit"),
         storageClass="ExposureCatalog",
         name="finalVisitSummary",
+        # TODO: remove on DM-39854.
+        deprecated=(
+            "Deprecated in favor of visitSummary.  Will be removed after v26."
```
isullivan, line 77: After v27?

python/lsst/ip/diffim/subtractImages.py
```diff
@@ -69,12 +72,26 @@ class SubtractInputConnections(lsst.pipe.base.PipelineTaskConnections,
...
+        )
+    )
+    visitSummary = connectionTypes.Input(
```
isullivan, line 80: For consistency, please either name this `calibrationsCatalog` in the connections, or use `visitSummary` later instead of `calibrationsCatalog`

isullivan (approved):
Looks good, but please settle on using either `calibrationsCatalog` or `visitSummary` consistently throughout.

python/lsst/ip/diffim/subtractImages.py
TallJimbo: When I wrote this, yes, it should have been v27.  Now that RFC-945 is adopted I'm going to anticipate its implementation by just a bit and leave it as v26.

PR #271, opened 2023-07-12
DM-39976: Replace masked image pixels with median values before convolution

isullivan:
Only use with preconvolution and the Score image used for detection for now.

### DIFF
--- a/python/lsst/ip/diffim/subtractImages.py
+++ b/python/lsst/ip/diffim/subtractImages.py
@@ -612,7 +612,10 @@ class AlardLuptonSubtractTask(lsst.pipe.base.PipelineTask):
     def _convolveExposure(exposure, kernel, convolutionControl,
                           bbox=None,
                           psf=None,
-                          photoCalib=None):
+                          photoCalib=None,
+                          interpolateBadMaskPlanes=False,
+                          badMaskPlanes=None,
+                          ):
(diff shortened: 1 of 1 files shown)


python/lsst/ip/diffim/subtractImages.py
```diff
@@ -1025,3 +1035,25 @@ def _shapeTest(exp1, exp2, fwhmExposureBuffer, fwhmExposureGrid):
...
+        Value to set when interpolation fails.
+    """
+    image = maskedImage.image.array
+    badPixels = (maskedImage.mask.array & maskedImage.mask.getPlaneBitMask(badMaskPlanes)) > 0
+    image[badPixels] = np.nan
+    if fallbackValue is None:
+        fallbackValue = np.nanmedian(image)
+    # For this initial implementation, skip the interpolation and just fill with
+    # the median value.
+    image[badPixels] = fallbackValue
```
BrunoSanchez, line 1059: This is good.
Fallback value could also be a local quantity? Probably for next change.

isullivan:
I think we can discuss new tickets with more detailed analysis and improvements using actual interpolation during sprint planning.

BrunoSanchez:
I have reviewed the changes and the notebooks with the results of run comparisons. 
This is surprisingly simple and yet yields noticeable lower amount of artifact transients.

PR #275, opened 2023-07-21
DM-40057: shorten deprecation timeline and add ticket TODOs

TallJimbo:


isullivan (approved):
These changes look fine, though looking through the rest of ip_diffim I think `ccdExposureIdBits` in `DiaPipelineTask` in `diaPipe.py` should also have a deprecation warning, but currently has none.

PR #276, opened 2023-07-27
DM-40194: Add fallback for gaps in the template PSF in decorrelation.

isullivan:
If the simple PSF image calculation fails during image decorrelation, compute an average PSF across the template. Note that this will only be used if the science image is convolved during A&L image differencing.

### DIFF
--- a/python/lsst/ip/diffim/imageDecorrelation.py
+++ b/python/lsst/ip/diffim/imageDecorrelation.py
@@ -28,10 +28,12 @@ import lsst.geom as geom
 import lsst.meas.algorithms as measAlg
 import lsst.pex.config as pexConfig
 import lsst.pipe.base as pipeBase
+from lsst.pex.exceptions import InvalidParameterError
 from lsst.utils.timer import timeMethod
 
 from .imageMapReduce import (ImageMapReduceConfig, ImageMapReduceTask,
                              ImageMapper)
+from .utils import computeAveragePsf
 
 __all__ = ("DecorrelateALKernelTask", "DecorrelateALKernelConfig",
            "DecorrelateALKernelMapper", "DecorrelateALKernelMapReduceConfig",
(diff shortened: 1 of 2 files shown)


python/lsst/ip/diffim/utils.py
```diff
@@ -1173,6 +1173,70 @@ def evaluateMeanPsfFwhm(exposure: afwImage.Exposure,
     return np.nanmean(width)
 
 
+def computeAveragePsf(exposure: afwImage.Exposure,
+                      psfExposureBuffer: float, psfExposureGrid: int) -> float:
```
arunkannawadi, line 1177: The return type is incorrect.

python/lsst/ip/diffim/utils.py
```diff
@@ -1173,6 +1173,70 @@ def evaluateMeanPsfFwhm(exposure: afwImage.Exposure,
...
+        The average PSF across the exposure.
+
+    Raises
+    ------
+    ValueError
+        Raised if the PSF cannot be computed at any of the grid points.
+
+    See Also
+    --------
+    evaluateMeanPsfFwhm
```
arunkannawadi, line 1204: Could you also backlink this from `evaluateMeanPsfFwhm`? And while we are modifying the docstring of `evaluateMeanPsfFwhm`, could you change the first line there to say mean instead of median?

python/lsst/ip/diffim/imageDecorrelation.py
```diff
@@ -261,7 +274,7 @@ def run(self, scienceExposure, templateExposure, subtractedExposure, psfMatching
             corr = self.computeDiffimCorrection(kArr, varianceMean, targetVarianceMean)
 
         correctedImage = self.computeCorrectedImage(corr.corrft, subtractedExposure.image.array)
-        correctedPsf = self.computeCorrectedDiffimPsf(corr.corrft, psfImg.array)
```
arunkannawadi, line 264: I'd just keep `psfImg` as `afw.Image` object. If you revert this change, then `computeAveragePsf` must return an object of that type.

arunkannawadi (approved):

python/lsst/ip/diffim/imageDecorrelation.py
```diff
@@ -27,11 +27,13 @@
 import lsst.geom as geom
 import lsst.meas.algorithms as measAlg
 import lsst.pex.config as pexConfig
+from lsst.pex.exceptions import InvalidParameterError
```
arunkannawadi, line 30: I don't think this repo uses `isort`, but the imports appear sorted already and it might be nice to maintain it by moving this down by one line.

isullivan: I'm confused. The imports look sorted to me with this line.

timj: Isort sorts imports before from x import

PR #277, opened 2023-08-21
DM-37911: Add nData to measRecord

bsmartradio:


python/lsst/ip/diffim/dipoleFitTask.py
```diff
@@ -1167,6 +1173,11 @@ def measure(self, measRecord, exposure, posExp=None, negExp=None):
         measRecord[self.signalToNoiseKey] = result.signalToNoise
         measRecord[self.chi2dofKey] = result.redChi2
 
+        if result.nData >= 1:
+            measRecord[self.nDataKey] = result.nData
+        else:
+            measRecord[self.nDataKey] = 0
```
mrawls, line 1179: I'd like to understand why the best fallback here is 0 and not np.nan.

mrawls (approved):
Does what it says on the tin! Assuming you're confident 0 is a good "no results" fallback value, I'm happy.

python/lsst/ip/diffim/dipoleFitTask.py
bsmartradio: I tried using np.nan, however despite allowing nans this causes the pipeline to fail. I wasn't able to find exactly what was tripping it up, but in this case any nan would be the same as 0 fitting points anyways, so I think it is okay.

PR #278, opened 2023-08-23
DM-39411: Add timing metrics for remaining AP pipeline tasks

kherner:


python/lsst/ip/diffim/getTemplate.py
kfindeisen, line 1: I wonder if more of these methods (e.g., `getOverlappingExposures`) should also be timed, but I guess that's a question for @isullivan.

kfindeisen (approved):
The changes themselves look fine, but I'm a bit puzzled by the commit message. What does "add required blank line" refer to? As explanatory descriptions go it's a bit... non-explanatory.

python/lsst/ip/diffim/getTemplate.py
kherner: Ay, I must have messed up the commit message merging when I did the second commit. I was missing a blank line before @timeMethod initially, and I foolishly pushed before I ran scons. I fixed it and then did the usual rebase/force push thing with the fix in, but maybe I was sloppy about getting the messages set up correctly. The commit message was supposed to be, "add timeMethod for GetTemplateTask".

kfindeisen: Could you fix the message, then? It might confuse others looking at the code history, as well.

PR #279, opened 2023-08-25
DM-15180: Add coordinate error fields to diaSrc

cmsaunders:


isullivan (approved):

isullivan (approved):
I believe you will also need to add `lsst.afw.table.updateSourceCoords(difference.getWcs(), diaSources)` after forced source measurement in `DetectAndMeasureTask.processResults`

PR #280, opened 2023-09-13
Added mask planes from template and science fake plane into difference

BrunoSanchez:
Added the needed mask plane into differences. These are two new planes named `'INJECTED'` and `'INJECTED_TEMPLATE'`

### DIFF
--- a/python/lsst/ip/diffim/subtractImages.py
+++ b/python/lsst/ip/diffim/subtractImages.py
@@ -195,12 +195,12 @@ class AlardLuptonSubtractBaseConfig(lsst.pex.config.Config):
     )
     badMaskPlanes = lsst.pex.config.ListField(
         dtype=str,
-        default=("NO_DATA", "BAD", "SAT", "EDGE"),
+        default=("NO_DATA", "BAD", "SAT", "EDGE", "FAKE"),
         doc="Mask planes to exclude when selecting sources for PSF matching."
     )
     preserveTemplateMask = lsst.pex.config.ListField(
         dtype=str,
-        default=("NO_DATA", "BAD", "SAT", "INJECTED", "INJECTED_CORE"),
+        default=("NO_DATA", "BAD", "SAT", "FAKE", "INJECTED", "INJECTED_CORE"),
         doc="Mask planes from the template to propagate to the image difference."
     )
 
@@ -461,7 +461,8 @@ class AlardLuptonSubtractTask(lsst.pipe.base.PipelineTask):
                                      psfMatchingKernel=kernelResult.psfMatchingKernel)
 
     def runConvolveScience(self, template, science, selectSources):
-        """Convolve the science image with a PSF-matching kernel and subtract the template image.
+        """Convolve the science image with a PSF-matching kernel and subtract
+        the template image.
 
         Parameters
         ----------
@@ -561,20 +562,14 @@ class AlardLuptonSubtractTask(lsst.pipe.base.PipelineTask):
         """
         # Erase existing detection mask planes.
         #  We don't want the detection mask from the science image
-        mask = difference.mask
-        mask &= ~(mask.getPlaneBitMask("DETECTED") | mask.getPlaneBitMask("DETECTED_NEGATIVE"))
 
-        # We have cleared the template mask plane, so copy the mask plane of
-        # the image difference so that we can calculate correct statistics
-        # during decorrelation. Do this regardless of whether decorrelation is
-        # used for consistency.
-        template[science.getBBox()].mask.array[...] = difference.mask.array[...]
+        self.updateMasks(template, science, difference)
+
         if self.config.doDecorrelation:
             self.log.info("Decorrelating image difference.")
             # We have cleared the template mask plane, so copy the mask plane of
             # the image difference so that we can calculate correct statistics
             # during decorrelation
-            template[science.getBBox()].mask.array[...] = difference.mask.array[...]
             correctedExposure = self.decorrelate.run(science, template[science.getBBox()], difference, kernel,
                                                      templateMatched=templateMatched,
                                                      preConvMode=preConvMode,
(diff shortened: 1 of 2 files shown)


python/lsst/ip/diffim/subtractImages.py
```diff
@@ -581,6 +576,42 @@ def finalize(self, template, science, difference, kernel,
...
+            diffInjectedBitMask = mask.getPlaneBitMask("INJECTED")
+
+            mask.addMaskPlane("INJECTED_TEMPLATE")
+            diffInjectedTemplateBitMask = mask.getPlaneBitMask("INJECTED_TEMPLATE")
+
+            scienceFakeBitMask = science.mask.getPlaneBitMask('FAKE')
+            templateFakeBitMask = template[bbox].mask.getPlaneBitMask('FAKE')
+
+            injScienceMaskArray = ((science.mask.array & scienceFakeBitMask) > 0)
+            injScienceMaskArray *= diffInjectedBitMask
```
isullivan, line 602: In testing, I get the following error from this line (and below): 
```numpy.core._exceptions._UFuncOutputCastingError: Cannot cast ufunc 'multiply' output from dtype('int64') to dtype('bool') with casting rule 'same_kind'```
I suggest removing the line `injScienceMaskArray *= diffInjectedBitMask` and using `mask.array[injScienceMaskArray] |= diffInjectedBitMask` below. You might want to change the variable name of `injScienceMaskArray` to reflect the new usage.

File: python/lsst/ip/diffim/subtractImages.py [hunk 2]
```diff
@@ -581,6 +576,42 @@ def finalize(self, template, science, difference, kernel,
...
+            injTemplateMaskArray = ((template[bbox].mask.array & templateFakeBitMask) > 0)
+            injTemplateMaskArray *= diffInjectedTemplateBitMask
+
+            mask.array |= injScienceMaskArray
```
isullivan, line 606: See my requested modification in the comment above.

python/lsst/ip/diffim/subtractImages.py
```diff
@@ -581,6 +576,42 @@ def finalize(self, template, science, difference, kernel,
             correctedExposure = difference
         return correctedExposure
 
+    def updateMasks(self, difference, science, template):
```
isullivan, line 579: Could you keep the same order of images as in `finalize`: template, science, difference

python/lsst/ip/diffim/subtractImages.py
```diff
@@ -712,7 +743,12 @@ def _checkMask(mask, sources, badMaskPlanes):
             kept (True) or rejected (False) based on the value of the
             mask plane at its location.
         """
-        badPixelMask = lsst.afw.image.Mask.getPlaneBitMask(badMaskPlanes)
+        setBadMaskPlanes = [
+            maskPlane for maskPlane in badMaskPlanes if maskPlane in mask.getMaskPlaneDict().keys()
+        ]
+
+        badPixelMask = mask.getPlaneBitMask(setBadMaskPlanes)
```
isullivan, line 750: Thanks for changing this. We need to switch from `lsst.afw.image.Mask.getPlaneBitMask` to `mask.getPlaneBitMask` in general (see https://github.com/lsst/meas_algorithms/pull/329#discussion_r1195795073)

python/lsst/ip/diffim/subtractImages.py
```diff
@@ -712,7 +743,12 @@ def _checkMask(mask, sources, badMaskPlanes):
             kept (True) or rejected (False) based on the value of the
             mask plane at its location.
         """
-        badPixelMask = lsst.afw.image.Mask.getPlaneBitMask(badMaskPlanes)
+        setBadMaskPlanes = [
+            maskPlane for maskPlane in badMaskPlanes if maskPlane in mask.getMaskPlaneDict().keys()
```
isullivan, line 747: Minor point, but I don't think the `.keys()` is necessary here.

isullivan (changes requested):
In addition to the code comments, please add a simple unit test of the new `updateMasks`.

python/lsst/ip/diffim/subtractImages.py
BrunoSanchez: Done!

python/lsst/ip/diffim/subtractImages.py
BrunoSanchez: Done

BrunoSanchez:
This is ready for second review!

tests/test_subtractTask.py
```diff
@@ -715,6 +715,76 @@ def _compare_apCorrMaps(self, a, b):
...
+        diff_mask = subtraction.difference.mask
+
+        # science mask should be now in INJECTED
+        inj_masked = (diff_mask.array & diff_mask.getPlaneBitMask("INJECTED")) > 0
+
+        # template mask should be now in INJECTED_TEMPLATE
+        injTmplt_masked = (diff_mask.array & diff_mask.getPlaneBitMask("INJECTED_TEMPLATE")) > 0
+
+        self.assertEqual(np.sum(inj_masked.astype(int)-science_fake_masked.astype(int)), 0)
+        self.assertEqual(np.sum(injTmplt_masked.astype(int)-template_fake_masked.astype(int)), 0)
```
isullivan, line 786: Since the mask plane is changing, this looks like a good way to do the test. For future reference, see https://pipelines.lsst.io/py-api/lsst.utils.tests.TestCase.html for useful built-in asserts, though I don't think any would quite work here.

tests/test_subtractTask.py
```diff
@@ -715,6 +715,76 @@ def _compare_apCorrMaps(self, a, b):
...
+        # created fakes and added them to the images
+        science.image.array += science_fake_img.image.array
+        template.image.array += tmplt_fake_img.image.array
+
+        # TODO: update to INJECTED names when source injection gets refactored
+        # adding mask planes to both science and template images
+        science.mask.addMaskPlane("FAKE")
+        template.mask.addMaskPlane("FAKE")
+
+        science_mask_planes = science.mask.getMaskPlaneDict()
```
isullivan, line 741: You could use `science_fake_mask = science.mask.getMaskPlane('FAKE')` here, and then just use `science_fake_mask` instead of `science_mask_planes['FAKE']` later. Same for the template below.

Or use `science_fake_mask = science.mask.getPlaneBitMask("FAKE")` if you use my suggestion below.

tests/test_subtractTask.py
```diff
@@ -715,6 +715,76 @@ def _compare_apCorrMaps(self, a, b):
...
+        y0 = science.mask.getY0()
+        for a_science_source in science_fake_sources:
+            # 3 x 3 masking of the source locations is fine
+            for dy in (-1, 0, 1):
+                science.mask.setMaskPlaneValues(
+                    science_mask_planes['FAKE'],
+                    int(a_science_source.getX() - x0)-1,
+                    int(a_science_source.getX() - x0)+1,
+                    int(a_science_source.getY() - y0)+dy
+                )
```
isullivan, line 754: This is fine, but I have an alternate suggestion:
```
bbox = lsst.geom.Box2I(lsst.geom.Point2I(a_science_source.getX(), a_science_source.getY()), lsst.geom.Extent2I(3, 3))
science[bbox].mask.array |= science_fake_mask
```
using `science_fake_mask` from my above suggestion.

isullivan (approved):
Looks good! I had a few suggested changes, though the only necessary one is to remove the dangling code comment.

BrunoSanchez:
Thanks for the comments. 

All of them are implemented!

BrunoSanchez:
This is the final Jenkins run that validates this merge: https://rubin-ci.slac.stanford.edu/job/stack-os-matrix/480/

PR #284, opened 2023-10-18
DM-32769: Implement streak detection and masking in single frame processing

mrawls:
This PR propagates streak mask info through `detectAndMeasureTask`.

### DIFF
--- a/python/lsst/ip/diffim/utils.py
+++ b/python/lsst/ip/diffim/utils.py
@@ -1265,6 +1265,7 @@ def detectTestSources(exposure):
         sigma=None,  # The appropriate sigma is calculated from the PSF
         doSmooth=True
     )
+    exposure.mask.addMaskPlane("STREAK")  # add empty streak mask plane in lieu of maskStreaksTask
     selectSources = detRet.sources
     selectMeasurement.run(measCat=selectSources, exposure=exposure)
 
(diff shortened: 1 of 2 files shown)


python/lsst/ip/diffim/utils.py
```diff
@@ -1265,6 +1265,7 @@ def detectTestSources(exposure):
         sigma=None,  # The appropriate sigma is calculated from the PSF
         doSmooth=True
     )
+    exposure.mask.addMaskPlane("STREAK")  # Cannot run maskStreaksTask here, as it's in pipe_tasks
```
bsmartradio, line 1268: Just curious why we need to know that we can't run maskStreakTask at this location. Is that something we might want to do?

bsmartradio (approved):
Does what is says, just curious about the one comment

python/lsst/ip/diffim/utils.py
bsmartradio: I see in another PR you do ` self.exposure.mask.addMaskPlane("STREAK")  # add empty streak mask plane in lieu of maskStreaksTask` as the explanation instead. Maybe swap to that?

mrawls: Good call for consistency's sake, will update. (I originally called the task here, and realized that wouldn't work since pipe_tasks explicitly depends on ip_diffim being setup, and the reverse cannot be true.)

PR #287, opened 2023-12-01
DM-41955: Remove unphysical diaSources from the output of detectAndMeasure

isullivan:


parejkoj (changes requested):
You need some tests of this code. It won't do what you think, because of local argument scoping in the function.

I think this is ok here as a quick fix, but I still think we should consider removing these sources higher up (as an early measurement plugin, preferably), so that we're not doing any further math on them.

python/lsst/ip/diffim/detectAndMeasure.py
```diff
@@ -141,6 +141,12 @@ class DetectAndMeasureConfig(pipeBase.PipelineTaskConfig,
         target=SkyObjectsTask,
         doc="Generate sky sources",
     )
+    badSourceFlags = lsst.pex.config.ListField(
+        dtype=str,
+        doc="Do not include sources with any of these flags set in the output catalog.",
```
parejkoj, line 146: I just realized my rewording could be misinterpreted (they're not "set in the output catalog"!). Maybe:

```suggestion
        doc="Sources with any of these flags set are not included in the output catalog.",
```

python/lsst/ip/diffim/detectAndMeasure.py
```diff
@@ -376,6 +383,33 @@ def processResults(self, science, matchedTemplate, difference, sources, table,
...
+        diaSources : `lsst.afw.table.SourceCatalog`
+            The updated catalog of detected sources, with any source that has a
+            flag in ``config.badSourceFlags`` set removed.
+        """
+        flags = np.ones(len(diaSources), dtype=bool)
+        for flag in self.config.badSourceFlags:
+            try:
+                flags &= ~diaSources[flag]
+            except Exception as e:
+                self.log.warning("Could not apply source flag: %s", e)
```
parejkoj, line 405: I don't like this try:catch; if you want to not raise on missing flags, you should only catch `KeyError`. More broadly, I think a missing flag should just raise: if someone configures their pipeline to expect a certain flag, it should be an error if it doesn't appear in the schema, otherwise it's hiding unintended behavior, like a typo in the flags list.

python/lsst/ip/diffim/detectAndMeasure.py
```diff
@@ -376,6 +383,33 @@ def processResults(self, science, matchedTemplate, difference, sources, table,
...
+        nBad = np.count_nonzero(~flags)
+        if nBad > 0:
+            self.log.warning(f"Found and removed {nBad} unphysical sources.")
```
parejkoj, line 408: No string formatting directly in log messages, use the log `%` formatters: 

https://developer.lsst.io/stack/logging.html#basic-usage-in-python

I'm torn on whether this should be `warning` or `info`: it's doing exactly the thing you asked for which implies `info`, but those things shouldn't have been there in the first place which implies `warning`. I wonder if what we really want is a count of sources with each of the "bad" flags in the loop above as an `info` message, so that we have better granularity about what's going on? Maybe this in the loop?

```
flagged = ~diaSources[flag]
nFlagged = sum(flagged)
if nFlagged > 0:
    self.log.info("Removed %d sources with %s.", nFlagged, flag)
```

tests/test_detectAndMeasure.py
```diff
@@ -189,6 +190,52 @@ def test_measurements_finite(self):
         self._check_values(output.diaSources.getY(), minValue=0, maxValue=ySize)
         self._check_values(output.diaSources.getPsfInstFlux())
 
+    def test_remove_unphysical(self):
+        """
+        """
```
parejkoj, line 195: ?

tests/test_detectAndMeasure.py
```diff
@@ -189,6 +190,52 @@ def test_measurements_finite(self):
...
+
```
parejkoj, line 196: unnecessary line

tests/test_detectAndMeasure.py
```diff
@@ -93,7 +93,7 @@ def _check_values(self, values, minValue=None, maxValue=None):
             self.assertTrue(np.all(values <= maxValue))
 
     def _setup_detection(self, doApCorr=False, doMerge=False,
-                         doSkySources=False, doForcedMeasurement=False):
+                         doSkySources=False, doForcedMeasurement=False, nSkySources=5, badSourceFlags=[]):
```
parejkoj, line 96: This is starting to get unwieldy (e.g., you've missed the Parameters on the new ones). I would replace these kwargs with a dictionary that you apply with `config.update(**kwargs)`. That way, you don't have to add new kwargs to this method as you want more config functionality.

tests/test_detectAndMeasure.py
```diff
@@ -189,6 +190,52 @@ def test_measurements_finite(self):
...
+        nSetBad = 5
+        for src in diaSources[0: nSetBad]:
+            src["slot_Centroid_x"] += xSize
+            src["slot_Centroid_y"] += ySize
+            src["base_PixelFlags_flag_offimage"] = True
+        # Verify that these sources are outside the image
+        badDiaSrc1 = ~bbox.contains(diaSources.getX(), diaSources.getY())
+        nBad1 = np.count_nonzero(badDiaSrc1)
+        self.assertEqual(nBad1, nSetBad)
+        diaSources2 = detectionTask.removeBadSources(diaSources)
```
parejkoj, line 231: This test wouldn't have caught the bug I described before (local scoping of the catalog), since it's not looking at the output of `run`. What if instead you set all of `difference.mask` to `BAD`, add the `badCenterAll` flag to the bad flags list, and then just call `detectionTask.run` and confirm that the source is not included in the output? It's crude, but your other tests should be checking that "normal" sources behave correctly, and we don't really care _why_ a given bad flag is set, just that it is.

parejkoj (changes requested):
Suggestion on how to rework the test to catch the bug you'd had before, and some questions about squashing exceptions and logging.

python/lsst/ip/diffim/detectAndMeasure.py
isullivan: I understand your objection to a try:catch here, and can certainly change this to a `KeyError`, but I would strongly prefer keeping it. My reasoning is that this should ideally not be finding and removing any bad sources, so I would rather it shout at people with a warning than have it break if the configured flag is missing.

tests/test_detectAndMeasure.py
isullivan: I disagree. If this test were implemented as `detectionTask.removeBadSources(diaSources)` then it would fail because the flagged sources are not removed from `diaSources` in this scope, as you pointed out. Now that the API is for `.removeBadSources` to return a source catalog, it seems appropriate to use that here.

python/lsst/ip/diffim/detectAndMeasure.py
```diff
@@ -140,6 +141,18 @@ class DetectAndMeasureConfig(pipeBase.PipelineTaskConfig,
...
         doc="Generate sky sources",
     )
+    badSourceFlags = lsst.pex.config.ListField(
+        dtype=str,
+        doc="Flags that, if set, the associated source should not "
+        "be included in the output diaSource catalog.",
+        default=("base_PixelFlags_flag_offimage",
+                 "base_PixelFlags_flag_interpolatedCenterAll",
+                 "base_PixelFlags_flag_saturatedCenterAll",
+                 "base_PixelFlags_flag_crCenterAll",
```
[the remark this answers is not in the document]
isullivan: I removed all of the `CenterAll` flags for now pending further investigation to unblock this ticket.

python/lsst/ip/diffim/detectAndMeasure.py
parejkoj: I very strongly disagree. An error that occurs on every single quantum (which is what will happen if the config doesn't match the schema) should be an actual exception raised up the chain, not a warning that will get lost. If it's an error, processing will stop early and we'll know to fix the misconfiguration. If it's a warning, we could process a whole data release without knowing that we meant to filter on something but didn't.

I'm not sure what you mean by "should ideally not be finding and removing any bad sources"? We know there will be "bad sources" because difference imaging will produce totally spurious detections due to interpolation. Unless we remove those sources much earlier (which is hard to do, given the design of the plugin system), we know they'll show up here and have to be removed.

tests/test_detectAndMeasure.py
parejkoj: If the goal is to test `removeBadSources` by itself as a unittest, then you don't need any of the other infrastructure here. All you need is to create a trivial SourceCatalog with a couple flags (`test_flag_1`, etc.), call `removeBadSources` on that and check that the things you didn't want aren't there. There's no need to create an image or run detection or anything.

If the goal is to test that `run` behaves correctly, then you need to actually test `run` itself. This test as written would not catch `run` not assigning the output of `removeBadSources` to something that gets returned. If this is the goal of this test you don't need to run detectAndMeasure in a way that doesn't remove anything: you're already testing that with the other tests in this class.

python/lsst/ip/diffim/detectAndMeasure.py
```diff
@@ -376,6 +383,45 @@ def processResults(self, science, matchedTemplate, difference, sources, table,
 
         return measurementResults
 
+    def removeBadSources(self, diaSources):
```
parejkoj, line 386: Oh, we should probably make this `_removeBadSources`: we're trying to make the "private" parts of Tasks more explicitly so.

python/lsst/ip/diffim/detectAndMeasure.py
```diff
@@ -376,6 +383,45 @@ def processResults(self, science, matchedTemplate, difference, sources, table,
...
+        self.metadata.add("nRemovedBadFlaggedSources", nBadTotal)
+        # Use slot_Centroid_x/y here instead of getX() method, since the former
+        #  works on non-contiguous source tables and the latter does not.
+        centroidFlag = np.isfinite(diaSources["slot_Centroid_x"]) & np.isfinite(diaSources["slot_Centroid_y"])
+        nBad = np.count_nonzero(~centroidFlag)
+        if nBad > 0:
+            self.log.info("Found and removed %d unphysical sources with non-finite centroid.", nBad)
+            self.metadata.add("nRemovedBadCentroidSources", nBadTotal)
+            nBadTotal += nBad
+            selector &= centroidFlag
```
parejkoj, line 422: You should be able to replace this entire thing with adding `slot_Centroid_flag` to the bad flags list. See DM-7102: that flag should be catching all cases where the centroid positions are non-finite.

python/lsst/ip/diffim/detectAndMeasure.py
TallJimbo: There's actually a special exception type for configuration/consistency errors that can only be caught during after execution (whenever possible we want to catch them in config validation or task construction, of course): [`lsst.pipe.base.InvalidQuantumError`](https://github.com/lsst/pipe_base/blob/9d3cecc6b43000bf6755f3880c79acf7bc896de6/python/lsst/pipe/base/_status.py#L77C7-L77C26).  I don't know the context well enough to say for certain whether it's the right call, but it seems potentially relevant.

parejkoj:
Thinking about this more, why can't we just use a ScienceSourceSelector, like Calibrate does, with `flags` and `requireFiniteRaDec`?

tests/test_detectAndMeasure.py
isullivan: The code comment here was out of date. With 20 sky sources in the test, there are some with unphysical coordinates which are removed in the first half of the test. I will fix the code comment, and I can also add an additional step that shows that some of the resulting `diaSource`s are unphysical if `badSourceFlags=[]`.

isullivan:
`ScienceSourceSelector` could be configured and used here to remove the flagged and unphysical sources, but I do not think it is a good fit here. It includes multiple types of selections (such as on flux, signal to noise, isolation, etc..) that we do not want people to accidentally configure here, and which would not be obvious to the user that they were being used to throw out good sources. I think the log message it prints would also be unhelpful: "Selected 102/104 sources" does not tell the user what is being selected or why.

python/lsst/ip/diffim/detectAndMeasure.py
isullivan: After a side discussion with @parejkoj , lsst.pipe.base.InvalidQuantumError in the task `__init__` looks like the right solution here.

python/lsst/ip/diffim/detectAndMeasure.py
isullivan: I want to explicitly catch `diaSources["slot_Centroid_x"]` (and `y`) here since that breaks downstream code. As @parejkoj and I discussed, this check should be unnecessary after DM-42313 merges since these nan centroids would then be picked up by the flag `base_PixelFlags_flag_offimage`.

parejkoj (approved):
Thanks for the cleanups!

PR #288, opened 2023-12-04
DM-41943: Do-over 41118 (opening for review)

yalsayyad:


python/lsst/ip/diffim/detectAndMeasure.py
```diff
@@ -453,6 +457,27 @@ def measureForcedSources(self, diaSources, science, wcs):
...
+        """
+        mask = difference.mask
+        goodPix = (mask.array & mask.getPlaneBitMask(self.config.detection.excludeMaskPlanes)) == 0
+        self.metadata.add("nGoodPixels", np.sum(goodPix))
+        self.metadata.add("nBadPixels", np.sum(~goodPix))
+        detPosPix = (mask.array & mask.getPlaneBitMask("DETECTED")) > 0
+        detNegPix = (mask.array & mask.getPlaneBitMask("DETECTED_NEGATIVE")) > 0
+        self.metadata.add("nPixelsDetectedPositive", np.sum(detPosPix))
+        self.metadata.add("nPixelsDetectedNegative", np.sum(detNegPix))
+        detPosPix *= ~goodPix
```
yalsayyad, line 476: Took me a sec to realize you meant `detPosPix &= ~goodPix`. 

What you have works, but was slower for this reader to read.

tests/test_subtractTask.py
```diff
@@ -67,6 +68,41 @@ def test_mismatched_template(self):
...
+        def _run_and_check_coverage(template_coverage):
+            template_cut = template.clone()
+            template_height = int(science_height*template_coverage + border)
+            template_cut.image.array[:, template_height:] = 0
+            template_cut.mask.array[:, template_height:] = template_cut.mask.getPlaneBitMask('NO_DATA')
+            config = subtractImages.AlardLuptonSubtractTask.ConfigClass()
+            if template_coverage < config.requiredTemplateFraction:
+                doRaise = True
+            elif template_coverage < config.minTemplateFractionForExpectedSuccess:
+                doRaise = True
```
yalsayyad, line 93: The task only raises `NoWorkFound` if you get a RuntimeError or lsst.pex.exceptions.Exception. I didn't follow how/that you designed the input image to subtract successfully for 0.7 but not for 0.19. Maybe an inline comment to help me out. 

Also, if you designed these input images so that the 0.19 fails, then that also means that if I change the default for `minTemplateFractionForExpectedSuccess`  to 0.18, this test would fail.

PR #290, opened 2023-12-16
DM-41391: Add limits on the number of sources used for the matching kernel

isullivan:
If more sources are available than the configured limit, select the highest signal-to-noise sources.

### DIFF
--- a/python/lsst/ip/diffim/subtractImages.py
+++ b/python/lsst/ip/diffim/subtractImages.py
@@ -193,12 +193,33 @@ class AlardLuptonSubtractBaseConfig(lsst.pex.config.Config):
         doc="Minimum signal to noise ratio of detected sources "
         "to use for calculating the PSF matching kernel."
     )
+    detectionThresholdMax = lsst.pex.config.Field(
+        dtype=float,
+        default=500,
+        doc="Maximum signal to noise ratio of detected sources "
+        "to use for calculating the PSF matching kernel."
+    )
+    maxKernelSources = lsst.pex.config.Field(
+        dtype=int,
+        default=1000,
+        doc="Maximum number of sources to use for calculating the PSF matching kernel."
+        "Set to -1 to disable."
+    )
+    minKernelSources = lsst.pex.config.Field(
+        dtype=int,
+        default=3,
+        doc="Minimum number of sources needed for calculating the PSF matching kernel."
+    )
     badSourceFlags = lsst.pex.config.ListField(
         dtype=str,
         doc="Flags that, if set, the associated source should not "
         "be used to determine the PSF matching kernel.",
         default=("sky_source", "slot_Centroid_flag",
-                 "slot_ApFlux_flag", "slot_PsfFlux_flag", ),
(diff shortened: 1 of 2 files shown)


python/lsst/ip/diffim/subtractImages.py
```diff
@@ -193,12 +193,33 @@ class AlardLuptonSubtractBaseConfig(lsst.pex.config.Config):
...
     badSourceFlags = lsst.pex.config.ListField(
         dtype=str,
         doc="Flags that, if set, the associated source should not "
         "be used to determine the PSF matching kernel.",
         default=("sky_source", "slot_Centroid_flag",
-                 "slot_ApFlux_flag", "slot_PsfFlux_flag", ),
+                 "slot_ApFlux_flag", "slot_PsfFlux_flag",
+                 "base_PixelFlags_flag_interpolated",
+                 "base_PixelFlags_flag_saturated",
+                 "base_PixelFlags_flag_bad",
```
BrunoSanchez, line 221: Should flags associated with injected sources be used here?

tests/test_subtractTask.py
```diff
@@ -402,6 +402,52 @@ def test_few_sources(self):
                                     "Cannot compute PSF matching kernel: too few sources selected."):
             task.run(template, science, sources)
 
+    def test_kernel_source_selector(self):
+        """Test with only 1 source, to check that we get a useful error.
```
BrunoSanchez, line 406: I think that this docstring is unaccurate. Add the correct description

BrunoSanchez (approved):
I read the code changes and checked the tests. Coverage doesn't drop and only minor changes requested.
Have you run a data processing check? I can if not.

python/lsst/ip/diffim/subtractImages.py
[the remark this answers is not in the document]
isullivan: Yes, once we have switched over to using the new source injection we can rely on those source flags, and remove all of the checks on the mask plane. If you tell me the new flags to use, we can add them now.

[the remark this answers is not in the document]
BrunoSanchez: I don't know exactly but I expect at least:
`injected` and `injected_core` flags, but I am figuring out how is this playing when we have injected from template ase well.

python/lsst/ip/diffim/subtractImages.py
```diff
@@ -193,12 +193,33 @@ class AlardLuptonSubtractBaseConfig(lsst.pex.config.Config):
         doc="Minimum signal to noise ratio of detected sources "
         "to use for calculating the PSF matching kernel."
     )
+    detectionThresholdMax = lsst.pex.config.Field(
+        dtype=float,
+        default=500,
+        doc="Maximum signal to noise ratio of detected sources "
+        "to use for calculating the PSF matching kernel."
```
BrunoSanchez, line 200: This is a good parameter to have available. The default value is OK, and it would be best for each instrument to have this value configured accordingly.

BrunoSanchez (approved):

PR #293, opened 2024-01-14
DM-39466: Added flagging of sources near the INJECTED and INJECTED_TEMPLATE mask plane

BrunoSanchez:


python/lsst/ip/diffim/subtractImages.py
```diff
@@ -651,18 +651,18 @@ def updateMasks(self, template, science, difference):
         mask = difference.mask
         mask &= ~(mask.getPlaneBitMask("DETECTED") | mask.getPlaneBitMask("DETECTED_NEGATIVE"))
 
+        self.log.info("Adding injected mask planes")
+        mask.addMaskPlane("INJECTED")
+        diffInjectedBitMask = mask.getPlaneBitMask("INJECTED")
+
+        mask.addMaskPlane("INJECTED_TEMPLATE")
+        diffInjTmpltBitMask = mask.getPlaneBitMask("INJECTED_TEMPLATE")
```
BrunoSanchez, line 659: Move this into the if statement block. Same in 656.

python/lsst/ip/diffim/detectAndMeasure.py
```diff
@@ -174,8 +174,14 @@ def setDefaults(self):
         self.forcedMeasurement.slots.shape = None
 
         # Keep track of which footprints contain streaks
-        self.measurement.plugins['base_PixelFlags'].masksFpAnywhere = ['STREAK']
-        self.measurement.plugins['base_PixelFlags'].masksFpCenter = ['STREAK']
+        self.measurement.plugins["base_PixelFlags"].masksFpAnywhere = ["STREAK"]
```
BrunoSanchez, line 177: Instead of append, just set the maskFpAnyWhere to the correct list.

isullivan (approved):
The code changes look fine, and my local tests with `ap_verify` pass. Please add a new unit test to check that `detectAndMeasure` correctly flags sources with footprints including pixels with the `INJECTED` or `INJECTED_TEMPLATE` mask set. A simple Jenkins re-run (without `ci_hsc` or `ci_imsim`) should be sufficient if you just add a test.

BrunoSanchez:
Succesful jenkins run: 
https://rubin-ci.slac.stanford.edu/blue/organizations/jenkins/stack-os-matrix/detail/stack-os-matrix/479/pipeline/

tests/test_detectAndMeasure.py
```diff
@@ -431,6 +431,87 @@ def _detection_wrapper(setFlags=True):
...
+        science_fake_img, science_fake_sources = makeTestImage(
+            psfSize=2.4, xSize=xSize, ySize=ySize, seed=5, nSrc=3, noiseLevel=0.25, fluxRange=1
+        )
+        template, _ = makeTestImage(psfSize=2.4, xSize=xSize, ySize=ySize, doApplyCalibration=True)
+        tmplt_fake_img, tmplt_fake_sources = makeTestImage(
+            psfSize=2.4, xSize=xSize, ySize=ySize, seed=9, nSrc=3, noiseLevel=0.25, fluxRange=1
+        )
+        # created fakes and added them to the images
+        science.image.array += science_fake_img.image.array
+        template.image.array += tmplt_fake_img.image.array
```
isullivan, line 450: You can drop the `.array` in these two lines

tests/test_detectAndMeasure.py
```diff
@@ -431,6 +431,87 @@ def _detection_wrapper(setFlags=True):
...
+        # adding mask planes to both science and template images
+        science.mask.addMaskPlane("FAKE")
+        science_fake_bitmask = science.mask.getPlaneBitMask("FAKE")
+        template.mask.addMaskPlane("FAKE")
+        template_fake_bitmask = template.mask.getPlaneBitMask("FAKE")
+
+        for a_science_source in science_fake_sources:
+            # 3 x 3 masking of the source locations is fine
+            bbox = lsst.geom.Box2I(
+                lsst.geom.Point2I(a_science_source.getX(), a_science_source.getY()), lsst.geom.Extent2I(3, 3)
```
isullivan, line 462: Perhaps use `bbox = a_science_source.getFootprint().getBBox()` instead.

tests/test_detectAndMeasure.py
```diff
@@ -431,6 +431,87 @@ def _detection_wrapper(setFlags=True):
...
+            )
+            science[bbox].mask.array |= science_fake_bitmask
+
+        for a_template_source in tmplt_fake_sources:
+            # 3 x 3 masking of the source locations is fine
+            bbox = lsst.geom.Box2I(
+                lsst.geom.Point2I(a_template_source.getX(), a_template_source.getY()),
+                lsst.geom.Extent2I(3, 3)
+            )
```
isullivan, line 471: Perhaps `bbox = a_template_source.getFootprint().getBBox()` instead.

tests/test_detectAndMeasure.py
```diff
@@ -431,6 +431,87 @@ def _detection_wrapper(setFlags=True):
...
+        # check subtraction mask plane is set where we set the previous masks
+        diff_mask = subtraction.difference.mask
+
+        # science mask should be now in INJECTED
+        inj_masked = (diff_mask.array & diff_mask.getPlaneBitMask("INJECTED")) > 0
+
+        # template mask should be now in INJECTED_TEMPLATE
+        injTmplt_masked = (diff_mask.array & diff_mask.getPlaneBitMask("INJECTED_TEMPLATE")) > 0
+
+        self.assertEqual(np.sum(inj_masked.astype(int)-science_fake_masked.astype(int)), 0)
```
isullivan, line 490: LSST developer guide requires spaces around `+` and `-` operators

tests/test_detectAndMeasure.py
```diff
@@ -431,6 +431,87 @@ def _detection_wrapper(setFlags=True):
...
+        self.assertEqual(np.sum(injTmplt_masked.astype(int)-template_fake_masked.astype(int)), 0)
```
isullivan, line 491: For these two asserts, use `self.assertMasksEqual(inj_masked.astype(int), science_fake_masked.astype(int))`, which will give a more helpful error message if they are different.
I don't think there is an assert that checks arrays of booleans, so I think you do have to still do the `.astype(int)`. The tests pass without it, but if the arrays are actually different then the assert itself causes an error, which is unhelpful!
Note that in this case you could use `self.assertMasksEqual ` as I suggest, or `self.assertImagesEqual` or `assertFloatsEqual`

tests/test_detectAndMeasure.py
```diff
@@ -431,6 +431,87 @@ def _detection_wrapper(setFlags=True):
...
+
+        # Now check that detection of fakes have the correct flag for injections
+        detectionTask = self._setup_detection()
+        excludeMaskPlanes = detectionTask.config.detection.excludeMaskPlanes
+        nBad = len(excludeMaskPlanes)
+        self.assertGreater(nBad, 0)
```
isullivan, line 497: Maybe make this more strict, with `self.assertEqual(nBad, 1)`

tests/test_detectAndMeasure.py
```diff
@@ -431,6 +431,87 @@ def _detection_wrapper(setFlags=True):
...
+                                   subtraction.matchedTemplate,
+                                   subtraction.difference)
+
+        sci_refIds = []
+        tmpl_refIds = []
+        for diaSrc in output.diaSources:
+            if diaSrc['base_PsfFlux_instFlux'] > 0:
+                self._check_diaSource(science_fake_sources, diaSrc, scale=1, refIds=sci_refIds)
+                self.assertTrue(diaSrc['base_PixelFlags_flag_injected'])
+                self.assertTrue(diaSrc['base_PixelFlags_flag_injectedCenter'])
```
isullivan, line 509: Add:
```
                self.assertFalse(diaSrc['base_PixelFlags_flag_injected_template'])
                self.assertFalse(diaSrc['base_PixelFlags_flag_injected_templateCenter'])
```

tests/test_detectAndMeasure.py
```diff
@@ -431,6 +431,87 @@ def _detection_wrapper(setFlags=True):
...
+            else:
+                self._check_diaSource(tmplt_fake_sources, diaSrc, scale=-1, refIds=tmpl_refIds)
```
isullivan, line 511: Add
```
                self.assertFalse(diaSrc['base_PixelFlags_flag_injected'])
                self.assertFalse(diaSrc['base_PixelFlags_flag_injectedCenter'])
```

isullivan (approved):
I've added my review of the new unit test

tests/test_detectAndMeasure.py
BrunoSanchez: Chose `assertFloatsEqual` in this case. 
I have followed your suggestion. Thanks

tests/test_detectAndMeasure.py
BrunoSanchez: Done

BrunoSanchez:
I have applied the changes requested for the test. This improves significantly the quality of this unit test. Thanks very much

PR #295, opened 2024-02-13
DM-42814: Cleanup makeKernel candidateList handling

parejkoj:


python/lsst/ip/diffim/psfMatch.py
```diff
@@ -38,6 +40,8 @@
 from . import diffimLib
 
 
+@deprecated(reason="This config class is no longer used. Will be removed after v27.",
+            version="v27.0", category=FutureWarning)
```
BrunoSanchez, line 44: Is this showing up in the docs and also during runtime?

src/KernelCandidate.cc
```diff
@@ -8,7 +8,7 @@
  *
  * @ingroup ip_diffim
  */
-
+#include <stdexcept>
```
BrunoSanchez, line 11: Is this the LSST exception module?

python/lsst/ip/diffim/makeKernel.py
```diff
@@ -419,7 +419,7 @@ def _buildCellSet(self, convolved, reference, candidateList):
                                                             scienceCutout,
                                                             candidateConfig)
 
-            self.log.debug("Candidate %d at %f, %f rating=%f",
+            self.log.debug("Candidate %d at %.2f, %.2f rating=%f",
```
BrunoSanchez, line 422: Good call here. Logs are usually overloaded

python/lsst/ip/diffim/makeKernel.py
```diff
@@ -273,71 +268,77 @@ def getSelectSources(self, exposure, sigma=None, doSmooth=True, idFactory=None):
...
+            size = 2*kernelSize + 1  # ensure the resulting box is odd
+            bbox = lsst.geom.Box2I.makeCenteredBox(candidate.getCentroid(),
+                                                   lsst.geom.Extent2I(size, size))
+            boxFootprint = lsst.afw.detection.Footprint(lsst.afw.geom.SpanSet(bbox))
+            boxFootprint.addPeak(peak.getFx(), peak.getFy(), peak.getPeakValue())
+            candidate.setFootprint(boxFootprint)
+
+            # Reject footprints not contained in either image.
+            if not reference.getBBox().contains(bbox) or not convolved.getBBox().contains(bbox):
+                good[i] = 0
```
BrunoSanchez, line 328: Is it preferred to assign `0` or `False` instead? For readability sake.

python/lsst/ip/diffim/makeKernel.py
```diff
@@ -273,71 +268,77 @@ def getSelectSources(self, exposure, sigma=None, doSmooth=True, idFactory=None):
...
+                continue
+            # Reject footprints with any bad mask bits set.
+            if (reference.subset(bbox).mask.array & bitmask).any():
+                good[i] = 0
```
BrunoSanchez, line 332: Same here

python/lsst/ip/diffim/psfMatch.py
parejkoj: This will appear any time someone instantiates this class, including in the logs of pipelines that use it.

But this reminded me that I need to also deprecate the ConfigField that uses it in `PsfMatchConfig`.

src/KernelCandidate.cc
parejkoj: Nope, `<>` includes are from the standard library. It was necessary for `std::logic_error`.

python/lsst/ip/diffim/makeKernel.py
parejkoj: Good catch. `0` is `False` in a numpy array, but you're right that making it explicitly `False` is clearer.

parejkoj:
Looking at this again, I think I will have to file an RFC, even though it seems silly: any deprecations do officially need one.

PR #297, opened 2024-02-22
DM-42814: Remove doc file for removed Task

parejkoj:
I forgot to remove the doc .rst file for the removed `DiaCatalogSourceSelectorTask`.


ktlim (approved):
I was worried there was a reference somewhere in the index.rst files, but it looks like they're all going off the files that exist, so this should be good.  Thanks!

PR #301, opened 2024-03-02
DM-42774: Rename difference image mask planes to include _TEMPLATE if they are set from the template

isullivan:


python/lsst/ip/diffim/utils.py
```diff
@@ -1018,7 +1018,7 @@ def computeAveragePsf(exposure: afwImage.Exposure,
     return psfImage
 
 
-def detectTestSources(exposure):
+def detectTestSources(exposure, addMaskPlanes=None):
```
BrunoSanchez, line 1021: Please add the docs for this new optional parameter.

tests/test_detectAndMeasure.py
BrunoSanchez, line 1: Looks good.

BrunoSanchez (approved):
Looks fine. 
I think it is much better than before. I think this is more clear, and it makes more sense, as opposed to inherit mask planes from the template images without an easy way to track provenance.

PR #302, opened 2024-03-08
DM-41108: Add missing template coverage fraction metric

isullivan:
A bugfix on DM-41943 removed the calculation of `templateCoverageFraction` but did not update analysis_tools. This adds the metric calculation back in.


kfindeisen:
Are you referring to #288? I don't see the `templateCoveragePercent` removal anywhere there....

isullivan:
Sorry, it was removed on https://github.com/lsst/ip_diffim/pull/286 without a ticket, since `main` was broken, and DM-41943 re-implemented the reverted code without this metric. I unfortunately did not also remove the metric from `analysis_tools`, which left it broken.

kfindeisen (approved):
Got it, now things make sense. I agree that this implementation is much better.

PR #305, opened 2024-03-21
DM-40507: Cleanup DipoleFitTask

parejkoj:


isullivan (approved):
Looks good overall. I am a little concerned that our fake source completeness gets a little worse when I run ap_verify with this ticket, but that is something for us to investigate further on another ticket.

python/lsst/ip/diffim/dipoleFitTask.py
```diff
@@ -100,70 +99,71 @@ class DipoleFitPluginConfig(measBase.SingleFramePluginConfig):
...
-        self.slots.centroid = "ip_diffim_NaiveDipoleCentroid"
+        # This will be switched to "ip_diffim_DipoleFit" as this task runs.
+        self.slots.centroid = "base_SdssCentroid"
         self.doReplaceWithNoise = False
 
 
 class DipoleFitTask(measBase.SingleFrameMeasurementTask):
-    """A task that fits a dipole to a difference image, with an optional separate detection image.
+    """A task that fits a dipole to a difference image, with an optional
+    separate detection image.
```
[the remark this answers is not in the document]
parejkoj: The `posExp/negExp` kwargs are what give this support.

python/lsst/ip/diffim/dipoleFitTask.py
```diff
@@ -1113,6 +1113,7 @@ def measureDipoles(self, measRecord, exposure, posExp=None, negExp=None):
         ):
             measRecord.set(self.classificationFlagKey, False)
             measRecord.set(self.classificationAttemptedFlagKey, False)
+            # TODO: Do we want to fail here, or just leave it marked as not fit?
```
parejkoj, line 1110: @isullivan : this question still needs answering: I would lean towards not "failing" them, and just leaving them marked as not being attempted to be fit nor as dipoles.

python/lsst/ip/diffim/dipoleFitTask.py
```diff
@@ -174,14 +174,15 @@ def run(self, sources, exposure, posExp=None, negExp=None, **kwargs):
         **kwargs
             Additional keyword arguments for `lsst.meas.base.sfm.SingleFrameMeasurementTask`.
         """
+        self.plugins = self.plugins_pre
```
isullivan, line 177: You describe what you're doing in the description above, but please add a code comment here and before the next `self.plugins...` line below stating what each `super().run(...` call is doing. I can see future developers getting confused.

python/lsst/ip/diffim/dipoleFitTask.py
```diff
@@ -936,24 +936,22 @@ def display2dArray(arr, title='Data', extent=None):
...
         else:
-            fig = plt.figure(figsize=(8, 2.5))
+            plt.figure(figsize=(8, 2.5))
             plt.subplot(1, 3, 1)
             display2dArray(z, 'Data', extent=extent)
             plt.subplot(1, 3, 2)
             display2dArray(fit, 'Model', extent=extent)
             plt.subplot(1, 3, 3)
             display2dArray(z - fit, 'Residual', extent=extent)
-            return fig
```
isullivan, line 956: Why remove `fig` from the return? Note that it is still documented as the return value. Also the function `display2dArray` does still return `fig`, but that isn't used.

python/lsst/ip/diffim/dipoleFitTask.py
```diff
@@ -1105,25 +1096,33 @@ def measure(self, measRecord, exposure, posExp=None, negExp=None):
...
         pks = measRecord.getFootprint().getPeaks()
 
         # Check if the footprint consists of a putative dipole - else don't fit it.
         if (
                 # One peak in the footprint (not a dipole)
-                (len(pks) <= 1)
+                ((nPeaks := len(pks)) <= 1)
                 # Peaks are the same sign (not a dipole)
-                or (len(pks) > 1 and (np.sign(pks[0].getPeakValue())
+                or (nPeaks > 1 and (np.sign(pks[0].getPeakValue())
```
isullivan, line 1107: Are the peaks guaranteed to be ordered from largest positive peak to largest negative peak? I am concerned about checking pks[0] and pks[-1] explicitly, instead of sorting the peaks. If the peaks *are* sorted, please add a code comment stating that assumption.

python/lsst/ip/diffim/dipoleFitTask.py
```diff
@@ -1105,25 +1096,33 @@ def measure(self, measRecord, exposure, posExp=None, negExp=None):
         result : TODO: DM-17458
             TODO: DM-17458
         """
-
```
isullivan, line 1108: I believe you can delete `Returns` and the TODOs in the docstring.

python/lsst/ip/diffim/dipoleFitTask.py
```diff
@@ -1152,23 +1149,23 @@ def measure(self, measRecord, exposure, posExp=None, negExp=None):
 
         # add chi2, coord/flux uncertainties (TBD), dipole classification
         # Add the relevant values to the measRecord
-        measRecord[self.posFluxKey] = result.posFlux
-        measRecord[self.posFluxErrKey] = result.signalToNoise   # to be changed to actual sigma!
-        measRecord[self.posCentroidKeyX] = result.posCentroidX
-        measRecord[self.posCentroidKeyY] = result.posCentroidY
+        measRecord[self.posFluxKey.getInstFlux()] = result.posFlux
```
isullivan, line 1152: I didn't realize this was how we needed to set by key!

tests/test_detectAndMeasure.py
```diff
@@ -720,28 +694,11 @@ def test_detect_dipoles(self):
         scienceKernel = science.psf.getKernel()
         score = subtractTask._convolveExposure(difference, scienceKernel, subtractTask.convolutionControl)
 
-        # Configure the detection Task
-        detectionTask = self._setup_detection(doMerge=False)
-
-        # Run detection and check the results
+        detectionTask = self._setup_detection(doMerge=True)
```
isullivan, line 697: You can remove `doMerge=True`

isullivan (approved):
Looks good. A few more comments and cleanups requested.

python/lsst/ip/diffim/dipoleFitTask.py
parejkoj: Because it was causing the `show` down below to not get used, and it wasn't used by anything.

python/lsst/ip/diffim/dipoleFitTask.py
parejkoj: Yes, our convention is that peaks are always ordered from highest to lowest. This is a pretty deep convention and a *lot* of code depends on it.

python/lsst/ip/diffim/dipoleFitTask.py
parejkoj: This is one way. I think when this was written, the string-based method (`measRecord["blah_instFlux"]`) was much slower, which has since been mostly fixed. But having the keys in the plugin makes it easier, since you can just get the field key directly.

PR #306, opened 2024-03-27
DM-43404: Create diffim QA metrics

isullivan:


python/lsst/ip/diffim/detectAndMeasure.py
```diff
@@ -547,13 +642,27 @@ def measureForcedSources(self, diaSources, science, wcs):
...
 
-    def calculateMetrics(self, difference):
+    def calculateMetrics(self, difference, diaSources, science, matchedTemplate, idFactory):
         """Add image QA metrics to the Task metadata.
 
         Parameters
         ----------
         difference : `lsst.afw.image.Exposure`
             The target image to calculate metrics for.
+        diaSources : TYPE
```
ebellm, line 652: Docstrings need updates in this section

python/lsst/ip/diffim/detectAndMeasure.py
```diff
@@ -156,8 +170,27 @@ class DetectAndMeasureConfig(pipeBase.PipelineTaskConfig,
...
+        dtype=str,
+        doc="List of mask planes to include in metrics",
+        default=('BAD', 'CLIPPED', 'CR', 'DETECTED', 'DETECTED_NEGATIVE', 'EDGE',
+                 'INEXACT_PSF', 'INJECTED', 'INJECTED_TEMPLATE', 'INTRP', 'NOT_DEBLENDED',
+                 'NO_DATA', 'REJECTED', 'SAT', 'SAT_TEMPLATE', 'SENSOR_EDGE', 'STREAK', 'SUSPECT',
+                 'UNMASKEDNAN',
+                 ),
+    )
+    metricSources = pexConfig.ConfigurableField(
+        target=SkyObjectsTask,
```
ebellm, line 186: I wonder if using sky objects as our probe will cause us to underestimate the mask fractions?  But as we discussed it's a reasonable place to start.

python/lsst/ip/diffim/detectAndMeasure.py
```diff
@@ -567,6 +676,65 @@ def calculateMetrics(self, difference):
...
+        src.set('y', peak['i_y'])
+        src.setCoord(science.wcs.pixelToSky(peak['i_x'], peak['i_y']))
+        selectSources = diaSources[bbox.contains(diaSources.getX(), diaSources.getY())]
+        if self.config.doSkySources:
+            selectSources = selectSources[~selectSources["sky_source"]]
+        sourceDensity = len(selectSources)/area
+        dipoleSources = selectSources[selectSources["ip_diffim_DipoleFit_flag_classification"]]
+        dipoleDensity = len(dipoleSources)/area
+        if dipoleSources:
+            meanDipoleOrientation = _angleMean(dipoleSources["ip_diffim_DipoleFit_orientation"])
```
ebellm, line 722: The docstring above said this was relative to the parallactic angle--I don't see that computation here, am I missing something?

python/lsst/ip/diffim/detectAndMeasure.py
```diff
@@ -653,3 +821,13 @@ def run(self, science, matchedTemplate, difference, scoreExposure,
 
         return self.processResults(science, matchedTemplate, difference, sources, idFactory,
                                    positiveFootprints=positives, negativeFootprints=negatives)
+
+
+def _angleMean(angles):
```
ebellm, line 826: a test of this function would help confirm we got the math right....

ebellm (changes requested):
Some clarifications here and there

python/lsst/ip/diffim/detectAndMeasure.py
```diff
@@ -247,6 +281,62 @@ def __init__(self, **kwargs):
         for flag in self.config.badSourceFlags:
             if flag not in self.schema:
                 raise pipeBase.InvalidQuantumError("Field %s not in schema" % flag)
+
+        if self.config.doWriteMetrics:
```
ebellm, line 285: As I said on the analysis_tools PR, I'm worried about calling these "metrics" as well when they are stored and handled quite differently from both `lsst.verify.Metrics` and `analysis_tools` metrics.

python/lsst/ip/diffim/detectAndMeasure.py
isullivan: I originally thought it should be relative to parallactic angle, but then decided to simplify the calculation. I'll update the docstring.

python/lsst/ip/diffim/detectAndMeasure.py
isullivan: Changed to `spatiallySampledMetrics`

ebellm (approved):

PR #310, opened 2024-04-09
DM-43736: Persist the PSF-matching kernel

isullivan:


python/lsst/ip/diffim/subtractImages.py
```diff
@@ -95,6 +95,12 @@ class SubtractImageOutputConnections(lsst.pipe.base.PipelineTaskConnections,
         storageClass="ExposureF",
         name="{fakesType}{coaddName}Diff_matchedExp",
     )
+    psfMatchingKernel = connectionTypes.Output(
+        doc="Kernel used to PSF match the science and template images.",
+        dimensions=("instrument", "visit", "detector"),
+        storageClass="MatchingKernel",
+        name="{fakesType}{coaddName}Diff_psfMatchKernel",
```
natelust, line 102: my only concern here, and it is a minor one, is that the name might be confusing seen in the registry or something. The dimensions of this type are visit, detector which are for the Exposure being matched, but the name refers to the coadd that is being matched to. This is not really a problem per-say, and I don't have an alternative solution, I just wanted to bring it to your attention in case you it sparked something in your mind.

natelust (approved):

python/lsst/ip/diffim/subtractImages.py
isullivan: I think that's OK, since we already have several dataset types with the same dimensions and names such as `{fakesType}{coaddName}Diff_templateExp` so this should appear alongside many similarly-named registry entries.

PR #313, opened 2024-04-23
DM-44029: Deprecate NaiveDipoleCentroid and NaiveDipoleFlux

parejkoj:


include/lsst/ip/diffim/DipoleAlgorithms.h
```diff
@@ -208,7 +208,9 @@ inline DipoleFluxAlgorithm::DipoleFluxAlgorithm(
 class that knows how to calculate centroids as a simple unweighted first
  * moment of the 3x3 region around the peaks
  */
-class NaiveDipoleFlux : public DipoleFluxAlgorithm {
+class [[deprecated(
+        "This algorithm is deprecated and will be removed after v27.")]]
```
arunkannawadi, line 212: Could you include the names of the algorithms in the deprecation warnings?

include/lsst/ip/diffim/DipoleAlgorithms.h
```diff
@@ -299,10 +303,10 @@ class PsfDipoleFlux : public DipoleFluxAlgorithm {
     {
         meas::base::CentroidResultKey::addFields(schema, name+"_pos_centroid", "psf fitted center of positive lobe", meas::base::SIGMA_ONLY);
         meas::base::CentroidResultKey::addFields(schema, name+"_neg_centroid", "psf fitted center of negative lobe", meas::base::SIGMA_ONLY);
-        meas::base::CentroidResultKey::addFields(schema, name+"_centroid", "average of negative and positive lobe positions", meas::base::SIGMA_ONLY);
+        meas::base::CentroidResultKey::addFields(schema, name, "average of negative and positive lobe positions", meas::base::SIGMA_ONLY);
         _posCentroid = meas::base::CentroidResultKey(schema[name+"_pos_centroid"]);
         _negCentroid = meas::base::CentroidResultKey(schema[name+"_neg_centroid"]);
-        _avgCentroid = meas::base::CentroidResultKey(schema[name+"_centroid"]);
```
arunkannawadi, line 305: Removing the suffixes seem like it could break other codes, or user scripts. For something that's going to be removed soon, is it necessary? At the very least, it should be on its own separate commit.

python/lsst/ip/diffim/dipoleAlgorithms.cc
```diff
@@ -158,6 +161,9 @@ void wrapDipoleAlgorithms(lsst::cpputils::python::WrapperCollection &wrappers) {
     declarePsfDipoleFlux(wrappers);
 }
 
+// Remove these pragmas on DM-44030
+#pragma GCC diagnostic pop
+
```
arunkannawadi, line 166: Could you add a comment on what these `pragma` s are doing?

arunkannawadi (approved):
LGTM! One major request is to not rename the field. There's also a reference to this algorithm in DipoleMeasurementTask.rst under the `doc/` directory. They should also be marked for removal in DM-44030.

include/lsst/ip/diffim/DipoleAlgorithms.h
parejkoj: That comes "for free" with the compiler warning that is emitted when you try to use the class.

python/lsst/ip/diffim/dipoleAlgorithms.cc
parejkoj: They're from the C++ deprecation docs: https://developer.lsst.io/stack/deprecating-interfaces.html#c-deprecation

include/lsst/ip/diffim/DipoleAlgorithms.h
parejkoj: Unfortunately, if I don't change it some existing tests (that use DipoleMeasurementTask, which are only used in tests now and one of the things I'd like to deprecate in the future) will fail because they set `ip_diffim_PsfDipoleFlux` as the centroid slot, and the centroid slot requires that "name_x"/"name_y" be the centroid fields, not "name_centroid_x".

Nothing in the stack has used these values for a long time, so I think we're ok.

arunkannawadi: Alright, then!

PR #315, opened 2024-05-16
DM-44363: clear the NOT_DEBLENDED mask plane for the difference image

isullivan:


python/lsst/ip/diffim/subtractImages.py
```diff
@@ -831,8 +831,6 @@ def _prepareInputs(self, template, science, visitSummary=None):
             self.log.info("Science variance scaling factor: %.2f", sciVarFactor)
             self.metadata.add("scaleScienceVarianceFactor", sciVarFactor)
 
-        # Erase existing detection mask planes.
-        #  We don't want the detection mask from the science image
```
parejkoj, line 835: Thanks for removing this: it was always going to be misleading compared with the docstring of `updateMasks`.

python/lsst/ip/diffim/detectAndMeasure.py
```diff
@@ -472,6 +472,10 @@ def deblend(footprints):
             sources['parent'] = 0
             return sources.copy(deep=True)
 
+        # Clear existing deblender mask planes, if present
+        if "NOT_DEBLENDED" in difference.mask.getMaskPlaneDict().keys():
+            mask = difference.mask
+            mask &= ~(mask.getPlaneBitMask("NOT_DEBLENDED"))
```
parejkoj, line 478: I think this should happen at the same place the clearing of the detected planes happens, at line 337. You can just add it to the ORed bits being cleared there.

tests/test_subtractTask.py
```diff
@@ -143,10 +140,7 @@ def test_clear_template_mask(self):
         diffimMask = output.difference.mask
         for maskPlane, scienceSum in scienceMaskCheck.items():
             diffimSum = np.sum(diffimMask.array & mask.getPlaneBitMask(maskPlane) > 0)
```
parejkoj, line 142: Highlighting here, but the line above about "Except the "DETECTED" planes should have been cleared" should be deleted. Please check for other places in the tests that make that claim about the mask, since it's now false.

parejkoj (commented):
I'm still not sure I understand what the tests are supposed to be doing (e.g. `self.assertTrue(diffimSum >= scienceSum)` is checking that there are more of that mask bit set in one image than the other? What is that informative of?).

Otherwise, I think this is ok, and there's not obviously other bits we should clear from the list you posted.

python/lsst/ip/diffim/detectAndMeasure.py
isullivan: We aren't guaranteed to have `NOT_DEBLENDED` present, so we still need to have this check. I can consolidate this with clearing the `DETECTED` mask.

python/lsst/ip/diffim/subtractImages.py
```diff
@@ -846,10 +844,7 @@ def updateMasks(self, template, science):
             in the task config.
         science : `lsst.afw.image.Exposure`
             Science exposure to subtract from the template.
-            The DETECTED and DETECTED_NEGATIVE mask planes of the science image
-            will be erased.
         """
-        self._clearMask(science.mask, clearMaskPlanes=["DETECTED", "DETECTED_NEGATIVE"])
```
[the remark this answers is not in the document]
isullivan: OK, so would you prefer I remove the second commit that modified `subtractImages`?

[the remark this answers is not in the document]
parejkoj: Yes, I think so. I think it is more misleading to have a detection mask set before we've done detection than whatever benefit we might gain in debugging, but I'm not 100% sure.

python/lsst/ip/diffim/detectAndMeasure.py
```diff
@@ -332,9 +332,13 @@ def run(self, science, matchedTemplate, difference,
...
             idFactory = lsst.meas.base.IdGenerator().make_table_id_factory()
 
-        # Ensure that we start with an empty detection mask.
+        # Ensure that we start with an empty detection and deblended mask.
         mask = difference.mask
-        mask &= ~(mask.getPlaneBitMask("DETECTED") | mask.getPlaneBitMask("DETECTED_NEGATIVE"))
+        clearMaskPlanes = ["DETECTED", "DETECTED_NEGATIVE", "NOT_DEBLENDED"]
+        for mp in clearMaskPlanes:
+            if mp not in mask.getMaskPlaneDict():
+                mask.addMaskPlane(mp)
```
parejkoj, line 340: I understand why you're doing this, but it makes me a bit uncomfortable. We should be getting a coherent mask plane dict from earlier in the pipeline, if we're not that's a problem.

At the very least a comment above these lines saying why you're adding them would be good.

parejkoj (approved):
One more comment.

PR #316, opened 2024-05-17
DM-44319: Refactor deblending in DetectAndMeasureTask

fred3m:
The previous deblending algorith was not properly deblending, as it was attempting to use meas_deblender for use cases that were not algorithmically suitable.

This commit only runs the deblender on blends that have all positive peaks, the regime that it is designed for.

### DIFF
--- a/python/lsst/ip/diffim/detectAndMeasure.py
+++ b/python/lsst/ip/diffim/detectAndMeasure.py
@@ -21,7 +21,6 @@
 
 import numpy as np
 
-import lsst.afw.detection as afwDetection
 import lsst.afw.table as afwTable
 import lsst.daf.base as dafBase
 import lsst.geom
@@ -104,7 +103,8 @@ class DetectAndMeasureConfig(pipeBase.PipelineTaskConfig,
         dtype=bool,
         default=True,
         doc="Merge positive and negative diaSources with grow radius "
-            "set by growFootprint"
+            "set by growFootprint",
+        deprecated="This field is no longer used and will be removed after v28."
     )
     doForcedMeasurement = pexConfig.Field(
         dtype=bool,
@@ -143,7 +143,8 @@ class DetectAndMeasureConfig(pipeBase.PipelineTaskConfig,
     growFootprint = pexConfig.Field(
         dtype=int,
         default=2,
-        doc="Grow positive and negative footprints by this many pixels before merging"
+        doc="Grow positive and negative footprints by this many pixels before merging",
+        deprecated="This field is no longer used and will be removed after v28."
     )
     diaSourceMatchRadius = pexConfig.Field(
         dtype=float,
@@ -346,16 +347,11 @@ class DetectAndMeasureTask(lsst.pipe.base.PipelineTask):
             doSmooth=True,
         )
 
-        sources, positives, negatives = self._deblend(difference,
-                                                      results.positive,
-                                                      results.negative)
+        sources = self._buildCatalogAndDeblend(difference, results.positive, results.negative, idFactory)
 
-        return self.processResults(science, matchedTemplate, difference, sources, idFactory,
-                                   positiveFootprints=positives,
-                                   negativeFootprints=negatives)
+        return self._measureSources(science, matchedTemplate, difference, sources)
 
-    def processResults(self, science, matchedTemplate, difference, sources, idFactory,
-                       positiveFootprints=None, negativeFootprints=None,):
+    def _measureSources(self, science, matchedTemplate, difference, initialDiaSources):
         """Measure and process the results of source detection.
 
         Parameters
@@ -367,15 +363,8 @@ class DetectAndMeasureTask(lsst.pipe.base.PipelineTask):
             difference image.
         difference : `lsst.afw.image.ExposureF`
             Result of subtracting template from the science image.
-        sources : `lsst.afw.table.SourceCatalog`
+        initialDiaSources : `lsst.afw.table.SourceCatalog`
             Detected sources on the difference exposure.
-        idFactory : `lsst.afw.table.IdFactory`
-            Generator object used to assign ids to detected sources in the
-            difference image.
-        positiveFootprints : `lsst.afw.detection.FootprintSet`, optional
-            Positive polarity footprints.
-        negativeFootprints : `lsst.afw.detection.FootprintSet`, optional
-            Negative polarity footprints.
 
         Returns
         -------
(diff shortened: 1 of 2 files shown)


python/lsst/ip/diffim/detectAndMeasure.py
```diff
@@ -346,16 +346,11 @@ def run(self, science, matchedTemplate, difference,
...
+        sources = self._buildCatalogAndDeblend(difference, results.positive, results.negative, idFactory)
 
-        return self.processResults(science, matchedTemplate, difference, sources, idFactory,
-                                   positiveFootprints=positives,
-                                   negativeFootprints=negatives)
+        return self.processResults(science, matchedTemplate, difference, sources)
 
-    def processResults(self, science, matchedTemplate, difference, sources, idFactory,
-                       positiveFootprints=None, negativeFootprints=None,):
+    def processResults(self, science, matchedTemplate, difference, initialDiaSources):
```
parejkoj, line 353: I'd love to change this to `_measureSources` or something. `processResults` has never been very informative. It's only sort of part of the public API, but noone should have been calling it before, so I think you're safe to rename it (and fix the one call of it in the Score task below)?

python/lsst/ip/diffim/detectAndMeasure.py
```diff
@@ -453,33 +425,61 @@ def _deblend(self, difference, positiveFootprints, negativeFootprints):
...
+        # Deblend the footprints that can be deblended with SourceDeblendTask.
+        temp_cat = afwTable.SourceCatalog(sources.table)
+        for sid in blend_ids:
+            temp_cat.append(sources.find(sid))
+        first_child_index = len(temp_cat)
+        self.deblend.run(exposure=difference, sources=temp_cat)
+
+        # Update the deblended parents and add their children
+        # to the sources catalog.
+        sources.extend(temp_cat[first_child_index:], deep=False)
```
parejkoj, line 464: I think you want to do a deep copy here to ensure the catalog is contiguous, but maybe that's not necessary for the measurement tasks that are run in after this? And then the catalog is written to disk. I guess we can see if we get failures when running on larger data?

python/lsst/ip/diffim/detectAndMeasure.py
```diff
@@ -453,33 +425,61 @@ def _deblend(self, difference, positiveFootprints, negativeFootprints):
...
+
+        # Set detection and primary flags
+        for sid in skipped_ids:
+            src = sources.find(sid)
+            self.deblend.skipParent(src, difference.mask)
+        self.setPrimaryFlags.run(sources)
```
parejkoj, line 470: I understand what setPrimaryFlags is doing here, but I'm not familiar with `skipParent`. Could you please expand the comment above, since I don't think we're setting any "flags" here, but rather the "not deblended" mask plane (I think?)?

parejkoj (commented):
One concern is that downstream processing will have to filter on something to remove the "multiple parent copies". I don't know if we should also do that on this ticket, but we maybe should. I think that would be in `FilterDiaSourceCatalogTask` in `ap_assocation`. I don't know if we need to do it elsewhere, too; having both parents and children in the diaSource catalogs is something we're not fully familiar with yet.

python/lsst/ip/diffim/detectAndMeasure.py
```diff
@@ -453,33 +425,61 @@ def _deblend(self, difference, positiveFootprints, negativeFootprints):
...
+        merged_footprints = positiveFootprints
+        merged_footprints.merge(negativeFootprints, 0, 0, False)
+
+        # Create a source catalog from the footprints.
+        table = afwTable.SourceTable.make(self.schema, idFactory)
+        sources = afwTable.SourceCatalog(table)
+        merged_footprints.makeSources(sources)
+
+        # Sky sources must be added before deblending, otherwise the
+        # sources with parent == 0 will be out of order and SFM cannot run.
```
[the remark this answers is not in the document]
fred3m: I'll change to "measurement tasks." The point is that measurement plugins require all of the children with the same parent ID to be contiguous, but the old version of this task added the sky sources after deblending, so if deblending had worked as expected then measurement would fail.

python/lsst/ip/diffim/detectAndMeasure.py
```diff
@@ -453,33 +425,61 @@ def _deblend(self, difference, positiveFootprints, negativeFootprints):
...
-
-        positives = deblend(positiveFootprints)
-        negatives = deblend(negativeFootprints)
-
-        sources = afwTable.SourceCatalog(self.schema)
-        sources.reserve(len(positives) + len(negatives))
-        sources.extend(positives, deep=True)
-        sources.extend(negatives, deep=True)
-        return sources, makeFootprints(positives), makeFootprints(negatives)
+        # Merge the positive and negative footprints, but don't grow them.
```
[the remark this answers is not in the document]
fred3m: I'll have to check the code here to double check, but I'm pretty sure that the footprints created by `SourceDetectionTask` have already been grown. So growing them again will create unnecessary blending.

python/lsst/ip/diffim/detectAndMeasure.py
```diff
@@ -453,33 +425,61 @@ def _deblend(self, difference, positiveFootprints, negativeFootprints):
...
+        table = afwTable.SourceTable.make(self.schema, idFactory)
+        sources = afwTable.SourceCatalog(table)
+        merged_footprints.makeSources(sources)
+
+        # Sky sources must be added before deblending, otherwise the
+        # sources with parent == 0 will be out of order and SFM cannot run.
+        if self.config.doSkySources:
+            self.addSkySources(sources, difference.mask, difference.info.id)
+
+        # Find the footprints with multiple positive peaks no negative peaks.
```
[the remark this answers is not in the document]
fred3m: Yes. I wasn't thinking about there being "negative footprints" that aren't part of a dipole, but I guess that a variable sources that is fainter in the visit than the template would show up that way, right?

python/lsst/ip/diffim/detectAndMeasure.py
fred3m: No, we actually don't want a deep copy (see Jim's comments in slack). The temporary catalog actually created a view to the original records, that way the source IDs stay in order and the original records are updated in place.

fred3m: I checked and verified that using this procedure the sources are always contiguous.

python/lsst/ip/diffim/detectAndMeasure.py
fred3m: Sure. The basic idea is that the deblender already has a method to set all of the appropriate flags for a parent source when deblending is skipped, so we just call that method to set those flags here.

python/lsst/ip/diffim/detectAndMeasure.py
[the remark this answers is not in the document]
parejkoj: I am a bit surprised by that, because I think that deblending was working on "isolated" positive blends before (it was just messing up other things), and we didn't see failures.

python/lsst/ip/diffim/detectAndMeasure.py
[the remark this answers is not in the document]
parejkoj: Yes, SourceDetectionTask grows the footprints. I'm pretty sure you're correct here, and that we've always been over-growing during the merge.

Which reminds me: if you're deprecating `doMerge`, you should also deprecate `growFootprint`, since that was only used in the old merge code.

python/lsst/ip/diffim/detectAndMeasure.py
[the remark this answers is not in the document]
parejkoj: Yes, that's exactly the case I'm thinking of. As I said, I think it's unlikely that there will be many such "two sources got fainter together" blends that aren't also blended with nearby positive sources, but we might as well just treat both cases the same.

python/lsst/ip/diffim/detectAndMeasure.py
parejkoj: I think calling it "detection" as the comment does isn't quite correct, though, since we've already done detection. Maybe `Set blending-related mask planes and flags on the non-deblended parents.` would be more clear?

fred3m: Yes, see my latest commit. But for the `setPrimary` section I still use this comment because technically all of the flags set by `isPrimary` are `detect_<flagName>`, since they are about detection.

python/lsst/ip/diffim/detectAndMeasure.py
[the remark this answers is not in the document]
fred3m: Yeah, it's actually because before it wasn't doing any deblending, so all of the sources had `parent==0`. The measurement framework has a call to `SourceCatalog.getChildren(0)`, where the argument is the parent ID. This is done to run all of the measurements on the parents, but in the new catalog if the sky objects were appended to it then the records with parent == 0 was discontinuous.

parejkoj:
Couple more comments about comments, otherwise this looks good, thank you

PR #318, opened 2024-05-21
DM-44150: Remove unnecessary DetectAndMeasure plugins

parejkoj:


python/lsst/ip/diffim/detectAndMeasure.py
```diff
@@ -199,16 +198,10 @@ def setDefaults(self):
...
-        # Add filtered flux measurement, the correct measurement for pre-convolved images.
-        self.measurement.algorithms.names.add("base_PeakLikelihoodFlux")
         self.measurement.plugins.names |= ["ext_trailedSources_Naive",
                                            "base_LocalPhotoCalib",
                                            "base_LocalWcs",
-                                           "ext_shapeHSM_HsmSourceMoments",
-                                           "ext_shapeHSM_HsmPsfMoments",
                                            ]
-        self.measurement.slots.psfShape = "ext_shapeHSM_HsmPsfMoments"
-        self.measurement.slots.shape = "ext_shapeHSM_HsmSourceMoments"
```
isullivan, line 211: I am slightly confused by these being moved to `dipoleFitTask`. Both seem fine, though setting here seems more general.

isullivan (approved):
Looks fine, though I am slightly confused by the move of HSM defaults to `dipoleFitTask`.

python/lsst/ip/diffim/detectAndMeasure.py
parejkoj: It's a good question. `dipoleFitTask` is our `MeasurementTask` for diffim; it's what runs all of our measurement plugins. So I think it's the lowest level place to put the list of what measurements we consider to be canonical. I guess nothing in dipoleFitTask depends on the output of a shape algorithm, so we could just not define anything for the shape slot in that task and do it here (maybe with a note in dipoleFit to that effect?)? I could go either way.

parejkoj: I've thought about this more and I think you're right and I'll take it a step further: I'm going to move the plugins currently set in dipoleFitTask that are unnecessary for just that task out of it and into detectAndMeasure. That makes dipoleFit more self-contained, and puts the rest of the diffim measurements together in one place.

parejkoj: Having poked at this some, it's not worth the added complexity of moving plugins out of dipoleFitTask, because of the various slots that have to be unset and then reset in detectAndMeasure.

PR #321, opened 2024-05-23
DM-44494: Change DipoleFit "orientation" from degrees to radians.

wmwv:
Changes unit of `orientation` for dipole to be degree instead of rad to match Science Pipeline standard.  Fixed calculation, specified unit, and tests.  ~6 lines changed.

Passes `ci_ap`:

https://rubin-ci.slac.stanford.edu/blue/organizations/jenkins/stack-os-matrix/detail/stack-os-matrix/1555/pipeline


isullivan (approved):
Looks good. I added a commit to also convert the utility function that reads the dipole orientation angle to radians.

PR #322, opened 2024-05-23
DM-39557: Create estimated depth metric difference imaging

mrawls:
This PR adds a new difference image limiting magnitude metric to the detectAndMeasureTask metadata, which analysis_tools will then put in the corresponding metric bundle. It will be useful for identifying regions where the template is of poor quality, e.g., few input visits, and we should not expect high quality diaSources.

### DIFF
--- a/python/lsst/ip/diffim/detectAndMeasure.py
+++ b/python/lsst/ip/diffim/detectAndMeasure.py
@@ -20,7 +20,6 @@
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
 import numpy as np
-from astropy import units as u
 
 import lsst.afw.detection as afwDetection
 import lsst.afw.table as afwTable
@@ -425,7 +424,7 @@ class DetectAndMeasureTask(lsst.pipe.base.PipelineTask):
         if self.config.doForcedMeasurement:
             self.measureForcedSources(diaSources, science, difference.getWcs())
 
-        self.calculateMetrics(difference, matchedTemplate, science)
+        self.calculateMetrics(difference)
 
         measurementResults = pipeBase.Struct(
             subtractedMeasuredExposure=difference,
@@ -592,7 +591,7 @@ class DetectAndMeasureTask(lsst.pipe.base.PipelineTask):
         for diaSource, forcedSource in zip(diaSources, forcedSources):
             diaSource.assign(forcedSource, mapper)
 
-    def calculateMetrics(self, difference, matchedTemplate, science):
+    def calculateMetrics(self, difference):
         """Add difference image QA metrics to the Task metadata.
 
         This may be used to produce corresponding metrics (see
@@ -602,12 +601,6 @@ class DetectAndMeasureTask(lsst.pipe.base.PipelineTask):
         ----------
         difference : `lsst.afw.image.Exposure`
             The target difference image to calculate metrics for.
-        matchedTemplate : `lsst.afw.image.Exposure`
-            The warped and PSF-matched template used to create the difference
-            image.
-        science : `lsst.afw.image.Exposure`
-            Science exposure that the template was subtracted from.
-
         """
         mask = difference.mask
         badPix = (mask.array & mask.getPlaneBitMask(self.config.detection.excludeMaskPlanes)) > 0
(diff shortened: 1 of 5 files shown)


python/lsst/ip/diffim/detectAndMeasure.py
```diff
@@ -623,6 +630,39 @@ def calculateMetrics(self, difference):
...
+        Returns
+        -------
+        maglim : `astropy.units.Quantity`
+            The limiting magnitude of the exposure.
+        """
+        zeropoint = exposure.getPhotoCalib().instFluxToMagnitude(1)
+        psf = exposure.getPsf()
+        psf_shape = psf.computeShape(exposure.getBBox().getCenter())
+        psf_size = psf_shape.getDeterminantRadius()
+        psf_area = np.pi*psf_size**2
```
isullivan, line 662: The same calculation can be simplified to `psf_size = psf_shape.getArea()`

However, you could also make it more accurate by trying to account for the elongation of the PSF, with something like `psf_area = np.pi*np.sqrt(psf_shape.getIxx()*psf_shape.getIyy()`. The square root is necessary since the second moments are already in units of distance squared, and you can use the cross terms to be more accurate.

tests/test_detectAndMeasure.py
```diff
@@ -562,6 +562,42 @@ def test_mask_streaks(self):
         streakMaskSet = (outMask & streakMask) > 0
         self.assertTrue(np.all(streakMaskSet[20:23, 40:200]))
 
+    def test_metadata_metrics(self):
+        """Verify fields are added to metadata when detection is run, and
+        that the difference image limiting magnitude is calculated correctly,
+        both with a "good" and "bad" seeing template.
+        """
+        science, sources = makeTestImage(psfSize=1.8, doApplyCalibration=True)
```
isullivan, line 570: `sources` is not actually used, so should probably be removed.

tests/test_detectAndMeasure.py
```diff
@@ -562,6 +562,42 @@ def test_mask_streaks(self):
...
+        # The metadata fields are attached to the detectionTask, so we do
+        # need to run that; run it for both "good" and "bad" seeing templates
+        detectionTask_good = self._setup_detection()
+        detectionTask_bad = self._setup_detection()
+        _ = detectionTask_good.run(science, matchedTemplate_good, difference_good)
+        _ = detectionTask_bad.run(science, matchedTemplate_bad, difference_bad)
+
+        # Test that the diffim limiting magnitudes are computed correctly
+        self.assertFloatsAlmostEqual(detectionTask_good.metadata['diffimLimitingMagnitude'],
+                                     27.8383475, atol=1e-6)
```
isullivan, line 589: You set the psf size, so you can calculate the reference value of the limiting magnitude in the test instead of hard-coding a value.

python/lsst/ip/diffim/detectAndMeasure.py
```diff
@@ -593,13 +594,19 @@ def measureForcedSources(self, diaSources, science, wcs):
         for diaSource, forcedSource in zip(diaSources, forcedSources):
             diaSource.assign(forcedSource, mapper)
 
-    def calculateMetrics(self, difference):
-        """Add image QA metrics to the Task metadata.
+    def calculateMetrics(self, difference, matchedTemplate, science):
+        """Add difference image QA metrics to the Task metadata.
+        This may be used to produce corresponding metrics (see
```
isullivan, line 599: Add a space between the first line of the docstring and the details.

python/lsst/ip/diffim/detectAndMeasure.py
```diff
@@ -593,13 +594,19 @@ def measureForcedSources(self, diaSources, science, wcs):
...
+        lsst.analysis.tools.tasks.diffimTaskDetectorVisitMetricAnalysis).
 
         Parameters
         ----------
         difference : `lsst.afw.image.Exposure`
-            The target image to calculate metrics for.
+            The target difference image to calculate metrics for.
+        matchedTemplate : `lsst.afw.image.Exposure`
+            The warped and PSF-matched template used to create difference.
```
isullivan, line 607:
```suggestion
            The warped and PSF-matched template used to create the difference
            image.
```

python/lsst/ip/diffim/detectAndMeasure.py
```diff
@@ -623,6 +630,39 @@ def calculateMetrics(self, difference):
                 self.metadata.add("%s_mask_fraction"%maskPlane.lower(), -1)
                 self.log.info("Unable to calculate metrics for mask plane %s: not in image"%maskPlane)
 
+        maglim_science = self._calculateMagLim(science)
+        fluxlim_science = (maglim_science*u.ABmag).to_value(u.nJy)
+        maglim_template = self._calculateMagLim(matchedTemplate)
+        fluxlim_template = (maglim_template*u.ABmag).to_value(u.nJy)
+        maglim_diffim = (np.sqrt(fluxlim_science**2 + fluxlim_template**2)*u.nJy).to(u.ABmag).value
```
isullivan, line 637: We might as well add the science and template limiting magnitudes to the metadata, since we have those values handy.

isullivan (changes requested):
I have a few cleanup suggestions, but also one more major one.

The major issue is that we must include error handling any time we compute a shape of the psf of the template, because the psf of the template is not defined in regions where there are gaps in coverage and the pipeline will fail with a `InvalidParameterError`. There is an example in the subtraction Task that handles this by falling back on evaluating the psf on a grid across the template if there is an error, though I think that is more cumbersome than is needed here.

python/lsst/ip/diffim/detectAndMeasure.py
```diff
@@ -623,6 +630,39 @@ def calculateMetrics(self, difference):
...
+        nsigma : `float`, optional
+            The detection threshold in sigma.
+
+        Returns
+        -------
+        maglim : `astropy.units.Quantity`
+            The limiting magnitude of the exposure.
+        """
+        zeropoint = exposure.getPhotoCalib().instFluxToMagnitude(1)
+        psf = exposure.getPsf()
```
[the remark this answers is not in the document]
mrawls: Thanks for flagging this. I've changed this so it will set the template limiting magnitude to `nan` if it hits this edge case. In this situation, it will adopt the (admittedly optimistic) science limiting magnitude for the diffim.

PR #325, opened 2024-05-28
DM-44491: Exclude more mask planes from detection

isullivan:


tests/test_detectAndMeasure.py
bsmartradio, line 1: Line 401 has ` test_edge_detections`, which looks as if it is simulating the bad flags and then checking they are properly excluded. This probably should have its name updated to indicate its testing all bad flags, not just the edge. Maybe change to `test_mask_detections`?

bsmartradio (approved):
Looks good, with one comment on possibly updating a test name.

tests/test_detectAndMeasure.py
isullivan: I'll change it to `test_exclude_mask_detections`.

PR #326, opened 2024-06-01
DM-43855: Improve getTemplateTask runtime

parejkoj:


python/lsst/ip/diffim/getTemplate.py
```diff
@@ -223,89 +232,202 @@ def run(self, coaddExposures, bbox, wcs, dataIds, physical_filter=None, **kwargs
...
+            record.setValidPolygon(afwGeom.Polygon(geom.Box2D(coadd.getBBox()).getCorners()))
+            record.set("tract", dataId["tract"])
+            record.set("patch", dataId["patch"])
+            # TODO: Modify this once we know which coadds were actually included?
+            record.set("weight", 1)
+
+        return images, catalog, totalBox
+
+    def _merge(self, maskedImages, bbox, wcs):
+        """Merge the images that came from one tract into a one larger image.
```
TallJimbo, line 346:
```suggestion
        """Merge the images that came from one tract into one larger image.
```

python/lsst/ip/diffim/getTemplate.py
```diff
@@ -223,89 +232,202 @@ def run(self, coaddExposures, bbox, wcs, dataIds, physical_filter=None, **kwargs
...
+            weight = afwImage.ImageF(maskedImage.variance.array**(-0.5))
+            is_nan = np.isnan(maskedImage.image.array)
+            # Note that modifying the patch MaskedImage in place is fine;
+            # we're throwing it away at the end anyway.
+            maskedImage.image.array[is_nan] = 0.0
+            maskedImage.variance.array[is_nan] = 0.0
+            # Reset mask, too, since these pixels don't contribute to sum.
+            maskedImage.mask.array[is_nan] = 0
+            # Cannot use `merged.maskedImage *= weight` because that operator
+            # multiplies by the weight twice; in this case `weight` are the
```
TallJimbo, line 375:
```suggestion
            # multiplies the variance by the weight twice; in this case `weight` are the
```

python/lsst/ip/diffim/getTemplate.py
```diff
@@ -223,89 +232,202 @@ def run(self, coaddExposures, bbox, wcs, dataIds, physical_filter=None, **kwargs
...
+        Returns
+        -------
+        merged : `lsst.afw.image.MaskedImage`
+            Merged image with all of the inputs at their respective bbox
+            positions.
+        """
+        merged = afwImage.ExposureF(bbox, wcs)
+        weights = afwImage.ImageF(bbox)
+        for maskedImage in maskedImages:
+            weight = afwImage.ImageF(maskedImage.variance.array**(-0.5))
```
TallJimbo, line 366: Should comment here or a little lower that we expect `weight` to be zero for the `is_nan` pixels automatically.  Might even be worth a check if it's not too expensive.

python/lsst/ip/diffim/getTemplate.py
```diff
@@ -223,89 +232,202 @@ def run(self, coaddExposures, bbox, wcs, dataIds, physical_filter=None, **kwargs
...
+            maskedImage.mask.array[is_nan] = 0
+            # Cannot use `merged.maskedImage *= weight` because that operator
+            # multiplies by the weight twice; in this case `weight` are the
+            # exact values we want to scale by.
+            maskedImage.image *= weight
+            maskedImage.variance *= weight
+            merged.maskedImage[maskedImage.getBBox()] += maskedImage
+            weights[maskedImage.getBBox()] += weight
+        # Cannot use `merged.maskedImage /= weights` because that operator
+        # divides by the weight twice; in this case `weights` are the exact
```
TallJimbo, line 382:
```suggestion
        # divides the variance by the weight twice; in this case `weights` are the exact
```

python/lsst/ip/diffim/getTemplate.py
```diff
@@ -223,89 +232,202 @@ def run(self, coaddExposures, bbox, wcs, dataIds, physical_filter=None, **kwargs
...
         # CoaddPsf centroid not only must overlap image, but must overlap the
         # part of image with data. Use centroid of region with data.
-        boolmask = templateExposure.mask.array & templateExposure.mask.getPlaneBitMask('NO_DATA') == 0
+        boolmask = template.mask.array & template.mask.getPlaneBitMask('NO_DATA') == 0
         maskx = afwImage.makeMaskFromArray(boolmask.astype(afwImage.MaskPixel))
         centerCoord = afwGeom.SpanSet.fromMask(maskx, 1).computeCentroid()
 
         ctrl = self.config.coaddPsf.makeControl()
-        coaddPsf = CoaddPsf(tractsCatalog, finalWcs, centerCoord, ctrl.warpingKernelName, ctrl.cacheSize)
+        # TODO: how could this possibly return None?
```
TallJimbo, line 425: If you mean the `ctrl` object, I don't think it can.

tests/test_getTemplate.py
```diff
@@ -0,0 +1,265 @@
...
+        # WCS should match our exposure, not any of the coadd tracts.
+        for tract in self.patches:
+            self.assertNotEqual(template.wcs, self.patches[tract][0].wcs)
+        self.assertEqual(template.wcs, self.exposure.wcs)
+        self.assertEqual(template.photoCalib, self.exposure.photoCalib)
+        self.assertEqual(template.getXY0(), expectedBox.getMin())
+        self.assertEqual(template.filter.bandLabel, "a")
+        self.assertEqual(template.filter.physicalLabel, "a_test")
+        self.assertEqual(template.psf.getComponentCount(), nInputs)
+        # TODO: need other things to test here?
```
TallJimbo, line 178: I think this can be dropped now.

tests/test_getTemplate.py
```diff
@@ -0,0 +1,265 @@
...
+        self.assertTrue(np.all(np.isfinite(template.image.array)))
+        # Because of the scale changes, there will be some ringing in the
+        # difference between the template and the original image; pick
+        # tolerances large enough to account for that.
+        self.assertImagesAlmostEqual(template.image, self.exposure[expectedBox].image,
+                                     rtol=.1, atol=4)
+        # Variance plane ==2 in the original image, but the warped images will
+        # have some structure due to the warping.
+        self.assertImagesAlmostEqual(template.variance, self.exposure[expectedBox].variance,
+                                     rtol=0.5, msg="variance planes differ")
```
TallJimbo, line 201: FWIW, if you want to be able to shrink this tolerance, I'd _guess_ the best way to do it would be to shrink the scaling from 1.05 to 1.01 or something.

TallJimbo (approved):
Looks great!  Comments are mostly flagging TODO comments that I found.

python/lsst/ip/diffim/getTemplate.py
```diff
@@ -146,11 +152,12 @@ def getOverlappingExposures(self, inputs):
...
-               List of Coadd exposures that overlap the detector (`list`
-               [`lsst.afw.image.Exposure`]).
+               Dict of Coadd exposures that overlap the detector, indexed on
+               tract id (`dict` [`int`, `list` [`lsst.afw.image.Exposure`] ]).
            ``dataIds``
-               List of data IDs of the coadd exposures that overlap the
-               detector (`list` [`lsst.daf.butler.DataCoordinate`]).
+               Dict of data IDs of the coadd exposures that overlap the
+               detector, indexed on tract id
+               (`dict` [`int`, `list [`lsst.daf.butler.DataCoordinate`] ]).
```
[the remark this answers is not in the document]
parejkoj: You're right, but I'm already changing the API slightly by making physical_filter a required arg. I don't want to rework `run` even more, since that would probably need an RFC.

python/lsst/ip/diffim/getTemplate.py
```diff
@@ -223,89 +232,202 @@ def run(self, coaddExposures, bbox, wcs, dataIds, physical_filter=None, **kwargs
...
+            totalBox = totalBox.expandedTo(coadd.getBBox())
+            record = catalog.addNew()
+            record.setPsf(coadd.psf)
+            record.setWcs(coadd.wcs)
+            record.setPhotoCalib(coadd.photoCalib)
+            record.setBBox(coadd.getBBox())
+            record.setValidPolygon(afwGeom.Polygon(geom.Box2D(coadd.getBBox()).getCorners()))
+            record.set("tract", dataId["tract"])
+            record.set("patch", dataId["patch"])
+            # TODO: Modify this once we know which coadds were actually included?
```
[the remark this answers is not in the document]
parejkoj: I don't know if it's necessary for anything downstream. I suppose I could remove it and see if anything breaks, since it provides no information (and at this point, actually incorrect information).

python/lsst/ip/diffim/getTemplate.py
parejkoj: Should I just make `is_nan` the union of the `NaN` image pixels and the non-finite variance pixels?

python/lsst/ip/diffim/getTemplate.py
parejkoj: That's what I thought. I copied this from the old code, so I'll just remove and we can add back some other test-and-raise if we need it.

tests/test_getTemplate.py
parejkoj: Ok. I wasn't sure if there was anything else I should be checking, especially regarding the PSF. Some tests are better than no tests, but I'd rather be as comprehensive as I could now.

python/lsst/ip/diffim/getTemplate.py
TallJimbo: Sure, that sounds reasonable.

python/lsst/ip/diffim/getTemplate.py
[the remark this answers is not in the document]
parejkoj: Ah,`CoaddPsf` needs there to be a `weights` field, so I'm going to leave it as is.

tests/test_getTemplate.py
parejkoj: Unfortunately, that doesn't seem to change it much. The 3 degree rotation seems to contribute quite a bit.

python/lsst/ip/diffim/getTemplate.py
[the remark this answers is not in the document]
TallJimbo: Hmm.  It probably doesn't matter much that this value is incorrect, then, since I'd expect the coadd PSFs from two different patches in an overlap region to be very similar, and downstream code is probably not super sensitive to PSF quality.  But we should update this comment to say something like that.

PR #330, opened 2024-07-05
DM-45131: Correct logger class in docstring

timj:


python/lsst/ip/diffim/subtractImages.py
```diff
@@ -1138,7 +1138,7 @@ def checkTemplateIsSufficient(templateExposure, logger, requiredTemplateFraction
     ----------
     templateExposure : `lsst.afw.image.ExposureF`
         The template exposure to check
-    logger : `lsst.log.Log`
+    logger : `loggnig.Logger`
```
mwittgen, line 1141:
```suggestion
    logger : `logging.Logger`
```

mwittgen (approved):
One typo

PR #332, opened 2024-07-12
DM-44997: Fix missing NO_DATA

parejkoj:


mrawls (approved):
Assuming Jenkins passes once it is online etc., this fix looks great.

PR #333, opened 2024-07-18
Fix incorrect parameter docstring

parejkoj:


python/lsst/ip/diffim/getTemplate.py
```diff
@@ -134,9 +134,9 @@ def getOverlappingExposures(self, inputs):
         Parameters
         ----------
         inputs : `dict` of task Inputs, containing:
-            - coaddExposureRefs : `list` \
-                                [`lsst.daf.butler.DeferredDatasetHandle` of \
-                                `lsst.afw.image.Exposure`]
+            - coaddExposure : `list` \
```
erykoff, line 137: Should be `coaddExposures` (plural), right?

PR #335, opened 2024-08-06
DM-10454: Add check that science and template have matching bands.

isullivan:


mrawls (approved):
Looks good, please also update the `_validateExposures` docstring to mention the band check is also happening here.

PR #339, opened 2024-09-06
DM-42087: Add more streak parameters to output

cmsaunders:


mrawls (commented):
Following discussion at the AP meeting, please configure `onlyMaskDetected` to False in (next to the maskStreaks and writeStreakInfo configs around line 170) so it is the default behavior in difference imaging. Then this is good to go.

cmsaunders:
I took a slightly different approach, since `maskStreaks` is a `ConfigurableField`, and added a line in `setDefaults` to configure `maskStreaks` with `onlyMaskDetected = False`.

python/lsst/ip/diffim/detectAndMeasure.py
```diff
@@ -221,6 +221,8 @@ def setDefaults(self):
             "STREAK", "INJECTED", "INJECTED_TEMPLATE"]
         self.skySources.avoidMask = ["DETECTED", "DETECTED_NEGATIVE", "BAD", "NO_DATA", "EDGE"]
 
+        self.maskStreaks.onlyMaskDetected = False
```
mrawls, line 224: This is a great place to set it, thank you - suggest one final inline comment e.g. `# Set the streak mask along the entire fit line, not only where the detected mask is set`

mrawls (approved):
Good catch, this looks great, one more in-line comment ought to do it!

PR #345, opened 2024-10-22
DM-45318: Add NO_DATA to detectAndMeasure's excludeMaskPlanes

mrawls:
This PR adds NO_DATA to the list of masks to ignore when detecting sources in difference images alongside EDGE, SAT, BAD, and INTRP.


abudlong (approved):

mrawls:
Plot twist, turns out the meas_algorithms change broke `testDipoleEdge`, which is designed to test that dipoles close to image edges are not detected. It makes sense that some of these should now be detected if they are close to the edge but (presumably) not _super_ close to the edge, so I'll see if I can fine-tune the test dipole coordinates to be more in the "super close" regime, unless @isullivan has a better suggestion.

PR #346, opened 2024-10-23
DM-47084: getOverlappingExposures docs and other cleanups

parejkoj:


python/lsst/ip/diffim/getTemplate.py
```diff
@@ -126,46 +126,51 @@ def runQuantum(self, butlerQC, inputRefs, outputRefs):
...
 
     def getOverlappingExposures(self, inputs):
-        """Return lists of coadds and their corresponding dataIds that overlap
-        the detector.
+        """Return a data structure containing the coadds that overlap the
+        specified bbox projected onto the sky, and a corresponding data
+        structure of their dataIds.
+        These are the appropriate inputs to this task's `run` method.
 
         The spatial index in the registry has generous padding and often
```
christinawilliams, line 134: I think I don't understand what "spatial index" here is referring to. If this is one of the inputs maybe its worth explaining which?

python/lsst/ip/diffim/getTemplate.py
```diff
@@ -126,46 +126,51 @@ def runQuantum(self, butlerQC, inputRefs, outputRefs):
...
         supplies patches near, but not directly overlapping the detector.
-        Filters inputs so that we don't have to read in all input coadds.
+        This filters the inputs so that we don't have to read in all
+        possibly-matching coadds.
```
christinawilliams, line 138: By "read in" does it mean load all the images in the input list (to check for overlap)? Perhaps this stems from my confusion over "spatial index in the registry" (does this mean butler registry, i.e. is referring to how all the images are indexed in the database?

python/lsst/ip/diffim/getTemplate.py
```diff
@@ -126,46 +126,51 @@ def runQuantum(self, butlerQC, inputRefs, outputRefs):
...
 
         Parameters
         ----------
         inputs : `dict` of task Inputs, containing:
             - coaddExposures : `list` \
                               [`lsst.daf.butler.DeferredDatasetHandle` of \
                                `lsst.afw.image.Exposure`]
-                Data references to exposures that might overlap the detector.
+                Data references to exposures that might overlap the desired
+                region.
```
christinawilliams, line 146: I think I was sending it a dictionary containing a list of data references? I find info on the format needed to be useful but I defer to developers the correct way to document these things

python/lsst/ip/diffim/getTemplate.py
```diff
@@ -126,46 +126,51 @@ def runQuantum(self, butlerQC, inputRefs, outputRefs):
...
-                resample the coaddExposures.
+                Template bounding box of the pixel geometry onto which the
+                coaddExposures will be resampled.
             - skyMap : `lsst.skymap.SkyMap`
-                Input definition of geometry/bbox and projection/wcs for
-                template exposures.
+                Geometry of the tracts and patches the coadds are defined on.
             - wcs : `lsst.afw.geom.SkyWcs`
-                Template WCS onto which to resample the coaddExposures.
+                Template WCS onto which the coadds will be resampled, .
```
christinawilliams, line 154: an extra comma at the end of the -wcs explanation

christinawilliams: Also separate question: does this function only work on coadds, or is it possible this would be used for calexp files also (i.e. perhaps more generic reference to input images rather than specifically coadds would make sense?)

christinawilliams (approved):
Thanks! I left some comments including minor typos and also some places where I thought more clarification would be useful (but I defer to the developers whether these should be included in order to stick with convention for documentation)

python/lsst/ip/diffim/getTemplate.py
```diff
@@ -44,35 +44,37 @@ class GetTemplateConnections(pipeBase.PipelineTaskConnections,
...
         dimensions=("skymap", ),
         storageClass="SkyMap",
     )
     # TODO DM-31292: Add option to use global external wcs from jointcal
     # Needed for DRP HSC
     coaddExposures = pipeBase.connectionTypes.Input(
-        doc="Input template to match and subtract from the exposure",
+        doc="Coadds that may overlap the desired region, as possible inputs to the template."
+            " Will be restricted to those that directly overlap the projected bounding box.",
         dimensions=("tract", "patch", "skymap", "band"),
```
christinawilliams, line 69: extra space before "Will"

christinawilliams (approved):
Looks good! Much improved (at least from the perspective of someone trying to learn how to use this). Just flagged one typo.

python/lsst/ip/diffim/getTemplate.py
parejkoj: As noted in the docstring, this just takes a list of datarefs. See the type listing above this line.

python/lsst/ip/diffim/getTemplate.py
parejkoj: Coadds are organized per tract/patch, other exposures are not. That tract/patch organization is a requirement here.

python/lsst/ip/diffim/getTemplate.py
parejkoj: That's necessary, so that there is a space after the period on the previous line.

python/lsst/ip/diffim/getTemplate.py
parejkoj: Please have another look at this text: I reworked it a bit to hopefully be more clear?

christinawilliams (approved):

PR #350, opened 2024-11-04
DM-47368: Restrict when subtractImages sources selector raises

isullivan:
Do not raise an error in source selection if allowKernelSourceDetection=True


python/lsst/ip/diffim/subtractImages.py
```diff
@@ -822,7 +822,8 @@ def _sourceSelector(self, sources, mask):
             self.log.error("Too few sources to calculate the PSF matching kernel: "
                            "%i selected but %i needed for the calculation.",
                            len(selectSources), self.config.minKernelSources)
-            raise RuntimeError("Cannot compute PSF matching kernel: too few sources selected.")
+            if ~self.config.allowKernelSourceDetection:
```
TallJimbo, line 825:
```suggestion
            if not self.config.allowKernelSourceDetection:
```

PR #352, opened 2024-11-05
DM-46852: Make detectAndMeasure robust to invalid PSF sigma 

isullivan:


python/lsst/ip/diffim/detectAndMeasure.py
```diff
@@ -365,6 +373,33 @@ def run(self, science, matchedTemplate, difference,
...
+        mask = difference.mask
+        for mp in self.config.clearMaskPlanes:
+            if mp not in mask.getMaskPlaneDict():
+                mask.addMaskPlane(mp)
+        mask &= ~mask.getPlaneBitMask(self.config.clearMaskPlanes)
+        sigma = difference.psf.computeShape(difference.psf.getAveragePosition()).getDeterminantRadius()
+        if np.isnan(sigma):
+            sigma = self.config.fallbackPsfSigma
+            self.log.warning("Could not determine average width of the difference image PSF. "
+                             "Using the default sigma of %d", self.config.fallbackPsfSigma)
```
parejkoj, line 400: Why can't we make this be an `UpstreamFailureNoWorkFound`, and do it as the first step of this method? See DM-46948 for details, but this case is exactly what that exception was made for.

Not having a useable PSF on the diffim seems like a totally reasonable failure condition. I don't like using some default psf size, since that will result in unreliable measurements and we're not doing anything to mark the output catalog as being untrustworthy.

python/lsst/ip/diffim/detectAndMeasure.py
```diff
@@ -189,6 +189,18 @@ class DetectAndMeasureConfig(pipeBase.PipelineTaskConfig,
                  "base_PixelFlags_flag_saturatedCenterAll",
                  ),
     )
+    clearMaskPlanes = lsst.pex.config.ListField(
+        dtype=str,
+        doc="Mask planes to clear before running detection.",
+        default=("DETECTED", "DETECTED_NEGATIVE", "NOT_DEBLENDED", "STREAK"),
```
parejkoj, line 195: Just to check: these fields are set because they were propagated from the science image?

python/lsst/ip/diffim/detectAndMeasure.py
```diff
@@ -339,13 +351,8 @@ def run(self, science, matchedTemplate, difference,
...
 
-        # Ensure that we start with an empty detection and deblended mask.
-        mask = difference.mask
-        clearMaskPlanes = ["DETECTED", "DETECTED_NEGATIVE", "NOT_DEBLENDED", "STREAK"]
-        for mp in clearMaskPlanes:
-            if mp not in mask.getMaskPlaneDict():
-                mask.addMaskPlane(mp)
-        mask &= ~mask.getPlaneBitMask(clearMaskPlanes)
+        sigma = self._prepareInputs(difference)
+        self.log.info("Using PSF sigma of %f to smooth the difference image for detection.", sigma)
```
parejkoj, line 355: This new log is worth keeping, but I think it should go into `_prepareInputs`, since you forgot to put this log after the other place and there's no need to pass `sigma` to detection since it will get it from the (now guaranteed valid) psf.

python/lsst/ip/diffim/imageDecorrelation.py
```diff
@@ -274,7 +274,7 @@ def run(self, scienceExposure, templateExposure, subtractedExposure, psfMatching
             corr = self.computeDiffimCorrection(kArr, varianceMean, targetVarianceMean)
 
         correctedImage = self.computeCorrectedImage(corr.corrft, subtractedExposure.image.array)
-        correctedPsf = self.computeCorrectedDiffimPsf(corr.corrft, psfImg.array)
+        # correctedPsf = self.computeCorrectedDiffimPsf(corr.corrft, psfImg.array)
```
parejkoj, line 277: This, and the other commented line below, should either be deleted, or have a TODO attached to them with the ticket number where they will be restored (or deleted).

python/lsst/ip/diffim/detectAndMeasure.py
```diff
@@ -365,6 +373,33 @@ def run(self, science, matchedTemplate, difference,
...
+        Parameters
+        ----------
+        difference : `lsst.afw.image.ExposureF`
+            The difference image that will be used for detecting diaSources.
+            The mask plane will be modified in place.
+
+        Returns
+        -------
+        sigma : `float`
+            The width of the PSF of the difference image.
```
parejkoj, line 388: As noted elsewhere, I think you can remove this return if you just log that info message at the end of this method.

parejkoj (changes requested):
I'm not a fan of using a default psf size, since that could produce secretly bad results, without any obvious way to know that they are bad. `UpstreamFailureNoWorkFound` was made for this case.

parejkoj (commented):
Oh, and please add a test case with a garbage PSF, to check that the correct exception is raised.

Relatedly: where does the `psf==None` case get tested for and rejected?

python/lsst/ip/diffim/detectAndMeasure.py
isullivan: I purposely left it out of `_prepareInputs`, because that is used by both regular diffim and preconvolution where it is incorrect (since we don't smooth the difference image for detection then)

python/lsst/ip/diffim/detectAndMeasure.py
isullivan: I went back and forth on this, so I am fine checking for a valid PSF here and raising an error. I will look into `UpstreamFailureNoWorkFound`

python/lsst/ip/diffim/detectAndMeasure.py
isullivan: Correct

isullivan:
I removed the default PSF size and the related sigma logging, and instead raise an error if the width is invalid. There is a new unit test for that case, as well.

There are no unit tests of `subtractImages._applyExternalCalibrations`, which is where a test of the case of the external PSF being None would be relevant. Adding that test seems beyond the scope of this ticket.

python/lsst/ip/diffim/detectAndMeasure.py
```diff
@@ -365,6 +364,31 @@ def run(self, science, matchedTemplate, difference,
...
+        Parameters
+        ----------
+        difference : `lsst.afw.image.ExposureF`
+            The difference image that will be used for detecting diaSources.
+            The mask plane will be modified in place.
+
+        Raises
+        ------
+        pipeBase.UpstreamFailureNoWorkFound
+            If the PSF is not usable for measurement
```
parejkoj, line 379:
```suggestion
            If the PSF is not usable for measurement.
```

python/lsst/ip/diffim/detectAndMeasure.py
```diff
@@ -365,6 +364,31 @@ def run(self, science, matchedTemplate, difference,
...
+
+        Parameters
+        ----------
+        difference : `lsst.afw.image.ExposureF`
+            The difference image that will be used for detecting diaSources.
+            The mask plane will be modified in place.
+
+        Raises
+        ------
+        pipeBase.UpstreamFailureNoWorkFound
```
parejkoj, line 378: I think this has to be like this in docs?

```suggestion
        lsst.pipe.base.UpstreamFailureNoWorkFound
```

python/lsst/ip/diffim/detectAndMeasure.py
```diff
@@ -365,6 +364,31 @@ def run(self, science, matchedTemplate, difference,
...
+        # Ensure that we start with an empty detection and deblended mask.
+        mask = difference.mask
+        for mp in self.config.clearMaskPlanes:
+            if mp not in mask.getMaskPlaneDict():
+                mask.addMaskPlane(mp)
+        mask &= ~mask.getPlaneBitMask(self.config.clearMaskPlanes)
+        # Check that we have a valid PSF now before we do more work
+        sigma = difference.psf.computeShape(difference.psf.getAveragePosition()).getDeterminantRadius()
+        if np.isnan(sigma):
+            raise pipeBase.UpstreamFailureNoWorkFound("Invalid PSF detected! PSF width evaluates to NaN.")
```
parejkoj, line 390: I'd put this as the first line of the method, since there's no point in doing the rest if this is raised.

tests/test_detectAndMeasure.py
```diff
@@ -161,6 +164,35 @@ def test_detection_xy0(self):
...
+        psf = measAlg.KernelPsf(psfcK)
+        difference.setPsf(psf)
+
+        # Configure the detection Task
+        detectionTask = self._setup_detection()
+
+        # Run detection and check the results
+        with self.assertRaises(UpstreamFailureNoWorkFound):
+            detectionTask.run(science, matchedTemplate, difference,
+                              idFactory=self.idGenerator.make_table_id_factory())
```
parejkoj, line 194: No need to pass an idFactory unless you specifically care about the ids that are returned (which there won't be, since it raised).

tests/test_detectAndMeasure.py
```diff
@@ -161,6 +164,35 @@ def test_detection_xy0(self):
...
+        rng = np.random.RandomState(333)
+        kwargs = {"seed": staticSeed, "psfSize": 2.4, "fluxLevel": fluxLevel, "x0": 12345, "y0": 67890}
+        science, sources = makeTestImage(noiseLevel=noiseLevel, noiseSeed=6, **kwargs)
+        matchedTemplate, _ = makeTestImage(noiseLevel=noiseLevel/4, noiseSeed=7, **kwargs)
+        difference = science.clone()
+
+        psfShape = difference.getPsf().computeBBox(lsst.geom.Point2D(0, 0)).getDimensions()
+        psfcI = afwImage.ImageD(lsst.geom.Extent2I(psfShape[1], psfShape[0]))
+        psfNew = rng.randn(psfShape[1], psfShape[0])
+        psfcI.array = psfNew
```
parejkoj, line 183: Is this guaranteed to always make a psf that will fail the tests in the Task? Would setting all the pixels to zero or -1 be just as good, and more reliable? I know you set the seed above, but I don't think that's guaranteed to produce the same results across all architectures, is it?

parejkoj (approved):
Handful of new questions, but this is better, thanks.

tests/test_detectAndMeasure.py
isullivan: You can't build a kernel with all zeros, and setting to a constant value doesn't actually raise an error. Running several different noise realizations reveals that that sometimes does result in a kernel that has a valid sigma and doesn't raise. I will instead build a fixed kernel I know should raise.

tests/test_detectAndMeasure.py
```diff
@@ -161,6 +164,34 @@ def test_detection_xy0(self):
...
+        fluxLevel = 500
+        kwargs = {"seed": staticSeed, "psfSize": 2.4, "fluxLevel": fluxLevel, "x0": 12345, "y0": 67890}
+        science, sources = makeTestImage(noiseLevel=noiseLevel, noiseSeed=6, **kwargs)
+        matchedTemplate, _ = makeTestImage(noiseLevel=noiseLevel/4, noiseSeed=7, **kwargs)
+        difference = science.clone()
+
+        psfShape = difference.getPsf().computeBBox(lsst.geom.Point2D(0, 0)).getDimensions()
+        psfcI = afwImage.ImageD(lsst.geom.Extent2I(psfShape[1], psfShape[0]))
+        psfNew = np.zeros(psfShape)
+        psfNew[0, :] = 1
```
parejkoj, line 182: Oh, yes, that's nice and pathological!

PR #355, opened 2024-11-06
DM-47010: Drop obsoleted TODO comment.

TallJimbo:
GetTemplateTask in DRP now reads 'pvi', which already has the final calibrations.


arunkannawadi:
Coming from the `drp_pipe` PR - it might be worth adding a template parameter that can be used to switch between `calexp` and `pvi` in the `bbox` and `wcs` connections.

TallJimbo:
You may be right, but I'm grapping with higher-level questions about conventions for defining similarly-named connections on DM-47320 right now, and I'd prefer to wait until those are resolved before adding any new config options (since those take an RFC and deprecation period to get rid of).

arunkannawadi (approved):

PR #356, opened 2024-11-07
DM-47272: Move test-only utilities to separate file

isullivan:


BrunoSanchez (approved):
This simple migration of utils code to the `test` folder simplifies the imports and separates responsabilities. 
Looks good.

PR #362, opened 2024-11-15
DM-41701: Remove deprecated code

isullivan:
Removes NaiveCentroid and badSourceFlags


doc/lsst.ip.diffim/tasks/lsst.ip.diffim.DipoleMeasurementTask.rst
```diff
@@ -19,12 +19,8 @@ Their respective algorithm control classes are defined in
...
 measurement will have _neg (negative) and _pos (positive lobe) fields.
 
-The first set of measurements uses a "naive" algorithm for centroid and flux
-measurements, implemented in NaiveDipoleCentroidControl and
-NaiveDipoleFluxControl. The algorithm uses a naive 3x3 weighted moment around
-the nominal centroids of each peak in the Source Footprint.  These algorithms
-fill the table fields ip_diffim_NaiveDipoleCentroid* and
-ip_diffim_NaiveDipoleFlux*
+The first set of measurements uses the SDSS algorithm for centroid and flux
+measurements.
```
parejkoj, line 23: I think it's worth expanding this a bit, something like:
```suggestion
The first set of measurements uses the SDSS algorithm for centroid and flux
measurements.
This centroid will be used as a fallback, if the below dipole fit fails or if the source does not have a positive and negative peak to attempt a dipole fit to.
```

PR #363, opened 2024-11-20
DM-45361: Improve maskStreaks line fitting and mask plane behavior

mrawls:
This PR adjusts how maskStreaksTask is implemented in difference imaging. It uses a mask plane binning strategy to more readily detect fainter streaks, thereby "connecting the dots" in the DETECTED mask plane before looking for linear features.

### DIFF
--- a/python/lsst/ip/diffim/detectAndMeasure.py
+++ b/python/lsst/ip/diffim/detectAndMeasure.py
@@ -22,6 +22,8 @@
 import numpy as np
 
 import lsst.afw.detection as afwDetection
+import lsst.afw.image as afwImage
+import lsst.afw.math as afwMath
 import lsst.afw.table as afwTable
 import lsst.daf.base as dafBase
 import lsst.geom
@@ -168,6 +170,12 @@ class DetectAndMeasureConfig(pipeBase.PipelineTaskConfig,
         target=MaskStreaksTask,
         doc="Subtask for masking streaks. Only used if doMaskStreaks is True. "
             "Adds a mask plane to an exposure, with the mask plane name set by streakMaskName.",
+    )
+    streakBinFactor = pexConfig.Field(
+        dtype=int,
+        default=4,
+        doc="Bin scale factor to use when rerunning detection for masking streaks. "
+            "Only used if doMaskStreaks is True.",
     )
     writeStreakInfo = pexConfig.Field(
         dtype=bool,
@@ -444,7 +452,7 @@ class DetectAndMeasureTask(lsst.pipe.base.PipelineTask):
         self.metadata["nMergedDiaSources"] = len(initialDiaSources)
 
         if self.config.doMaskStreaks:
-            streakInfo = self._runStreakMasking(difference.maskedImage)
+            streakInfo = self._runStreakMasking(difference)
 
         if self.config.doSkySources:
             self.addSkySources(initialDiaSources, difference.mask, difference.info.id)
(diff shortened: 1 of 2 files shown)


python/lsst/ip/diffim/detectAndMeasure.py
```diff
@@ -22,6 +22,8 @@
 import numpy as np
 
 import lsst.afw.detection as afwDetection
+import lsst.afw.image as afwImage
+import lsst.afw.math as afwMath  # FOR TESTING BINNING
```
bsmartradio, line 26: Do we still need the clarifying comment since afwMath is necessary for making the binned image?

python/lsst/ip/diffim/detectAndMeasure.py
```diff
@@ -681,7 +695,31 @@ def _runStreakMasking(self, maskedImage):
...
         """
-        streaks = self.maskStreaks.run(maskedImage)
+        maskedImage = difference.maskedImage
+        # Bin the diffim to enhance low surface brightness streaks
+        binnedMaskedImage = afwMath.binImage(maskedImage,
+                                             self.config.streakBinFactor,
+                                             self.config.streakBinFactor)
+        binnedExposure = afwImage.ExposureF(binnedMaskedImage.getBBox())
+        binnedExposure.setMaskedImage(binnedMaskedImage)
+        binnedExposure.setPsf(difference.psf)  # exposure must have a PSF
```
bsmartradio, line 705: So I have been informed that there is a possibility that you can reach this point in the pipeline and an exposure might not have a PSF. How does that get handled in this situation? Does this automatically fail out gracefully or do we need to add some error handling just in case.

tests/test_detectAndMeasure.py
```diff
@@ -594,6 +597,35 @@ def test_mask_streaks(self):
...
+        output = detectionTask.run(science, matchedTemplate, difference)
+        outMask = output.subtractedMeasuredExposure.mask.array
+        streakMask = output.subtractedMeasuredExposure.mask.getPlaneBitMask("STREAK")
+        streakMaskSet = (outMask & streakMask) > 0
+        # Check that pixel values in streak_image equal the streak amplitude
+        streak_check = np.where(streak_image.array == amplitude)
+        self.assertTrue(np.all(streakMaskSet[streak_check]))
+        # Check that the entire image was not masked STREAK
+        self.assertFalse(np.all(streakMaskSet))
+
```
bsmartradio, line 628: If error handling is needed in the main code, then you should include a null test where things go wrong so you can check that the error is properly caught.

bsmartradio (approved):
Everything looks good but Colin informed me that there is a chance that you can reach ip_diffim with no psf, and I wasn't 100% sure how the code would fail out and if it would be graceful. If not, it might require some error handling for the weird edge case.

python/lsst/ip/diffim/detectAndMeasure.py
mrawls: Thank you for thinking of this — indeed this was a past problem, so I verified that the `_prepareInputs` call in the `run` method of detectAndMeasure will raise well before we get to streak masking if the PSF is un-compute-shape-able.

tests/test_detectAndMeasure.py
mrawls: Not relevant since we can never land here without a valid PSF

PR #364, opened 2024-11-22
DM-47680: Fix warnings about missing coordinate errors in unit tests

isullivan:


mrawls (approved):
Aha, THAT was where it needed to be! Thanks for finding and fixing 😄

PR #366, opened 2024-12-06
DM-48009: Revert #359 "Rename dipole classification "flags” and add attempted dipole…

isullivan:
… classification to APDB"

This reverts commit 3af77a3252e7600649c081fba370d1809b517e4c.


kfindeisen (approved):
It would be better if the PR description said it reverts #359, otherwise looks good.

PR #367, opened 2024-12-09
DM-47901: Configure streak masking for ComCam

isullivan:


python/lsst/ip/diffim/detectAndMeasure.py
```diff
@@ -702,10 +724,13 @@ def _runStreakMasking(self, difference):
...
         binnedExposure = afwImage.ExposureF(binnedMaskedImage.getBBox())
         binnedExposure.setMaskedImage(binnedMaskedImage)
-        binnedExposure.setPsf(difference.psf)  # exposure must have a PSF
+        # Clear the DETECTED mask plane before streak detection
+        binnedExposure.mask &= ~binnedExposure.mask.getPlaneBitMask('DETECTED')
         # Rerun detection to set the DETECTED mask plane on binnedExposure
-        _table = afwTable.SourceTable.make(self.schema)
-        self.detection.run(table=_table, exposure=binnedExposure, doSmooth=True)
+        sigma = difference.psf.computeShape(difference.psf.getAveragePosition()).getDeterminantRadius()
+        _table = afwTable.SourceTable.make(afwTable.SourceTable.makeMinimalSchema())
```
mrawls, line 731: Do you specifically not want the `self.schema` we used previously that had some extra columns floating around? I assumed it was more efficient to pass along an existing schema than making a new one from scratch, but it probably doesn't matter much either way.

python/lsst/ip/diffim/detectAndMeasure.py
```diff
@@ -217,6 +221,27 @@ def setDefaults(self):
...
+        # detected mask is set.
+        self.maskStreaks.onlyMaskDetected = False
+        # Restrict streak masking from growing too large
+        self.maskStreaks.maxStreakWidth = 100
+        # Restrict the number of iterations allowed for fitting streaks
+        # When the fit is good it should solve quickly, and exit a bad fit quickly
+        self.maskStreaks.maxFitIter = 10
+        # Only mask to 2 sigma in width
+        self.maskStreaks.nSigmaMask = 2
+        self.maskStreaks.absMinimumKernelHeight = 2
```
mrawls, line 243: Please add docstring for the `absMinimumKernelHeight` config setting

mrawls (approved):

python/lsst/ip/diffim/detectAndMeasure.py
isullivan: The minimal schema is trivial to create, and makes detection slightly faster since the resulting table has far fewer columns.

PR #368, opened 2024-12-12
DM-48106: Rename dipole classification "flag" and add attempted dipole classification

abudlong:


andy-slac (approved):
Looks good to me, with the same caveat that I'm not an expert on this code.

bsmartradio (approved):
Looks good

PR #369, opened 2024-12-24
DM-48256: Fix RunTimeError 'Record data is not contiguous in memory.' when running subtraction

mrawls:


python/lsst/ip/diffim/subtractImages.py
```diff
@@ -853,6 +853,10 @@ def _checkMask(mask, sources, excludeMaskPlanes, checkAdjacent=True):
         xv = (np.rint(sources.getX() - mask.getX0())).astype(int)
         yv = (np.rint(sources.getY() - mask.getY0())).astype(int)
 
+        # The source catalog may not be contiguous.
```
parejkoj, line 856: I don't think this comment is necessary: the lines below say the same thing.

python/lsst/ip/diffim/subtractImages.py
```diff
@@ -853,6 +853,10 @@ def _checkMask(mask, sources, excludeMaskPlanes, checkAdjacent=True):
...
+        if not sources.isContiguous():
```
parejkoj, line 857: I'm pretty sure this is too late to fix the problem, since the error report has it coming from the `xv=...` line above. When I said line 854, it must have been on a different version of the file. What I had meant was just after the call to `self.sourceSelector.selectSources` in `_sourceSelector()`. Except `_checkMask` is called on `sources`, not `selected`, so I don't see how it could fail in this way, unless the input catalog is broken somehow?

parejkoj (changes requested):
Could we get a unittest that confirms the bug is fixed? Because as coded, this wouldn't fix it (and I'm now not sure what the underlying bug truly is anyway!).

python/lsst/ip/diffim/subtractImages.py
mrawls: I put the putative fix in `_checkMask` because it operates on the source table in question, but you're probably right that this would best be handled earlier.

mrawls:
I'd like to add a unit test but I'm not sure how to make a non-contiguous source catalog on purpose. I think the test would best go [here](https://github.com/lsst/ip_diffim/blob/main/tests/test_subtractTask.py#L454); I can work on this tomorrow.

parejkoj:
I think we need to find out how this was triggered in this first place. I posted a followup question on the ticket, because it doesn't seem possible to have failed with a non-contiguous, unmodified catalog.

PR #377, opened 2025-02-07
DM-48807: Add detectAndMeasure negative peak tests

mrawls:
This PR adds tests for the bug fixed in DM-48704.


jsv1206 (approved):
The test looks good.

PR #378, opened 2025-02-11
DM-48879: Improve ip_diffim detectAndMeasure behavior around bright stars

ebellm:


python/lsst/ip/diffim/detectAndMeasure.py
```diff
@@ -215,10 +220,7 @@ def setDefaults(self):
         self.detection.reEstimateBackground = False
         self.detection.thresholdType = "pixel_stdev"
         self.detection.excludeMaskPlanes = ["EDGE",
-                                            "SAT",
                                             "BAD",
-                                            "INTRP",
-                                            "NO_DATA",
```
isullivan, line 221: Minor point, but please move the `excludeMaskPlanes` changes to their own commit

isullivan (approved):
I am surprised our unit tests pass with deblending off by default. I take it that is what John was referring to in group meeting this afternoon, about the missing commit from his deblending work last Spring.
My one request would be to move the mask plane config to a standalone commit.

PR #379, opened 2025-02-12
DM-48701: Fix incorrect footprint merging in diffim

parejkoj:


tests/test_detectAndMeasure.py
```diff
@@ -957,11 +972,71 @@ def testDeblendNegatives(self):
...
+        self.assertEqual(peaks[0].getIx(), 75)
+        self.assertEqual(peaks[0].getIy(), 124)
+
+        mask = np.isclose(result.diaSources["slot_Centroid_x"], 175) & \
+            np.isclose(result.diaSources["slot_Centroid_y"], 124)
+        self.assertEqual(mask.sum(), 1)
+        peaks = result.diaSources[mask][0].getFootprint().peaks
+        self.assertEqual(len(peaks), 1)
+        self.assertEqual(peaks[0].getIx(), 175)
+        self.assertEqual(peaks[0].getIy(), 124)
```
fred3m, line 1038: I'm not sure that this is the most useful test anymore. While it did catch the one specific bug in `mergeFootprintSets` it doesn't check for the general problem of having duplicate peaks. I would suggest something like
```
peak_x = np.array([peak['f_x'] for src in catalog[catalog["parent"]==0] for peak in src.getFootprint().peaks])
peak_y = np.array([peak['f_y'] for src in catalog[catalog["parent"]==0] for peak in src.getFootprint().peaks])
for i in range(len(peak_x)-1):
    for j in range(i, len(peak_x)):
        self.assertFalse((peak_x[i] == peak_x[j]) & (peak_y[i] == peak_y[j]))
```
Note: this code is untested and there is probably a way to vectorize it using `np.unique`. Something like (also untested)
```
coords = np.column_stack((peak_x, peak_y))
unique_coords, counts = np.unique(coords, axis=0, return_counts=True)
self.assertEqual(np.sum(counts > 1), 0)
```

fred3m (approved):
Looks good except that I made a request to make sure that the test looks for all duplicate peaks, not just the one from the mergeFootprintSet bug.

python/lsst/ip/diffim/detectAndMeasure.py
```diff
@@ -467,11 +471,22 @@ def processResults(self, science, matchedTemplate, difference, sources, idFactor
...
         if self.config.doMerge:
-            fpSet = positiveFootprints
-            fpSet.merge(negativeFootprints, self.config.growFootprint,
-                        self.config.growFootprint, False)
-            initialDiaSources = afwTable.SourceCatalog(self.schema)
-            fpSet.makeSources(initialDiaSources)
+            # preserve peak schema, if there are any footprints
+            if len(positives) > 0:
+                schema = positives[0].getFootprint().peaks.schema
+            elif len(positives) > 0:
```
ebellm, line 477:
```suggestion
            elif len(negatives) > 0:
```

ebellm (changes requested):
one small bug to fix

tests/test_detectAndMeasure.py
parejkoj: Thanks Fred, that was a nearly correct bit of code you produced there! I've added that and checked that it does fail (with one count==2) for the original bug condition, but left the original tests with a comment about why.

ebellm (approved):

PR #380, opened 2025-02-13
DM-47642: Update AL deconvolution to share code with convolution case.

isullivan:
As noted on the ticket (https://rubinobs.atlassian.net/browse/DM-47642), the original motivation for the choice of basis functions for the deconvolution kernel appears to have been a misinterpretation of the cited paper. The simpler approach gives comparable results, is easier to modify, and does not give the illusion of a mathematical foundation.

The new configs result in a better subtraction for one unit test that used deconvolution, so the test value needed to be updated.


BrunoSanchez (approved):
Looks good to me.

PR #382, opened 2025-02-19
DM-48354: Invalid diffim pixels in AP forced photometry

abudlong:


bsmartradio (approved):
In general this looks good, but there should be a new unit test added to `test_getTemplate.py` to ensure this is working as intended.

bsmartradio (changes requested):
Add a unit test to check changes. Sorry clicked the wrong review tick had to do this twice!

PR #384, opened 2025-02-26
DM-49071: Do not save zero length diaSource catalogs 

fred3m:


TallJimbo (approved):

python/lsst/ip/diffim/detectAndMeasure.py
```diff
@@ -521,8 +521,11 @@ def processResults(self, science, matchedTemplate, difference, sources, idFactor
 
         measurementResults = pipeBase.Struct(
             subtractedMeasuredExposure=difference,
-            diaSources=diaSources,
         )
+
+        if len(diaSources) > 0:
+            measurementResults.diaSources = diaSources
+
```
arunkannawadi, line 528: I missed the conversation happening in real time (I was dropping off someone at the airport) but my first thought here is that this is going to get cumbersome in the long run. But I see Jim has approved (thanks Jim) and I need to catch up on the thread before I have strong opinions.

PR #385, opened 2025-02-26
DM-49210: Set is_negative flag for negative-detected diffim sources

parejkoj:


tests/test_detectAndMeasure.py
```diff
@@ -977,32 +983,44 @@ def testDeblendNegatives(self):
...
+        As implemented, the diffim will have three positive sources (one a
+        blended pair), and 7 negative sources (two a blended pair).
         """
         # Make a science image with multiple blends, designed to trigger the
         # negative-footprint-related bug in `FootprintSet.merge`.
         bbox = lsst.geom.Box2I(lsst.geom.Point2I(0, 0), lsst.geom.Point2I(400, 200))
         dataset = lsst.meas.base.tests.TestDataset(bbox)
+        # positive isolated source on diffim
         dataset.addSource(instFlux=.5E5, centroid=lsst.geom.Point2D(25, 26))
+        # negative isolated source on diffim
```
nevencaplar, line 1001: I do not understand how this creates an isolated negative source - but this is not the part of this review, and  I can follow up later.

nevencaplar (approved):

tests/test_detectAndMeasure.py
parejkoj: Ah, I need to clarify a comment above this. These sources are added to the *template*, and thus will be negative on the *diffim* because the science image is empty. Two regions of the image are made negative on the template, and thus become positive sources on the diffim. I'll push a clarification.

parejkoj: And this definitely makes me wish we'd merged the cleanups I'd made on DM-46908, because that would have made this code a lot cleaner (I could have just made the inserted sources explicitly negative from the start).

parejkoj: See if this clarifies things for you: https://github.com/lsst/ip_diffim/pull/385/files#diff-27a812f8403a8dded0a333ce0ca81d2885eb83d57345d4bfd883f267b7dcb4e8L984

PR #388, opened 2025-03-13
DM-49378: Trim totalBox to destination warped bbox in getTemplate

parejkoj:


python/lsst/ip/diffim/getTemplate.py
```diff
@@ -277,25 +278,33 @@ def run(self, *, coaddExposureHandles, bbox, wcs, dataIds, physical_filter):
...
             del unwarped
-            if not np.any(np.isfinite(potentialInput.image.array)):
-                self.log.info("No overlap from coadds in tract %s; not including in output.", tract)
+            if np.all(potentialInput.mask.array & potentialInput.mask.getPlaneBitMask("NO_DATA")):
+                self.log.info("No overlap from coadd patches in tract %s; not including in output.", tract)
                 continue
 
             catalogs.append(catalog)
-            warped[tract] = potentialInput
-            warped[tract].setWcs(wcs)
```
isullivan, line 294: Was this line unnecessary, because it's set in `_merge`?

python/lsst/ip/diffim/getTemplate.py
```diff
@@ -277,25 +278,33 @@ def run(self, *, coaddExposureHandles, bbox, wcs, dataIds, physical_filter):
         for tract in coaddExposureHandles:
             maskedImages, catalog, totalBox = self._makeExposureCatalog(coaddExposureHandles[tract],
                                                                         dataIds[tract])
+            warpedBox = computeWarpedBBox(catalog[0].wcs, bbox, wcs)
+            warpedBox.grow(5)  # to ensure we catch all relevant input pixels
```
isullivan, line 282: My understanding of `computeWarpedBBox` is that that already errs on the side of a larger box, so growing it here is probably unnecessary. That said, an extra 5 pixels shouldn't make much of a difference, and it will still be fewer pixels than before.

isullivan (approved):
The code looks good. I don't have any requested changes.

python/lsst/ip/diffim/getTemplate.py
parejkoj: Yes, the merged result has the wcs when it is created.

python/lsst/ip/diffim/getTemplate.py
parejkoj: I found that without growing the box by a few pixels, the corners in the tests weren't correct, so I decided to be a bit generous here.

python/lsst/ip/diffim/getTemplate.py
```diff
@@ -297,12 +297,18 @@ def run(self, *, coaddExposureHandles, bbox, wcs, dataIds, physical_filter):
                 self.log.info("No overlap from coadd patches in tract %s; not including in output.", tract)
                 continue
 
-            catalogs.append(catalog)
+            # Trim the exposure catalog to just the patches that were used.
+            temp = afwTable.ExposureCatalog(self.schema)
```
isullivan, line 301: Please use something like `tempCatalog` instead of just `temp` for the variable name.

isullivan, line 301: Please use something like `tempCatalog` instead of just `temp` for the variable name.

PR #406, opened 2025-06-11
DM-51329: Use a slightly larger binsize for background subtraction

isullivan:
The changes on DM-50980 fixed a minor bug that had resulted in using larger bins for background subtraction.


leeskelvin (approved):
I don't have any opinions on the updated value of `binSize`, as I assume your testing has shown this to be optimal.

PR #408, opened 2025-06-12
DM-51368: Move kernel candidate residual metric to utils

isullivan:


python/lsst/ip/diffim/detectAndMeasure.py
```diff
@@ -995,7 +946,7 @@ class DetectAndMeasureScoreTask(DetectAndMeasureTask):
     _DefaultName = "detectAndMeasureScore"
 
     @timeMethod
-    def run(self, science, matchedTemplate, difference, scoreExposure,
+    def run(self, science, matchedTemplate, difference, scoreExposure, kernelSources,
```
fred3m, line 949: Update docstring

python/lsst/ip/diffim/utils.py
```diff
@@ -331,3 +334,83 @@ def evaluateMaskFraction(mask, maskPlane):
...
+        difference_footprints = np.zeros(n)
+        ratio = np.zeros(n)
+        for i, record in enumerate(sources):
+            footprint = record.getFootprint()
+            heavy = afwDetection.makeHeavyFootprint(footprint, science.maskedImage)
+            heavy_diff = afwDetection.makeHeavyFootprint(footprint, difference.maskedImage)
+            science_footprints[i] = abs(heavy.getImageArray()).mean()
+            difference_footprints[i] = abs(heavy_diff.getImageArray()).mean()
+            ratio[i] = abs((difference_footprints[i] - sky)/science_footprints[i])
+        return science_footprints, difference_footprints, ratio
```
fred3m, line 393: I know that this code was just copy pasted but looking at it I want to make sure that we know these sources are not blended (in either the science image or difference image). If it's guaranteed that the input sources are isolated then this code is fine. My main concern is that even if the difference image footprint is isolated, the science image footprint might not be, so I'm not sure how valid this comparison is. If this does need a fix I think that this is outside the scope of this ticket, but if there is a potential issue a ticket should probably be opened to investigate.

isullivan: The input catalog is just the output from `calibrateImage`, with additional selection based on the catalog and the science and template mask planes. The source selector is configured with these options: https://github.com/lsst/ip_diffim/blob/40d01f57dd6cb9c17e63eb5795442390ec25579e/python/lsst/ip/diffim/subtractImages.py#L250-L253
Does that seem sufficient to you, or would you still be concerned?

fred3m: I guess it depends on how this metric is used/interpreted. My thought is that even if the object is isolated in the difference image, it could still be blended with a non-variable source so the science footprint will be polluted in some cases. So my question is just "does this matter." If not then this i fine.

isullivan: That should be fine for the metric, at least. The kernel candidates used here are the sources we have chosen to be isolated and psf-like on the science image, and their footprints have been used to calculate the PSF matching kernel between the template and science images. These sources should be extremely well subtracted as long as everything is working well, so what this metric tries to capture is how much power remains in the difference image compared with the science image. Outliers are rejected during the kernel fit, which should remove any transients from the catalog, but if they are not removed we will probably have a bad fit overall.
Short answer: I think you are right to be concerned about the sources selected and used for psf matching, but once that selection is made the same set of sources should be used for this metric.

PR #417, opened 2025-07-04
DM-51632: LSSTCam DRP error in subtractImages: "Unable to determine kernel sum; 0 candidates"

BrunoSanchez:
Replace uninformative RuntimeError for custom AlgorithmError. 
Verify cells are not empty in PsfMatch


isullivan (approved):
Apologies for the delay in the review! I tested these changes and they are an improvement.

PR #428, opened 2025-08-20
DM-52258: Only run glint detection on good diaSources

isullivan:


mrawls (commented):
Thanks for this improvement - I agree catching glints after removing unphysical sources and making sure we aren't letting sky sources be part of a glint trail is what we want. Just one clarifying question I'd like to understand before hitting "approve."

PR #449, opened 2026-01-29
DM-53963: Calculate image/patch overlap in patch coordinates

cmsaunders:


erykoff (approved):

erykoff (commented):
It is probably worth adding a comment that we are projecting the detector into the patch (using the well defined patch WCS) which is more stable than projecting the patch into the detector coordinates (using the not-happy-to-be-extrapolated detector WCS)

PR #453, opened 2026-04-10
DM-54625: Update preconvolution with recent diffim improvements

isullivan:


python/lsst/ip/diffim/detectAndMeasure.py
```diff
@@ -1388,6 +1388,10 @@ def run(self, science, matchedTemplate, difference, scoreExposure, kernelSources
             Generator object used to assign ids to detected sources in the
             difference image. Ids from this generator are not set until after
             deblending and merging positive/negative peaks.
+        measurementResults : `lsst.pipe.base.Struct`, optional
+            Result struct that is modified to allow saving of partial outputs
+            for some failure conditions. If the task completes successfully,
+            this is also returned.
```
kfindeisen, line 1505: Is this the standard process for partial outputs now? I thought they were attached to the exception or metadata...

python/lsst/ip/diffim/makeKernel.py
```diff
@@ -208,8 +210,8 @@ def selectKernelSources(self, template, science, candidateList=None, preconvolve
                                                      fwhmExposureBuffer=self.config.fwhmExposureBuffer,
                                                      fwhmExposureGrid=self.config.fwhmExposureGrid
                                                      )
-        if preconvolved:
-            scienceFwhmPix *= np.sqrt(2)
+            if preconvolved:
+                scienceFwhmPix *= np.sqrt(2)
```
kfindeisen: There seems to be a lot of code duplication here (e.g., two `if (scienceFwhmPix is None) or (templateFwhmPix is None)` blocks). File a ticket to clean it up?

kfindeisen: For now, you probably should duplicate the comment, too.

python/lsst/ip/diffim/subtractImages.py
```diff
@@ -948,8 +955,9 @@ def _sourceSelector(self, template, science, sources, fallback=False, preconvolv
             selected = self.fallbackSourceSelector.selectSources(sources).selected
         else:
             selected = self.sourceSelector.selectSources(sources).selected
-        sciencePsfSize = self.sciencePsfSize*np.sqrt(2) if preconvolved else self.sciencePsfSize
-        kSize = self.makeKernel.makeKernelBasisList(self.templatePsfSize, sciencePsfSize)[0].getWidth()
+        if scienceFwhmPix is None:
+            scienceFwhmPix = self.sciencePsfSize*np.sqrt(2) if preconvolved else self.sciencePsfSize
```
kfindeisen: All these `if scienceFwhmPix is None` make me a bit nervous, because they're sensitive to the exact program logic -- if statements later get rearranged so that one of the tests happens after `scienceFwhmPix` is initialized, they'll give the wrong answer.

It might be safer to have a "measure FWHM" flag that's set at `run` start and guaranteed not to be modified after.

kfindeisen: It's also confusing that there's a `scienceFwhmPix` and a `sciencePsfSize` that, from context, is _also_ in pixels. Would a name like `kernelEffectiveFwhm` make them easier to tell apart? Or `kernelScale` (does it have to be a FWHM)?

python/lsst/ip/diffim/subtractImages.py
```diff
@@ -1229,15 +1237,24 @@ def run(self, template, science, sources, visitSummary=None):
...
+        # a matched-filter correlation rather than a convolution, which
+        # centers the score on the true source location even for asymmetric
+        # PSFs. The same reflected kernel is reused as ``preConvKernel`` in
+        # the decorrelation step so that both stages are self-consistent.
+        convolutionKernel = self._makePreconvolutionKernel(science.psf)
+        matchedScience = self._convolveExposure(science, convolutionKernel, self.convolutionControl,
                                                 interpolateBadMaskPlanes=True)
         self.metadata["convolvedExposure"] = "Preconvolution"
+
+        self.preconvolvedSciencePsfSize = self.sciencePsfSize*np.sqrt(2)
```
kfindeisen: Again, I wonder if there's a way to consolidate all the "scale up by sqrt(2) if preconvolving" logic. I understand that this particular code path calls `_sourceSelector` with `scienceFwhmPix` set, so the earlier lines don't apply, but it's a bit too 🍝 for my liking.

python/lsst/ip/diffim/subtractImages.py
```diff
@@ -1229,15 +1237,24 @@ def run(self, template, science, sources, visitSummary=None):
...
         self._prepareInputs(template, science, visitSummary=visitSummary)
 
-        # TODO: DM-37212 we need to mirror the kernel in order to get correct cross correlation
-        scienceKernel = science.psf.getKernel()
-        matchedScience = self._convolveExposure(science, scienceKernel, self.convolutionControl,
+        # Reflect the PSF kernel in both axes so the preconvolution implements
+        # a matched-filter correlation rather than a convolution, which
+        # centers the score on the true source location even for asymmetric
+        # PSFs. The same reflected kernel is reused as ``preConvKernel`` in
+        # the decorrelation step so that both stages are self-consistent.
```
kfindeisen: This comment is a bit verbose; I suggest cutting out the parts that are in the `_makePreconvolutionKernel` docs (which might well be all of it?)

python/lsst/ip/diffim/subtractImages.py
```diff
@@ -1251,6 +1268,35 @@ def run(self, template, science, sources, visitSummary=None):
...
+
+        Returns
+        -------
+        kernel : `~lsst.afw.math.FixedKernel`
+            The PSF reflected about both axes, normalized to sum to one.
+        """
+        avgPos = psf.getAveragePosition()
+        localKernel = psf.getLocalKernel(avgPos)
+        kimg = lsst.afw.image.ImageD(localKernel.getDimensions())
+        localKernel.computeImage(kimg, True)  # normalize to unit sum
```
kfindeisen: Can the `True` be a keyword argument instead? That would make it more self-documenting.

python/lsst/ip/diffim/subtractImages.py
```diff
@@ -1251,6 +1268,35 @@ def run(self, template, science, sources, visitSummary=None):
...
+
+        Parameters
+        ----------
+        psf : `~lsst.afw.detection.Psf`
+            The PSF to derive the preconvolution kernel from.
+
+        Returns
+        -------
+        kernel : `~lsst.afw.math.FixedKernel`
+            The PSF reflected about both axes, normalized to sum to one.
```
kfindeisen, line 1303: Should this actually be specced as `FixedKernel`, or should it just be `Kernel` and have the exact class be an implementation detail? Same with the reference in the extended description.

python/lsst/ip/diffim/subtractImages.py
```diff
@@ -1287,10 +1287,25 @@ def _makePreconvolutionKernel(psf):
...
             The PSF reflected about both axes, normalized to sum to one.
+
+        Raises
+        ------
+        RuntimeError
+            Raised if the PSF kernel has an even size along either axis.
+            The reflection ``[::-1, ::-1]`` only preserves the peak pixel
+            (and therefore the kernel center) for odd-sized kernels; an
+            even-sized kernel would be silently off-center by one pixel
+            after reflection and produce a misaligned matched filter.
```
kfindeisen: I don't think this much detail is actually necessary; you could just say "It's impossible to center an even-sized kernel" and leave out the exact implementation.

python/lsst/ip/diffim/subtractImages.py
```diff
@@ -1287,10 +1287,25 @@ def _makePreconvolutionKernel(psf):
...
         """
         avgPos = psf.getAveragePosition()
         localKernel = psf.getLocalKernel(avgPos)
-        kimg = lsst.afw.image.ImageD(localKernel.getDimensions())
+        dims = localKernel.getDimensions()
+        width, height = dims.getX(), dims.getY()
+        if width % 2 == 0 or height % 2 == 0:
+            raise RuntimeError(
+                f"Preconvolution requires an odd-sized PSF kernel, got {width}x{height}. "
+            )
```
kfindeisen, line 1317:
```suggestion
        dims = localKernel.getDimensions()
        if dims.x % 2 == 0 or dims.y % 2 == 0:
            raise RuntimeError(
                f"Preconvolution requires an odd-sized PSF kernel, got {dims.x}x{dims.y}. "
            )
```
I assume you specifically want the "x"? If you're fine with ordered pair notation, you could just print `dims`.

tests/test_detectAndMeasure.py
```diff
@@ -485,7 +485,7 @@ def test_detect_dipoles(self):
         offset = 1
         xSize = 300
         ySize = 300
-        kernelSize = 32
+        kernelSize = 31
```
kfindeisen, line 503: Add a test case for even (or semi-even) kernels?

python/lsst/ip/diffim/subtractImages.py
kfindeisen: I think you got it right on `makeTestImage` -- `ValueError` is the more appropriate exception for an invalid input.

python/lsst/ip/diffim/subtractImages.py
```diff
@@ -1268,6 +1268,35 @@ def run(self, template, science, sources, visitSummary=None):
 
         return subtractResults
 
+    @staticmethod
+    def _flagScoreEdge(exposure, innerBBox):
+        """Set the ``EDGE`` mask bit on pixels outside ``innerBBox``.
```
kfindeisen: Referring to parameters in brief descriptions [is discouraged](https://developer.lsst.io/python/numpydoc.html#on-api-and-parameter-names-in-summary-sentences). What about "Set the `EDGE` mask bit on pixels outside a known-valid region"?

python/lsst/ip/diffim/subtractImages.py
```diff
@@ -1268,6 +1268,35 @@ def run(self, template, science, sources, visitSummary=None):
...
+
+        Parameters
+        ----------
+        exposure : `~lsst.afw.image.Mask`
```
kfindeisen: From the code, it looks like this is an `Exposure`/`ExposureF` and not a `Mask`. (But maybe you _should_ take a mask? That would make this method more general.)

python/lsst/ip/diffim/subtractImages.py
```diff
@@ -1268,6 +1268,35 @@ def run(self, template, science, sources, visitSummary=None):
...
+        mask = exposure.mask
+        bbox = exposure.getBBox()
+        edgeBit = mask.getPlaneBitMask("EDGE")
+        dx0 = innerBBox.getMinX() - bbox.getMinX()
+        dx1 = bbox.getMaxX() - innerBBox.getMaxX()
+        dy0 = innerBBox.getMinY() - bbox.getMinY()
+        dy1 = bbox.getMaxY() - innerBBox.getMaxY()
+        if dy0 > 0:
+            mask.array[:dy0, :] |= edgeBit
+        if dy1 > 0:
```
kfindeisen, line 1278: It feels like there should be a cleaner way to get some kind of "difference of boxes", but I'll admit I don't see one.

However, how does this interact with the exposure/mask's XY0? `getBBox` [takes `PARENT` origin by default](http://doxygen.lsst.codes/stack/doxygen/x_mainDoxyDoc/classlsst_1_1afw_1_1image_1_1_image_base.html#a646fdfcf9586ed776b0c5fc8d70839c5), but there's no docs on what that actually means.

python/lsst/ip/diffim/detectAndMeasure.py
```diff
@@ -1407,17 +1407,49 @@ def run(self, science, matchedTemplate, difference, scoreExposure, kernelSources
         if idFactory is None:
             idFactory = lsst.meas.base.IdGenerator().make_table_id_factory()
 
-        self._prepareInputs(scoreExposure)
+        if self.config.doSubtractBackground:
+            # Background is measured on the difference image, and then
+            # subtracted from both the difference and score images.
+            # The preconvolution kernel is normalized to 1, so the
+            # same background level applies to both.
```
kfindeisen: I think this comment is more confusing than helpful, since it doesn't seem to correspond to the code below it. The way I see it:
- background is measured on the difference image
- it is *not* subtracted from the difference image (that's deferred to the second pass)
- it's temporarily subtracted from the score image

tests/test_detectAndMeasure.py
```diff
@@ -1197,6 +1197,69 @@ def _detection_wrapper(setFlags=True):
...
+        staticSeed = 1
+        transientSeed = 6
+        fluxLevel = 500
+        xSize = 512
+        ySize = 512
+        x0 = 123
+        y0 = 456
+        kwargs = {"seed": staticSeed, "psfSize": 2.4, "fluxLevel": fluxLevel,
+                  "xSize": xSize, "ySize": ySize, "x0": x0, "y0": y0}
+        params = [2.2, 2.1, 2.0, 1.2, 1.1, 1.0]
```
kfindeisen: I suggest calling this `chebyshev_params` to be less confusing.

tests/test_detectAndMeasure.py
```diff
@@ -1197,6 +1197,69 @@ def _detection_wrapper(setFlags=True):
...
+                                             background=background_model, **kwargs)
+        matchedTemplate, _ = makeTestImage(noiseLevel=noiseLevel/4, noiseSeed=7, **kwargs)
+        subtractTask = subtractImages.AlardLuptonPreconvolveSubtractTask()
+        scienceKernel = scienceBase.psf.getKernel()
+
+        # Configure the detection Task
+        detectionTask = self._setup_detection(doMerge=False, doSubtractBackground=True)
+        kwargs["seed"] = transientSeed
+        kwargs["nSrc"] = 10
+        kwargs["fluxLevel"] = 1000
```
kfindeisen: Move the magic numbers up to the main block (especially `fluxLevel`)?

tests/test_detectAndMeasure.py
```diff
@@ -1197,6 +1197,69 @@ def _detection_wrapper(setFlags=True):
...
+
+        def _detection_wrapper(positive=True):
```
kfindeisen: I suggest a more active and descriptive name like `_check_detections`.

Nested functions are hard to read, especially when they're long. Do you actually benefit substantially from having a nested definition instead of just passing the images etc. into a separate method? Or from having a generic method that does *all* the work (including image setup)? I'm pretty sure the latter would also keep you from having to make so many defensive copies, because each image would only be used once.

tests/test_detectAndMeasure.py
```diff
@@ -1197,6 +1197,69 @@ def _detection_wrapper(setFlags=True):
...
+                science.maskedImage -= transients.maskedImage
+            difference = science.clone()
+            difference.maskedImage -= matchedTemplate.maskedImage
+            score = subtractTask._convolveExposure(difference, scienceKernel,
+                                                   subtractTask.convolutionControl)
+            # Keep a copy of the score image before detection so we can verify
+            # that the background was actually subtracted from it.
+            originalScoreMedian = np.median(score.image.array[np.isfinite(score.image.array)])
+            originalDiffimMedian = np.median(difference.image.array[np.isfinite(difference.image.array)])
+            output = detectionTask.run(science, matchedTemplate, difference, score, sources)
```
kfindeisen: You're not actually keeping a copy of the whole image like the comment says?

tests/test_detectAndMeasure.py
```diff
@@ -1197,6 +1197,69 @@ def _detection_wrapper(setFlags=True):
...
+
+            # The background subtraction should leave a non-empty BackgroundList.
+            self.assertGreater(len(output.differenceBackground), 0)
+            # The difference image should have been modified in place by the
+            # background subtraction, so its median should shift noticeably.
+            subtractedScoreMedian = np.median(score.image.array[np.isfinite(score.image.array)])
+            self.assertNotAlmostEqual(originalScoreMedian, subtractedScoreMedian, places=3)
```
kfindeisen: Haven't seen `NotAlmostEqual` before. That's almost not too many modifiers... 😵‍💫 

Could you make this assertion stronger and have an `assertLess`? Or can the test produce negative backgrounds?

python/lsst/ip/diffim/detectAndMeasure.py
kfindeisen: This seems unnecessarily convoluted. Let me see if I understand the control flow here:
* With background subtraction, detection is run without cosmic ray masking on a coarse score image, then the background is recalculated, then detection is run again with cosmic ray masking and the final score image.
* Without background subtraction, what would have been the predetection pass is now the main one, run with cosmic ray masking and the only score image.

If it's really necessary to do cosmic ray masking exactly once, then it seems more reasonable to make (new) lines 1449-1459 the baseline, and make lines 1430-1447 conditional on background subtraction. Then lines 1424-1428 would be unnecessary, and 1454-1459 is always the final detection and can be thought of as such.

If that won't work, maybe this is a sign that the two workflows are different enough that they should be represented by different methods (with shared blocks factored out, of course).

tests/test_detectAndMeasure.py
```diff
@@ -1261,6 +1261,53 @@ def _detection_wrapper(positive=True):
...
+
+        # Test that no CRs are detected when none are present.
+        transients, _ = makeTestImage(noiseLevel=noiseLevel, noiseSeed=8, nSrc=10, **kwargs)
+        science.maskedImage += transients.maskedImage
+        difference = science.clone()
+        difference.maskedImage -= matchedTemplate.maskedImage
+        score = subtractTask._convolveExposure(difference, scienceKernel, subtractTask.convolutionControl)
+        output = detectionTask.run(science, matchedTemplate, difference, score, sources)
+        crMaskSet = (output.subtractedMeasuredExposure.mask.array & crMask) > 0
+        self.assertTrue(np.all(crMaskSet == 0))
```
kfindeisen: I'm not familiar with numpy, but it seems like these could be merged?
```suggestion
        self.assertFalse(np.any((output.subtractedMeasuredExposure.mask.array & crMask) > 0))
```

tests/test_detectAndMeasure.py
```diff
@@ -1261,6 +1261,53 @@ def _detection_wrapper(positive=True):
...
+        # Inject CR-like shapes into the difference image. CR detection runs
+        # on the difference, and the mask should propagate to the measured
+        # exposure returned by the task.
+        difference.image.array[crX0:crX0+1, crY0:crY0+5] += 1234
+        difference.image.array[crX1:crX1+5, crY1:crY1+1] += 1234
+        score = subtractTask._convolveExposure(difference, scienceKernel, subtractTask.convolutionControl)
+        output = detectionTask.run(science, matchedTemplate, difference, score, sources)
+        crMaskSet = (output.subtractedMeasuredExposure.mask.array & crMask) > 0
+        self.assertTrue(np.all(crMaskSet[crX0:crX0+1, crY0:crY0+5]))
+        self.assertTrue(np.all(crMaskSet[crX1:crX1+5, crY1:crY1+1]))
```
kfindeisen: I assume it's unsafe to test that the mask is still unset outside these boxes?

tests/test_subtractTask.py
kfindeisen, line 1: I think the commit message "Reduce PSF size in tests to keep PSF fully contained in kernel size" might require a bit more explanation. What is wrong with testing PSF > kernel?

Would it be possible to keep at least one test with large PSFs? I assume the code needs to be able to handle this case in the real world.

tests/test_subtractTask.py
```diff
@@ -155,7 +155,7 @@ def _run_and_check_coverage(template_coverage,
 
     def test_clear_template_mask(self):
         noiseLevel = 1.
-        science, sources = makeTestImage(psfSize=3.0, noiseLevel=noiseLevel, noiseSeed=6)
+        science, sources = makeTestImage(psfSize=2.4, noiseLevel=noiseLevel, noiseSeed=6)
```
kfindeisen: Any chance these 16-18 instances of `psfSize=2.4` could be merged? At the very least, this seems like something you could put in the magic number blocks.

tests/test_subtractTask.py
```diff
@@ -338,7 +338,7 @@ def test_science_better(self):
 
         def _run_and_check_images(statsCtrl, statsCtrlDetect, scienceNoiseLevel, templateNoiseLevel):
             science, sources = makeTestImage(psfSize=2.0, noiseLevel=scienceNoiseLevel, noiseSeed=6)
-            template, _ = makeTestImage(psfSize=3.0, noiseLevel=templateNoiseLevel, noiseSeed=7,
+            template, _ = makeTestImage(psfSize=2.3, noiseLevel=templateNoiseLevel, noiseSeed=7,
```
kfindeisen: Typo (probably doesn't matter)?

python/lsst/ip/diffim/makeKernelBasisList.py
```diff
@@ -291,11 +291,14 @@ def _calculateBasisSigmas(referenceSigma, targetSigma, basisMinSigma, basisGauss
...
-    #  factor of basisGaussBeta larger than the minimum sigma, use
-    #  kernelSigma/basisGaussBeta as the size of the first gaussian.
-    if ((kernelSigma/basisGaussBeta) > basisMinSigma) & (basisNGauss > 1):
-        basisSigmaGauss = [kernelSigma/basisGaussBeta, ]
+    # The smallest basis Gaussian sigma is kernelSigma/basisGaussBeta to center
+    # the geometric basis spacing around kernelSigma, but bounded below by
+    # basisMinSigma. Using max() here (rather than a kernelSigma-vs-threshold
+    # branch) keeps the basis continuous as kernelSigma crosses
+    # basisMinSigma*basisGaussBeta, avoiding a discontinuous jump in the AL
+    # basis that produced unstable kernels.
```
kfindeisen: I find this comment a bit hard to follow. In particular, the "rather than a kernelSigma-vs-threshold branch" makes sense only in the context of the deleted line 297.

I also can't really parse "The smallest basis Gaussian sigma is kernelSigma/basisGaussBeta to center the geometric basis spacing around kernelSigma", but maybe that makes more sense to an expert?

python/lsst/ip/diffim/psfMatch.py
```diff
@@ -595,7 +595,12 @@ def _diagnostic(self, kernelCellSet, spatialSolution, spatialKernel, spatialBg):
         """
         # What is the final kernel sum
         kImage = afwImage.ImageD(spatialKernel.getDimensions())
-        kSum = spatialKernel.computeImage(kImage, False)
+        # Evaluate the kernel at the cell-set center, not at the world
+        # origin (the default (0, 0) lies outside the cellset bbox for any
+        # non-origin region and produces a meaningless kernel sum).
```
kfindeisen: I think you can get rid of the bit in parentheses -- it should be obvious that distant coordinates are going to give you unreliable results!

python/lsst/ip/diffim/psfMatch.py
```diff
@@ -595,7 +595,12 @@ def _diagnostic(self, kernelCellSet, spatialSolution, spatialKernel, spatialBg):
...
+        regionBBox = kernelCellSet.getBBox()
+        xcen, ycen = regionBBox.getCenter()
```
kfindeisen: To save you one extra variable to keep track of:
```suggestion
        xcen, ycen = kernelCellSet.getBBox().getCenter()
```

python/lsst/ip/diffim/psfMatch.py
```diff
@@ -595,7 +595,12 @@ def _diagnostic(self, kernelCellSet, spatialSolution, spatialKernel, spatialBg):
...
+        kSum = spatialKernel.computeImage(kImage, False, xcen, ycen)
```
kfindeisen: Can you add a keyword for the `False`, so that it's clearer what it is?

python/lsst/ip/diffim/subtractImages.py
```diff
@@ -708,7 +708,8 @@ def runConvolveScience(self, template, science, psfMatchingKernel, backgroundMod
         bbox = science.getBBox()
 
         kernelImage = lsst.afw.image.ImageD(psfMatchingKernel.getDimensions())
-        norm = psfMatchingKernel.computeImage(kernelImage, doNormalize=False)
+        xcen, ycen = bbox.getCenter()
+        norm = psfMatchingKernel.computeImage(kernelImage, False, xcen, ycen)
```
kfindeisen: I recommend keeping the `doNormalize=`, it makes the code easier to understand.

python/lsst/ip/diffim/subtractImages.py
```diff
@@ -721,9 +721,10 @@ def runConvolveScience(self, template, science, psfMatchingKernel, backgroundMod
         matchedTemplate.setPhotoCalib(science.photoCalib)
 
         if backgroundModel is not None:
-            modelParams = backgroundModel.getParameters()
             # We must invert the background model if the matching kernel is solved for the science image.
-            backgroundModel.setParameters([-p for p in modelParams])
+            invertedBackground = backgroundModel.clone()
+            invertedBackground.setParameters([-p for p in backgroundModel.getParameters()])
+            backgroundModel = invertedBackground
```
kfindeisen, line 729: I assume this is for exception safety? If so, that might be worth a brief comment (the technique is called "copy-and-swap", but only a C++ programmer is likely to recognize the term).

python/lsst/ip/diffim/subtractImages.py
```diff
@@ -809,29 +809,23 @@ def _calculateMagLim(self, exposure, nsigma=5.0, fallbackPsfSize=None):
...
-                psf_area = np.pi*(fallbackPsfSize/2)**2
-                zeropoint = exposure.photoCalib.instFluxToMagnitude(1)
-                maglim = zeropoint - 2.5*np.log10(nsigma*np.sqrt(psf_area))
-            else:
+            if fallbackPsfSize is None:
                 self.log.info("Unable to evaluate PSF, setting maglim to nan")
-                maglim = np.nan
+                return np.nan
+            self.log.info("Unable to evaluate PSF, using fallback FWHM %f", fallbackPsfSize)
+            psf_area = np.pi*(fallbackPsfSize/2)**2
```
kfindeisen, line 824: While this code is certainly cleaner (and 👍 from that standpoint), it also seems to be equivalent. In which case was the exception being "silently squashed" before but not now?

python/lsst/ip/diffim/utils.py
kfindeisen, line 1: Instead of "Fix docstring", how about "Add more details to setSourceFootprints docstring"? While it's certainly an improvement, I'm not sure it counts as a fix.

python/lsst/ip/diffim/utils.py
```diff
@@ -502,17 +502,23 @@ def checkMask(mask, sources, excludeMaskPlanes):
...
-        The "radius" of the footprint, i.e half the size of the bounding box.
+        Width (in pixels) of the PSF-matching kernel. The footprint side is
+        ``2*kernelSize + 1`` so that the kernel can be convolved across the
+        source without falling off the stamp.
 
     Returns
     -------
     sources : `lsst.afw.table.SourceCatalog`
-        The modified source catalog
+        The same (modified) source catalog, returned for convenience.
```
kfindeisen: For clarity, I suggest referring to `sources` explicitly.

python/lsst/ip/diffim/utils.py
```diff
@@ -502,17 +502,23 @@ def checkMask(mask, sources, excludeMaskPlanes):
...
+    of the original footprint.
+
     Parameters
     ----------
     sources : `lsst.afw.table.SourceCatalog`
-        The source catalog to add footprints to.
+        The source catalog to add footprints to. Modified in place.
     kernelSize : `int`
-        The "radius" of the footprint, i.e half the size of the bounding box.
+        Width (in pixels) of the PSF-matching kernel. The footprint side is
```
kfindeisen:
```suggestion
        Width (in pixels) of the PSF-matching kernel. The footprint size is
```

python/lsst/ip/diffim/detectAndMeasure.py
kfindeisen, line 1457: This doesn't seem right -- `*Diff_scoreExp` used to be an input, but is now an output? That would be a breaking change to the pipeline structure.

Also, the two connections have identical docs, making it hard to understand what the difference is.

python/lsst/ip/diffim/detectAndMeasure.py
```diff
@@ -346,6 +346,19 @@ class DetectAndMeasureConfig(pipeBase.PipelineTaskConfig,
...
+        default=True,
+        doc="Reject any peaks detected on ``badMaskPlanes`` before measurement."
+            "These should all be rejected downstream by ``badSourceFlags``, "
+            "but filtering earlier can save time."
+    )
+    badMaskPlanes = lsst.pex.config.ListField(
+        dtype=str,
+        doc="Detections whose footprint peak lies on a pixel with any of these "
+            "mask planes set will be rejected before measurement.",
+        default=("NO_DATA", "BAD", "SAT", "EDGE"),
```
kfindeisen, line 361: Logically, this is a set (a plane either is used for rejection, or it isn't). Physically, a `ListField` should probably be a list (so that you can e.g. append to the defaults), and that's what you'd get from YAML parsing anyway. Why make it a tuple (i.e., generalized parameter list)?

python/lsst/ip/diffim/detectAndMeasure.py
```diff
@@ -940,6 +958,55 @@ def deblend(footprints, negative=False):
...
+    def _rejectBadMaskedDetections(self, initialDiaSources, mask):
+        """Remove detections whose footprint peak lies on a bad-masked pixel.
+
+        Detections are rejected if any peak of the source footprint falls on a
+        pixel with any of the ``config.badMaskPlanes`` bits set.
+
+        Parameters
+        ----------
+        initialDiaSources : `lsst.afw.table.SourceCatalog`
+            The catalog of detected sources, before measurement.
```
kfindeisen: I don't think the "before measurement" is relevant here -- these are just detections to be filtered.

python/lsst/ip/diffim/detectAndMeasure.py
kfindeisen: Should you document that missing planes are silently ignored, or is that the existing convention for mask usage?

python/lsst/ip/diffim/detectAndMeasure.py
```diff
@@ -940,6 +958,55 @@ def deblend(footprints, negative=False):
...
+
+        presentPlanes = [mp for mp in self.config.badMaskPlanes
+                         if mp in mask.getMaskPlaneDict()]
+        if not presentPlanes:
+            self.metadata["nRejectedBadMaskPlanes"] = 0
+            return initialDiaSources
+
+        badBitMask = mask.getPlaneBitMask(presentPlanes)
+        x0 = mask.getX0()
+        y0 = mask.getY0()
```
kfindeisen: Would `x0, y0 = getXY0()` work here? I think the Python interface for `Point2I` is iterable.

python/lsst/ip/diffim/detectAndMeasure.py
```diff
@@ -940,6 +958,55 @@ def deblend(footprints, negative=False):
...
+                iy = peak.getIy() - y0
+                if mask.array[iy, ix] & badBitMask:
+                    selector[i] = False
+                    break
+        nRejected = np.count_nonzero(~selector)
+        self.metadata["nRejectedBadMaskPlanes"] = nRejected
+        if nRejected > 0:
+            self.log.info("Rejected %d detections on pixels with bad mask "
+                          "planes %s before measurement.",
+                          nRejected, list(presentPlanes))
```
kfindeisen: `presentPlanes` is already a list, by construction.

tests/test_detectAndMeasure.py
```diff
@@ -1210,11 +1213,18 @@ def test_detect_transients_with_background(self):
         ySize = 512
         x0 = 123
         y0 = 456
+        kernelSize = 31
+        templateBorderSize = kernelSize//2
```
kfindeisen, line 1251: Again, if you're going to have a comment for one copy of the code, you probably need it everywhere. (Can the copies be factored out?)

tests/test_subtractTask.py
kfindeisen, line 1: I suggest merging "Use smaller template coverage" with "Reduce PSF size in tests", or at least moving it to immediately after.

kfindeisen (changes requested):
As forewarned, lots of comments. Some of the factoring I'm asking for is pre-existing and should probably be deferred to a later work item, but I suggest at least streamlining the "when to do CR removal vs. background removal" logic. And I really don't understand how the `Diff_scoreExp` vs. `Diff_scoreTempExp` swap doesn't break things.

python/lsst/ip/diffim/detectAndMeasure.py
isullivan: Yikes! Good catch - since we overwrite these defaults in the pipelines, I didn't get any errors in testing from this copy-paste bug. I'll fix it!

python/lsst/ip/diffim/detectAndMeasure.py
isullivan: Yes, the metadata is used for writing the custom exception message, but we have to define a struct like `measurementResults` that can be modified in place so that any partial outputs can be passed out.

python/lsst/ip/diffim/makeKernel.py
isullivan: This was an easy refactor, so I made a helper function that could be used in both places.

python/lsst/ip/diffim/makeKernelBasisList.py
isullivan: I rephrased and condensed the comment.

python/lsst/ip/diffim/subtractImages.py
isullivan: The bug was including a `return` statement inside a `finally` block, which I only recently learned about. In Python, that silently squashes any exception that was not explicitly caught.

python/lsst/ip/diffim/subtractImages.py
isullivan: This started out taking a `Mask`, but then I changed it to `Exposure` so I could grab the `bbox`. I'll update the docstring.

python/lsst/ip/diffim/subtractImages.py
isullivan: I have added a `usePreconvolution` flag and `self.matchedPsfSize` so that these sizes can be set up at the start and then used consistently.

tests/test_detectAndMeasure.py
isullivan: Sure. Since the raise is in `makeTestImage` I added the test to `test_utils.py`

tests/test_detectAndMeasure.py
isullivan: I'm happy to change the name to `_check_detections` everywhere that is appropriate, but I don't think I am up to refactoring all of the unit tests to use a generic method.

tests/test_detectAndMeasure.py
isullivan: Negative backgrounds are perfectly reasonable, especially on a difference image. I can use `assertLess` with the absolute value, though.

python/lsst/ip/diffim/subtractImages.py
isullivan: I've re-written everything to set `self.matchedPsfSize` at the start, which eliminates all of these tests later.

tests/test_detectAndMeasure.py
isullivan: That is a good check. I re-wrote the two CR-masking tests to reduce code duplication of the pixel ranges and add the test you suggested that the CR mask is not set outside of the CR-injected pixels.

python/lsst/ip/diffim/subtractImages.py
isullivan: No, this is not needed in Python. I just find it easier to read this way, and it's a small set of parameters so it's not a big loss in efficiency.

tests/test_subtractTask.py
isullivan: The real code scales the size of the matching kernel up when needed, so it should not hit the condition that was problematic in the test. That is probably worth adding a unit test to the tests of `makeKernel.py` to make sure that is really the case, but I think that is beyond the scope of this ticket.

python/lsst/ip/diffim/detectAndMeasure.py
isullivan: What the comment is trying to explain is that the background is measured and subtracted from the difference image, and because the preconvolution kernel should not affect the background, that same background is also subtracted from the score image. I would welcome any suggestions you have to make the wording of the comment more clear.

isullivan:
This is ready for another round of review. The commits that start "fixup" are intended to be squashed into the referenced commit. I left them pre-rebasing so that the changes are visible.

python/lsst/ip/diffim/subtractImages.py
kfindeisen: [`Mask`](http://doxygen.lsst.codes/stack/doxygen/x_mainDoxyDoc/classlsst_1_1afw_1_1image_1_1_mask.html) has the same `getBBox` method -- all the "image-like" classes do.

python/lsst/ip/diffim/subtractImages.py
kfindeisen: Oops, missed that part. Yeah, I can see how that would be disruptive (in any language, not just Python). 😵‍💫

python/lsst/ip/diffim/makeKernel.py
```diff
@@ -155,9 +155,11 @@ def run(self, template, science, kernelSources, preconvolved=False,
...
                                                      fwhmExposureGrid=self.config.fwhmExposureGrid
                                                      )
-
-        if preconvolved:
-            scienceFwhmPix *= np.sqrt(2)
+            # Apply the Gaussian approximation FWHM boost only when
+            # the size was auto-computed. An explicit ``scienceFwhmPix`` is
+            # assumed already to reflect the preconvolved effective width.
+            if preconvolved:
+                scienceFwhmPix *= np.sqrt(2)
```
kfindeisen: This code runs if `scienceFwhmPix` was passed in but `templateFwhmPix` was `None`. Was that the intent? Same with the indented block on lines 213-214 and the version moved into `_checkPsfWidths`.

python/lsst/ip/diffim/detectAndMeasure.py
```diff
@@ -1407,17 +1407,49 @@ def run(self, science, matchedTemplate, difference, scoreExposure, kernelSources
...
 
-        self._prepareInputs(scoreExposure)
+        if self.config.doSubtractBackground:
+            # Background is measured on the difference image, and then
+            # subtracted from both the difference and score images.
+            # The preconvolution kernel is normalized to 1, so the
+            # same background level applies to both.
+            background = self.subtractInitialBackground.run(difference.clone()).background
+            detectionScoreExposure = scoreExposure.clone()
+            detectionScoreExposure.image -= background.getImage()
```
kfindeisen: I still think it's confusing to talk about subtracting the background from the difference image when you're going out of your way to not do that. Maybe just have the last two lines (with "both difference and score images")?

tests/test_detectAndMeasure.py
```diff
@@ -1261,6 +1263,54 @@ def _detection_wrapper(positive=True):
         _detection_wrapper(positive=True)
         _detection_wrapper(positive=False)
 
+    def test_mask_cosmic_rays(self):
```
kfindeisen, line 1308: Redefinition of method from line 417.

python/lsst/ip/diffim/subtractImages.py
kfindeisen: Could you expand the commit message to mention that this is for legibility? Right now it just says "Clone the background model before inverting", which implies that there's some obvious(?) reason why you have to do that (e.g., a bug fix).

python/lsst/ip/diffim/subtractImages.py
kfindeisen: Could you mention the return-from-finally in the commit message, so that it's clearer what's going on?

python/lsst/ip/diffim/detectAndMeasure.py
```diff
@@ -1342,6 +1342,12 @@ class DetectAndMeasureScoreConnections(DetectAndMeasureConnections):
...
         dimensions=("instrument", "visit", "detector"),
         storageClass="ExposureF",
+        name="{fakesType}{coaddName}Diff_scoreTempExp",
+    )
+    scoreMeasuredExposure = pipeBase.connectionTypes.Output(
+        doc="Score image after backgrond subtraction and detection.",
+        dimensions=("instrument", "visit", "detector"),
+        storageClass="ExposureF",
         name="{fakesType}{coaddName}Diff_scoreExp",
     )
```
kfindeisen, line 1457: This is still turning `*Diff_scoreExp` from an input to an output.

tests/test_subtractTask.py
```diff
@@ -40,6 +40,10 @@
 
 
 class AlardLuptonSubtractTestBase:
+    goodPsfSize = 2.0
+    midPsfSize = 2.4
+    badPsfSize = 2.8
+
```
kfindeisen, line 46: Thanks for this -- I missed that there were different sizes the first time around!

python/lsst/ip/diffim/psfMatch.py
```diff
@@ -595,12 +595,9 @@ def _diagnostic(self, kernelCellSet, spatialSolution, spatialKernel, spatialBg):
...
         kImage = afwImage.ImageD(spatialKernel.getDimensions())
-        # Evaluate the kernel at the cell-set center, not at the world
-        # origin (the default (0, 0) lies outside the cellset bbox for any
-        # non-origin region and produces a meaningless kernel sum).
-        regionBBox = kernelCellSet.getBBox()
-        xcen, ycen = regionBBox.getCenter()
-        kSum = spatialKernel.computeImage(kImage, False, xcen, ycen)
+        # Evaluate the kernel at the cell-set center, not at the world origin.
+        xcen, ycen = kernelCellSet.getBBox().getCenter()
+        kSum = spatialKernel.computeImage(kImage, doNormalize=False, x=xcen, y=ycen)
```
kfindeisen, line 600: I suggest merging these changes with the commits that added these calls in the first place.

python/lsst/ip/diffim/makeKernel.py
```diff
@@ -133,33 +136,9 @@ def run(self, template, science, kernelSources, preconvolved=False,
...
-                                                     fwhmExposureGrid=self.config.fwhmExposureGrid
-                                                     )
-            # Apply the Gaussian approximation FWHM boost only when
-            # the size was auto-computed. An explicit ``scienceFwhmPix`` is
-            # assumed already to reflect the preconvolved effective width.
-            if preconvolved:
-                scienceFwhmPix *= np.sqrt(2)
+        templateFwhmPix, scienceFwhmPix = self._checkPsfWidths(template, science,
+                                                               templateFwhmPix, scienceFwhmPix,
+                                                               preconvolved=False)
```
kfindeisen: This is setting `preconvolved=False` regardless of the value of `preconvolved` that was passed to `run`. Did you mean `preconvolved=preconvolved`? Same with the call below.

python/lsst/ip/diffim/makeKernel.py
```diff
@@ -122,6 +122,9 @@ def run(self, template, science, kernelSources, preconvolved=False,
             Typically the output of `MakeKernelTask.selectKernelSources`.
         preconvolved : `bool`, optional
             Was the science image convolved with its own PSF?
+        templateFwhmPix, scienceFwhmPix : `float`, optional
```
kfindeisen:
```suggestion
        templateFwhmPix, scienceFwhmPix : `float` or `None`, optional
```

python/lsst/ip/diffim/makeKernel.py
```diff
@@ -182,36 +161,18 @@ def selectKernelSources(self, template, science, candidateList=None, preconvolve
             Sources to check as possible kernel candidates.
         preconvolved : `bool`, optional
             Was the science image convolved with its own PSF?
+        templateFwhmPix, scienceFwhmPix : `float`, optional
```
kfindeisen:
```suggestion
        templateFwhmPix, scienceFwhmPix : `float` or `None`, optional
```

python/lsst/ip/diffim/makeKernel.py
```diff
@@ -454,3 +415,52 @@ def _adaptCellSize(self, candidateList):
             New dimensions to use for the kernel.
         """
         return self.kConfig.sizeCellX, self.kConfig.sizeCellY
+
+    def _checkPsfWidths(self, template, science, templateFwhmPix, scienceFwhmPix, preconvolved=False):
```
kfindeisen, line 419: A minor point, but it would be slightly less confusing if the order of parameters was the same from method to method.

python/lsst/ip/diffim/makeKernel.py
```diff
@@ -454,3 +415,52 @@ def _adaptCellSize(self, candidateList):
...
+        """Check the science and template FWHM.
+
+        Parameters
+        ----------
+        template : `lsst.afw.image.Exposure`
+            Exposure that will be convolved.
+        science : `lsst.afw.image.Exposure`
+            The exposure that will be matched.
+        templateFwhmPix, scienceFwhmPix : `float`, optional
```
kfindeisen: These parameters don't have a default, so aren't optional:
```suggestion
        templateFwhmPix, scienceFwhmPix : `float` or `None`
```

python/lsst/ip/diffim/subtractImages.py
```diff
@@ -451,9 +451,11 @@ def run(self, template, science, sources, visitSummary=None):
                 Sources from the input catalog that were used to construct the
                 PSF-matching kernel.
         """
+        self.usePreconvolution = False
         self._prepareInputs(template, science, visitSummary=visitSummary)
 
         convolveTemplate = self.chooseConvolutionMethod(template, science)
+        self.matchedPsfSize = self.sciencePsfSize if convolveTemplate else self.templatePsfSize
```
kfindeisen, line 461: Should the commit message read "exactly once" instead of "only once"?

python/lsst/ip/diffim/subtractImages.py
```diff
@@ -451,9 +451,11 @@ def run(self, template, science, sources, visitSummary=None):
                 Sources from the input catalog that were used to construct the
                 PSF-matching kernel.
         """
+        self.usePreconvolution = False
```
kfindeisen: Something seems fishy here. The only references that set `usePreconvolution` are:
* `AlardLuptonSubtractTask.run` sets `self.usePreconvolution = False`
* `AlardLuptonPreconvolveSubtractTask.run` sets `self.usePreconvolution = True`
* `SimplifiedSubtractTask.run` sets `self.usePreconvolution = False`

If the goal is to define `usePreconvolution` as a function of the implementation class instead of the arguments, then it would be better to set as a class constant, where at least it's clear that there won't be any side effects or weird interactions (e.g., what if `AlardLuptonPreconvolveSubtractTask.run` called `super().run()`, unlikely as it is?). It also avoids any potential problems from `usePreconvolution` being referenced _before_ `run` is called.

(TBH, making this a class variable/constant still feels wrong, but I can't come up with a better idea at the moment.)

python/lsst/ip/diffim/subtractImages.py
```diff
@@ -561,7 +565,8 @@ def runMakeKernel(self, template, science, sources=None, convolveTemplate=True,
                 Sources from the input catalog that were used to construct the
                 PSF-matching kernel.
         """
-
+        if self.usePreconvolution:
+            raise RuntimeError("Incorrect matching kernel calculation configured.")
```
kfindeisen: Be more specific about _how_ it's incorrect?

python/lsst/ip/diffim/subtractImages.py
```diff
@@ -1223,19 +1223,20 @@ def run(self, template, science, sources, visitSummary=None):
...
 
         convolutionKernel = self._makePreconvolutionKernel(science.psf)
         matchedScience = self._convolveExposure(science, convolutionKernel, self.convolutionControl,
                                                 interpolateBadMaskPlanes=True)
         self.metadata["convolvedExposure"] = "Preconvolution"
 
-        self.preconvolvedSciencePsfSize = self.sciencePsfSize*np.sqrt(2)
-        self.log.info("Preconvolved science PSF FWHM: %f pixels", self.preconvolvedSciencePsfSize)
-        self.metadata["preconvolvedSciencePsfSize"] = self.preconvolvedSciencePsfSize
+        self.matchedPsfSize = self.sciencePsfSize*np.sqrt(2)
```
kfindeisen, line 1237: Duplicate of line 1228 (though I think this spot is better, since it's immediately before first use).

python/lsst/ip/diffim/detectAndMeasure.py
```diff
@@ -664,20 +664,16 @@ def run(self, science, matchedTemplate, difference, kernelSources=None,
...
         self._prepareInputs(detectionExposure)
 
-        if self.config.doFindCosmicRays and not self.config.doSubtractBackground:
-            # Detect and interpolate over any remaining cosmic rays
-            # This only needs to run once, right before the final detection.
-            self.findAndMaskCosmicRays(detectionExposure)
+        # Detect and interpolate over any remaining cosmic rays. When
+        # background subtraction is enabled, this is deferred to the second
+        # pass so it runs only once on a properly background-subtracted image.
+        self._findInitialCosmicRays(differenceExposure=difference)
```
kfindeisen: This isn't what I had in mind; I thought it would be simpler to have something like:
```
# Preprocess or skip
if doSubtractBackground:
    detectionExposure = copy
    background = rough background
    self._prepareInputs(detectionExposure)  # Only run on first pass, but not second?
    results = detection.run(detectionExposure)
    background = final background
else:
    background = empty
    self._prepareInputs(difference)  # Run on only pass

# Final processing
self.findAndMaskCosmicRays(difference)  # current code aliases as detectionExposure if not doSubtractBackground
results = detection.run(difference)
```

What you're proposing here obfuscates the multiple `doSubtractBackground` checks instead of eliminating them.

kfindeisen (changes requested):
Mostly minor comments, but I'm still concerned about the change in role for `*Diff_scoreExp` -- something you said was a bug earlier? If it is, it would affect lsst/ap_pipe#269 as well.

python/lsst/ip/diffim/makeKernel.py
isullivan: Yes, that was intentional. Both `scienceFwhmPix` and `templateFwhmPix` have to be specified or else both are recalculated, and in that case the `preconvolved` check needs to be made.

tests/test_detectAndMeasure.py
isullivan: It is not obvious, but this is a different version than the first one. I'll update the docstring to mention the difference.

python/lsst/ip/diffim/detectAndMeasure.py
isullivan: That is correct, and is intended. There was a bug in the original implementation that got the default names wrong, but the final output is intended to be called the `Score` exposure. That includes background subtraction and detection, which only happens in `detectAndMeasure(Score)`. I am confident that this task is not being used by anyone yet, so I wanted to make the change while that was still easy.

kfindeisen (approved):
With the connections question resolved, I have no major objections.

python/lsst/ip/diffim/detectAndMeasure.py
isullivan: I've pushed a new version that refactors background subtraction and cosmic ray masking.

python/lsst/ip/diffim/subtractImages.py
isullivan: I've moved `usePreconvolution` to a class variable, and checked that it is used instead of other flags where possible.

PR #454, opened 2026-04-11
Add variance to spatiallySampledMetrics

Gerenjie:


python/lsst/ip/diffim/computeSpatiallySampledMetrics.py
```diff
@@ -231,8 +243,8 @@ def run(self, science, template, difference, diaSources, psfMatchingKernel):
             self._evaluateLocalMetric(src, science, template, difference, diaSources,
                                       metricsMaskPlanes=metricsMaskPlanes,
                                       psfMatchingKernel=psfMatchingKernel)
-
-        return pipeBase.Struct(spatiallySampledMetrics=spatiallySampledMetrics.asAstropy())
+        spatiallySampledMetrics = spatiallySampledMetrics.copy(deep=True).asAstropy()
```
isullivan, line 246: Why is the additional copy necessary?

isullivan (approved):
Looks good, just one question.

python/lsst/ip/diffim/computeSpatiallySampledMetrics.py
Gerenjie: Without the copy, it fails with a "memory is not contiguous" error on the .asAstropy() when # sources is not 100.

isullivan: Aha! So this is the part that fixes that bug. Thanks!

PR #455, opened 2026-04-23
DM-54758: Fix sigmaToFwhm conversion typo

isullivan:


erykoff (approved):
Would be nice if we had this somewhere to import rather than having to do it by hand everywhere in the stack.  But I have confirmed this is now correct.

taranu:
> Would be nice if we had this somewhere to import

`from astropy.stats import gaussian_sigma_to_fwhm, gaussian_fwhm_to_sigma`
and also
`from lsst.gauss2d import M_HWHM_SIGMA, M_SIGMA_HWHM`

... although you would not want to depend on gauss2d just for that.

isullivan:
Sure, we already import astropy here, so I'll use that for the conversions.

PR #461, opened 2026-05-28
DM-50420: Inject variable sources into the template and science images

BrunoSanchez:
Add visit information to templates useful for template fake injection

### DIFF
--- a/python/lsst/ip/diffim/getTemplate.py
+++ b/python/lsst/ip/diffim/getTemplate.py
@@ -25,7 +25,7 @@ import numpy as np
 import lsst.afw.image as afwImage
 import lsst.geom as geom
 import lsst.afw.geom as afwGeom
-from lsst.afw.image import ExposureInfo, VisitInfo
+from lsst.afw.image import VisitInfo
 import lsst.afw.table as afwTable
 from lsst.afw.math._warper import computeWarpedBBox
 import lsst.afw.math as afwMath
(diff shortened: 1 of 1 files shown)


python/lsst/ip/diffim/getTemplate.py
```diff
@@ -383,9 +385,14 @@ def run(self, *, coaddExposureHandles, bbox, wcs, dataIds, physical_filter):
 
         # Set a mask plane for any regions with exceptionally high variance.
         self.checkHighVariance(template)
-        template.setPsf(self._makePsf(template, catalog, wcs))
+        visitId = outputRefs.template.dataId['visit'] if outputRefs is not None else 1000
+        template.setInfo(
```
isullivan: Use `template.getInfo().setVisitInfo(VisitInfo(id=visitId))` instead, so that the rest of the exposureInfo is preserved.

python/lsst/ip/diffim/getTemplate.py
```diff
@@ -276,7 +278,7 @@ def getExposures(self, coaddExposureHandles, bbox, skymap, wcs):
         return pipeBase.Struct(coaddExposures=coaddExposures, dataIds=dataIds)
 
     @timeMethod
-    def run(self, *, coaddExposureHandles, bbox, wcs, dataIds, physical_filter):
+    def run(self, *, coaddExposureHandles, bbox, wcs, dataIds, physical_filter, outputRefs=None):
```
isullivan, line 281: I don't like the `run` method needing to know anything about the `outputRefs`, that should be handled in `runQuantum`

isullivan (approved):
I put my suggested changes together on a new commit. Please look it over and change as needed!

PR #465, opened 2026-07-15
DM-55502: Sort peaks in each footprint by significance before measurement

isullivan:


BrunoSanchez (approved):
Looks good to me, test passing is the key point.

PR #468, opened 2026-08-20
DM-55865: Make templates' data constraint configurable in pipelines 

hsinfang:


python/lsst/ip/diffim/getTemplate.py
```diff
@@ -129,6 +137,12 @@ class GetTemplateConfig(
         target=ScaleVarianceTask,
         doc="Subtask to rescale the variance of the template to the statistically expected level."
     )
+    requireCoaddAtGraphBuild = pexConfig.Field(
+        dtype=bool,
+        default=False,
+        doc="If True, include the coadd dataset existence in the"
```
isullivan, line 144: Very minor, but a space is needed either after "the" or before "initial"

python/lsst/ip/diffim/getTemplate.py
```diff
@@ -86,6 +87,13 @@ class GetTemplateConnections(
         name="{fakesType}{coaddName}Diff_templateExp{warpTypeSuffix}",
     )
 
+    def __init__(self, *, config=None):
+        if config.requireCoaddAtGraphBuild:
```
isullivan, line 92: It is not currently needed, but the other tasks in `ip_diffim` generally expect `__init__` to include  `super().__init__(config=config)`

isullivan (approved):
A couple minor comments.

PR #469, opened 2026-08-31
DM-55993: Move option to scale template variance

isullivan:


python/lsst/ip/diffim/getTemplate.py
```diff
@@ -394,6 +413,7 @@ def run(self, *, coaddExposureHandles, bbox, wcs, dataIds, physical_filter, visi
         template.setFilter(afwImage.FilterLabel(band, physical_filter))
         template.setPhotoCalib(photoCalib)
         template.setPsf(self._makePsf(template, catalog, wcs))
+
```
kherner, line 416: Was this blank line intentional?

isullivan: It was an artifact of adding lines there originally, then moving them earlier. But, I liked the visual space so I left it in. I can remove it if you prefer!

kherner (approved):

